Issue #27, opened 2015-07-06
Swagger paths without operationId property should not crash connexion


hjacobs:
To ease incremental development, Connexion should not fail if one of the operationId properties is missing.

$ python3 hack.py 
Traceback (most recent call last):
  File "hack.py", line 4, in <module>
    app.add_api('swagger.yaml')
  File "/usr/local/lib/python3.4/dist-packages/connexion/app.py", line 81, in add_api
    api = connexion.api.Api(yaml_path, base_path, arguments, swagger_ui)
  File "/usr/local/lib/python3.4/dist-packages/connexion/api.py", line 73, in __init__
    self.add_paths()
  File "/usr/local/lib/python3.4/dist-packages/connexion/api.py", line 106, in add_paths
    self.add_operation(method, path, endpoint)
  File "/usr/local/lib/python3.4/dist-packages/connexion/api.py", line 90, in add_operation
    security_definitions=self.security_definitions)
  File "/usr/local/lib/python3.4/dist-packages/connexion/operation.py", line 52, in __init__
    self.operation_id = operation['operationId']
KeyError: 'operationId'

Issue #4, opened 2015-06-12
Add validation for Swagger schema definitions


hjacobs:
Requests should be validated automatically using the schema definitions defined in the Swagger file (similar to what https://github.com/sarnowski/swagger1st/blob/master/src/io/sarnowski/swagger1st/parser.clj does).

PR #28, opened 2015-07-06
wrap add_operation with try/except to continue even if operationId or…

hjacobs:
… function is missing


hjacobs:
workaround/fix for #27

hjacobs:
added same basic validation for #4

PR #29, opened 2015-07-07
Schema Validation

jmcs:


connexion/decorators/validation.py
```diff
@@ -1,25 +1,64 @@
...
             raise abort(400)
-        for required_key in schema.get('required', []):
+
+        # verify if required keys are present
+        required_keys = schema.get('required', [])
+        logger.debug('... required keys: %s', required_keys)
+        for required_key in schema.get('required', required_keys):
             if required_key not in data:
+                logger.debug("... '%s' missing", required_key)
                 raise abort(400)
```
hjacobs: we should add better error messages, but we can do that later :-)

PR #30, opened 2015-07-07
Add release.sh helper script to build and release to PyPI

hjacobs:


release.sh
```diff
@@ -0,0 +1,28 @@
...
+python3 setup.py clean
+python3 setup.py test
+python3 setup.py flake8
+
+git add setup.py
+
+git commit -m "Bumped version to $version"
+git push
+
+python3 setup.py sdist upload
```
jmcs: bdist_wheel is better for distribution than sdist. I think we can use both in sequence but I've never done it.

hjacobs:
seems to work:

python3 setup.py sdist bdist_wheel upload

Issue #51, opened 2015-08-18
JSONEncoder datetime.date encoding fails


cansik:
The default function of the [JSONEncoder](https://github.com/zalando/connexion/blob/master/connexion/decorators/produces.py) does only handle `datetime.datetime` types in a special way. So if you insert a `datetime.date` the method fails.

Would be great if you could extend the default function with a special treatment for `datetime.date` objects:

```python
def default(self, o):
    if isinstance(o, datetime.datetime):
        # only supports UTC timestamps
        return o.isoformat('T') + 'Z'
        
    if isinstance(o, datetime.date):
        return o.isoformat()
    return json.JSONEncoder.default(self, o)
```

This converts the `date` into a [W3C](http://www.w3.org/TR/NOTE-datetime) conform string format.

hjacobs:
Your fix looks straight forward :-)

PR #52, opened 2015-08-18
added date support for JSONEncoder

cansik:
This fixes the JSONEncoder bug as described in #51 .


cansik:
Couldn't find a test function which tests this method. This is the reason why the code-coverage drops.

Issue #48, opened 2015-08-13
Query and Path Parameter Validation


hjacobs:
Only "body" parameters are validated right now, we need to validate other parameters ("path", "query") too.

hjacobs:
Basic query parameter validation is now part of Connexion 0.9.1, I will add basic path parameter validation next.

cansik:
:+1:

hjacobs:
Path and header parameters are now also validated in Connexion 0.9.2.
Note that some less common Swagger validation rules are not yet implemented.

samuelvonstachelski:
:+1:

samuelvonstachelski:
I get the following exception on all path parameters even though the swagger editor tells me, that my file is correct...

    ERROR:connexion.api:Failed to add operation for PARAMETERS /containers/<id>
    Traceback (most recent call last):
      File "/usr/local/lib/python2.7/dist-packages/connexion/api.py", line 131, in add_paths
        self.add_operation(method, path, endpoint)
      File "/usr/local/lib/python2.7/dist-packages/connexion/api.py", line 112, in add_operation
        parameter_definitions=self.parameter_definitions)
      File "/usr/local/lib/python2.7/dist-packages/connexion/operation.py", line 72, in __init__
        self.operation_id = operation['operationId']
    TypeError: list indices must be integers, not str

hjacobs:
@samuelvonstachelski  can you paste your swagger.yaml here?

hjacobs:
It looks like your operation is not a map, but a list (this would not conform with Swagger spec).

samuelvonstachelski:
swagger: "2.0"
info: 
  title: "FLARECAST pipeline service"
  version: "0.1"
consumes: 
  - application/json
produces: 
  - application/json
paths: 
  /:
    get:
      operationId: pipeline.routes.root
      responses:
        200:
          description: List of base resources
          schema:
            type: array
            items:
              enum:
                - algorithms
                - containers
                - pipeline
  /algorithms: 
    get: 
      operationId: pipeline.routes.algorithms
      responses: 
        200: 
          description: List IDs of all algorithm images
          schema: 
            type: array
            items:
              $ref: "#/definitions/DockerId"
  /algorithms/{id}: 
    get: 
      operationId: pipeline.routes.get_algorithm
      responses: 
        200: 
          description: Description of algorithm image
          schema:
            properties:
              DockerInspect:
                type: object
              UserParams:
                type: string
        404:
          description: Id not found
[...]

PR #56, opened 2015-08-26
Feature/parameter validation issue 48

hjacobs:
Basic parameter validation supporting integer, number and boolean query parameters, format and pattern.


hjacobs:
Partly implements #48

PR #58, opened 2015-08-28
Allow configurable path and URL for swagger-ui

jvinet:
This PR modifies the App and API classes to allow the caller to specify a different directory for the swagger-ui files and/or a different URL path to access the API documentation.

Additionally, it does not emit a loud warning if the API spec specifies a security method other than oauth2. Instead, it warns the user (via a lower logging level) that she should handle the security herself (eg, through a custom decorator).


hjacobs:
Can you check the Travis CI build please?

You can check flake8 formatting locally by running:

```
python3 setup.py flake8

hjacobs:
BTW, thanks for the PR --- more configuration options make sense in this case :-)

PR #60, opened 2015-09-04
Add validation for array types in query or path

jvinet:
This PR adds support for validating arrays that are present within the 'query' or 'path' inputs. Prior to this, arrays were only validated if they were present in the 'body' of a request.

A corresponding unit test was also added.


hjacobs:
Great, thanks! :+1:

PR #67, opened 2015-09-25
Serialization: Don't append 'Z' to datetime objects if they already have a timezone

jvinet:
When serializing responses to JSON, don't append 'Z' to datetime objects if they already have an associated timezone; otherwise, datetime strings can violate RFC 3339 (eg: '2015-09-25T23:14:42+00:00Z'). Note the extraneous 'Z' in the example.


jmcs:
Thanks for the Pull Request.

Issue #66, opened 2015-09-25
List as body parameter fails in validation


cansik:
As I saw you modified body parameter handling in version 0.10.2 #59. Now the problem is that lists cause an error because you try to call `items` on a `list` object which fails.

It looks like you just handle dicts (JSON Objects) as body parameters ([code](https://github.com/zalando/connexion/blob/9593e9e4611d9d434f00d7057b4e5f2de4308273/connexion/decorators/parameter.py#L57)):

```python
        # Add body parameters
        for key, value in body_parameters.items():
            if key not in arguments:
                logger.debug("Body parameter '%s' not in function arguments", key)
```

My swagger config for this route looks like this:

```yaml
  /provenance/:
    post:
      tags:
        - INGEST
      description: creates a new provenance if it does not exist
      operationId: propertyservice.routes.provenance.create
      parameters:
        - name: provenance_data
          in: body
          description: provenance_data
          required: true
          schema:
            type: array
            items:
              type: string
      responses:
        200:
          description: OK
```

And this is what I send with the POST request as body:

```json
["test2"]
```

#### Full error message
```python
INFO:werkzeug:127.0.0.1 - - [25/Sep/2015 13:52:47] "POST /provenance/ HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Library/Python/2.7/site-packages/flask_cors/extension.py", line 110, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Library/Python/2.7/site-packages/flask_cors/extension.py", line 110, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/Library/Python/2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
[...]

PR #68, opened 2015-09-28
Handle non object body parameters

jmcs:
Fix for #66.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -39,9 +39,6 @@ def parameter_to_arg(parameters, function):
     """
     body_parameters = [parameter for parameter in parameters if parameter['in'] == 'body'] or [{}]
     body_name = body_parameters[0].get('name')
-    body_schema = body_parameters[0].get('schema', {})  # type: dict[str, str]
-    body_properties = body_schema.get('properties', {})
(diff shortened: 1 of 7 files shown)


hjacobs:
There can be at most one body parameter according to the Swagger spec (http://swagger.io/specification/). I think getting the value from the body parsed as dict is wrong.

```
There can be one "body" parameter at most.

File: connexion/decorators/parameter.py [hunk 1]
```diff
@@ -49,20 +51,27 @@ def wrapper(*args, **kwargs):
...
-            body_parameters = flask.request.json or {}  # type: dict
+            body_parameters = flask.request.json or {}
         except exceptions.BadRequest:
             body_parameters = {}
 
         # Add body parameters
-        for key, value in body_parameters.items():
-            if key not in arguments:
-                logger.debug("Body parameter '%s' not in function arguments", key)
+        if isinstance(body_parameters, dict):
```
hjacobs: I think getting the value from the body parsed as dict is wrong. 'There can be one "body" parameter at most.' the Swagger spec says.

jmcs: When connexion receives a request with type ``object`` body_parameters is a dict and I was getting the keys from inside that object and passing them to the function as arguments. Do you think it would be better to provide the body as a single argument in this case?

hjacobs:
Looks better :+1:

PR #70, opened 2015-10-02
Include oauth authentication information on request

jmcs:


hjacobs:
@jmcs can you check this? Hardcoding the "uid" scope does not look right, maybe we can use tokeninfo.get('uid') and/or allow defining a function to get the "user" from the token info data

PR #72, opened 2015-10-09
Options Parameter for Werkzeug

cansik:
### What?
The Flask app provides a `kwargs` argument called `options` to forwarded them to the underlying **Werkzeug** server. Currently **connexion** does not provide a way to forward parameter to Werkzeug, so I've implemented the parameter passing into the **connexion app**.

### Why?
We need this to disable the **reloader** in debug mode.

### Example

```python
app = connexion.App(
    __name__,
    port=8000,
    specification_dir='/swagger/',
    debug=True,
    # options kwargs
    use_reloader=False
)
```

### DIFF
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -25,7 +25,7 @@ logger = logging.getLogger('connexion.app')
 
 class App:
     def __init__(self, import_name, port=5000, specification_dir='', server=None, arguments=None, debug=False,
-                 swagger_ui=True, swagger_path=None, swagger_url=None, **options):
+                 swagger_ui=True, swagger_path=None, swagger_url=None):
         """
         :param import_name: the name of the application package
         :type import_name: str
@@ -45,8 +45,6 @@ class App:
         :type swagger_path: string | None
         :param swagger_url: URL to access swagger-ui documentation
         :type swagger_url: string | None
-        :param options: options to be forwarded to the underlying Werkzeug server
-        :type options: dict
         """
         self.app = flask.Flask(import_name)
 
@@ -73,7 +71,6 @@ class App:
         self.swagger_ui = swagger_ui
         self.swagger_path = swagger_path
         self.swagger_url = swagger_url
-        self.options = options
 
     @staticmethod
     def common_error_handler(e):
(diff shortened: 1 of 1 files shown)


jmcs:
I agree that there should be a way to provide options to the server, but I think the options would be more useful in app.run because they depend on the server that will be used. And the method should probably also pass the kwargs to the other servers.

cansik:
I've implemented it this way because you don't have any parameters on `app.run()`. So I thought this is your design approach to use the constructor instead the run method to pass arguments to the server (like port and debug flag).

So wouldn't it be better if you pass all the server related stuff in the run method?

jmcs:
In hindsight yes.

cansik:
The possibilities we have is to:
1. use the constructor to match the actual architecture design
2.make an exception and use `app.run(**options)`
3. redesign the architecture and split up server and app params
4. just copy all the server params to the `app.run` method and stay comaptible

So what would you prefer?

jmcs:
In point 4 do you mean allowing run to override the values provided in the constructor? This would be the better option because it would allow us to remain compatible with applications that already use connexion while avoiding making the constructor messier in the future.

cansik:
Yes that was my intention with point 4. It's not that beautiful but I think everyone could live with this.
Maybe something like a deprecated warning would be nice for the constructor arguments.

Is it ok if I update my pull-request or would you like to change the this?

jmcs:
If you don't mind it's ok for you to change the pull request.

cansik:
So I've copied now following parameter to the `app.run` method:
* port
* server
* debug
* **options

At the moment the options parameter does only affect the **flask** / **werkzeug** server and ***not*** **tornado** and **gevent**. That is because I couldn't find similar methods for those. Would be great if you could send me these if there are.

The default values of the parameter is `None` because I have to check if they have been changed. And `None` is not a valid value for those parameters at the moment. If you may think `None` is a valid value for server or port, we have to extend the method to act like the constructor and use a **sentinel** to check for changes in the parameter.

jmcs:
Thanks for the PR.

PR #77, opened 2015-10-27
Proposed technique for returning custom headers (with simple unit test)

danballance:
Proposed technique for returning custom headers (with simple unit test)

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -57,17 +57,18 @@ class BaseSerializer:
         """
         url = flask.request.url
         logger.debug('Getting data and status code', extra={'data': data, 'data_type': type(data), 'url': url})
+        status_code, headers = 200, []
         if isinstance(data, flask.Response):
             data = data
             status_code = data.status_code
+        elif isinstance(data, tuple) and len(data) == 3:
+            data, status_code, headers = data
         elif isinstance(data, tuple) and len(data) == 2:
             data, status_code = data
-        else:
-            status_code = 200
         logger.debug('Got data and status code (%d)', status_code, extra={'data': data,
                                                                           'data_type': type(data),
                                                                           'url': url})
-        return data, status_code
+        return data, status_code, headers
 
     def __call__(self, function):
         """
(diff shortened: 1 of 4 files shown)


danballance:
Looks like I reduced the code coverage a little. If you want to accept this, I'll improve the coverage of the unit test.

hjacobs:
@jmcs could you check this?

connexion/decorators/produces.py
```diff
@@ -57,17 +57,18 @@ def get_data_status_code(data):
         """
         url = flask.request.url
         logger.debug('Getting data and status code', extra={'data': data, 'data_type': type(data), 'url': url})
+        status_code, headers = 200, []
```
jmcs: I would make headers `None` or `{}` by default.

connexion/decorators/produces.py
```diff
@@ -57,17 +57,18 @@ def get_data_status_code(data):
         """
```
jmcs: please update the rtype to (object, int, dict)

connexion/decorators/produces.py
danballance: Hi,

That's fine by me. My reasoning for using a list of tuples was in case the
sequencing of the headers was important to a user for some reason, but I
agree a dictionary is a simpler approach. Could we use an ordered
dictionary maybe?

I just had a quick look at the HTTP spec and it seems the ordering of
headers is not significant, except in the slightly more edge case of where
multiple headers with the same name are present:

"The order in which header fields with differing field names are received
is not significant. However, it is "good practice" to send general-header
fields first, followed by request-header or response- header fields, and
ending with the entity-header fields.

Multiple message-header fields with the same field-name MAY be present in a
message if and only if the entire field-value for that header field is
defined as a comma-separated list [i.e., #(values)]. It MUST be possible to
combine the multiple header fields into one "field-name: field-value" pair,
without changing the semantics of the message, by appending each subsequent
field-value to the first, each separated by a comma. The order in which
header fields with the same field-name are received is therefore
significant to the interpretation of the combined field value, and thus a
proxy MUST NOT change the order of these field values when a message is
forwarded."

( http://tools.ietf.org/html/rfc2616#page-31 )

It's your call, what shall we go with, list of tuples, ordered dictionary
or plain dictionary?

Dan

jmcs: For the default, empty, value it can be just an empty dict. I'll add the recommendation to use an ordered dict in the case the order matters as part of #78.

danballance:
One other question I had was would you like me to rename the function get_data_status_code to something more meaningful since it now returns data, status_code and headers?

connexion/decorators/produces.py
danballance: No problem.

jmcs:
> There are only two hard problems in Computer Science: cache invalidation and naming things. - Phil Karlton

Do you have any suggestion for the new name?

Issue #76, opened 2015-10-23
Document, how headers are passed to the methods


valgog:
Please document and provide some examples of how headers are passed into the methods

hjacobs:
There is no special trick here: you can simply access the Flask headers via "flask.request.headers", see http://flask.pocoo.org/docs/0.10/api/#flask.Request.headers

valgog:
This way one can say, that all parameters are accessible with this "trick". 

Headers are 1st class citizens in Swagger and are typed as well. So it seems that they should be pushed into the method as a parameter and typing should be checked and probably values converted, just like for the normal parameters.

danballance:
That would be a nice feature if it was possible.

hjacobs:
@valgog ok, I agree. I would define most headers not in the Swagger file (as they are usually technical metadata), but I see that there are valid use cases for using headers as first class citizens as defined in Swagger spec.

danballance:
I agree with @hjacobs here - in that we wouldn't want to define every header. Rather just have the option to define one or more important headers in the spec. If a header is specified, then validating it would be great. Ideally this would work in both directions - request and response headers.

I had an example recently where in my spec I want to specify that a Location header is returned for a POST operation. We have the functionality now to return that extra header from the view. It would be great if we could verify the header is returned and possibly validate it too. I didn't think about this one I made my pull request unfortunately.

kleijnweb:
I don't really see a use case for passing to controllers. Headers are to be picked up by the infrastructure (gateways and the framework), usually. I realize you can specify `in: header` according to the docs, but that doesn't necessarily mean they should be passed to controllers. They *should* be validated and type casted though.

The Swagger spec is IMO a bit vague on header params. I mean this (from the spec):

```yml
name: token
in: header
description: token to be passed as a header
required: true
type: array
items:
  type: integer
  format: int64
collectionFormat: csv
``` 

Is fine, except for the name. Are we not using the "X-" extension prefix for HTTP header names? Dash-separated-upper-case-first? *Or* are we leaving out a transformation from header name to parameter name? In the latter case the interface would be incomplete so we have to assume the former. So we're talking:

```yml
name: X-Token
in: header
description: token to be passed as a header
required: true
type: array
items:
  type: integer
  format: int64
collectionFormat: csv
``` 

So we could strip `X-`, convert the rest to snake-case and pass it to the function. But what you probably want is to have it picked up by some middleware (eg Flask decorators), instead.

[...]

PR #84, opened 2015-11-01
Validating the developer's returned response against the specification

danballance:
Hi guys,

I wonder if somebody might be willing to review this initial code? (Mentioned by me here: https://github.com/zalando/connexion/issues/76) The idea is to validate the headers and response body returned by the developer from the view func against what was declared in the Swagger specification. If the headers and response schema do not match then an error 500 is generated. I personally think this would be a useful part of the development process because Connexion would be checking to ensure that what is returned as a response actually matches what was promised.

A few notes that came up as I did this work:

1. I passed a reference to the Operation object into the Serializer so that I could access what I needed from the spec. I hope that's okay?
2. The test_app.py unit tests and app.yaml spec are getting quite long now. They might need re-factoring into multiple files? I'm happy to take a look at this if you think it would help.
3. At the moment all declared response headers in the Spec are being treated as required. I can't see a definitive description of the correct behaviour in the spec: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#headersObject. Any ideas/sources on this?
4. TODO: I still need to add actual validation of the header by type etc.
5. I'm not 100% sure on this - but now that I have started validating the schemas of response bodies I have had to alter the response types of some of the early tests. For example, the GET operation on "/greetings/{name}" looks to me like it should be object and not string. Dose this look correct to you?

Comments, feedback, amends etc welcomed :)

Dan

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -17,6 +17,10 @@ import flask
 import functools
 import json
 import logging
+from ..exceptions import NonConformingResponse
+from ..problem import problem
+from .validation import RequestBodyValidator
+import ast
 
 logger = logging.getLogger('connexion.decorators.produces')
 
@@ -43,11 +47,13 @@ class JSONEncoder(json.JSONEncoder):
 
 
 class BaseSerializer:
-    def __init__(self, mimetype='text/plain'):
+    def __init__(self, mimetype='text/plain', operation={}):
         """
         :type mimetype: str
+        :type operation: Operation
         """
         self.mimetype = mimetype
+        self.operation = operation
 
     @staticmethod
     def get_full_response(data):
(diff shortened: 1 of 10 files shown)


danballance:
Okay looks like I've broken the Python 3 tests. I need to polish up my Python 3 code a little. I'll get this sorted if you think the idea is worth pursuing.

hjacobs:
@danballance yes, I think validating responses can make sense, but I would make it optional (configurable) and not on by default. Validating responses is basically checking the app's code for sanity and appears to me like having "assert" statements in the code.

In short: go ahead, I like the possibility to validate responses (conforming to Swagger spec) --- just make it optional :smile:

connexion/decorators/produces.py
```diff
@@ -101,15 +182,21 @@ def __call__(self, function):
         def wrapper(*args, **kwargs):
             url = flask.request.url
             data, status_code, headers = self.get_full_response(function(*args, **kwargs))
+            logger.debug((data, status_code, headers))
```
jmcs: Please put some text here so it's clear in the logs what it is.

connexion/decorators/produces.py
```diff
@@ -76,6 +82,81 @@ def get_full_response(data):
...
+                return float(data_string)
+            except ValueError:
+                pass
+        try:  # integer?
+            return int(data_string)
+        except ValueError:
+            pass
+        return data_string  # leave as a string
+
+    def validate_response(self, response, status_code, mimetype):
```
jmcs: It would probably be better to have this a as independent, optional, decorator instead of being inside the `produces` decorator.

hjacobs: I agree with @jmcs: the whole response validation feature should be a separate decorator which can be turned on / off (and is "off" by default)

danballance:
Hi guys, thanks for the feedback and suggestions. I'll take a look this week.

connexion/decorators/produces.py
danballance: Good points, will will re-factor :+1:

hjacobs:
@danballance thank you for contributing :sunny:

connexion/decorators/produces.py
danballance: Hi @jmcs,  @hjacobs,

Having taken a look at the code again, I'm struggling a little to break this out into a separate decorator. This may be due to a misunderstanding of Python decorators on my behalf, but the reason I originally left this code in the Serializer is because it needs access to data that the serializer produces - namely the flask.Response, the status_code and the headers. (I.e. the result of self.get_full_response.)

So I'm struggling a little bit to see how I can process the result of self.get_full_response within a different decorator?

If anyone had 5 minutes to guide me a little here that woul be great. Apologies if I have missed something obvious.

danballance: Unless maybe you mean another super class to BaseSerializer which can also access self.get_full_response?

connexion/decorators/produces.py
danballance: Apologies, this was some testing code of mine that ended up in the commit. I've removed for now.

danballance:
Okay, so I have re-factored a little. The main change which I hope you are okay with is to create a BaseDecorator class to hold the function get_full_response which both the Serializer decorators and the new responses decorator both need access to.

Moving the response validation into a separate decorator that calls get_full_response I think has worked quite nicely because it no longer needs to reverse engineer the types from the string contained in the Response object and can instead directly access the data returned by the view function.

There's a new parameter validate_responses that is optionally passed to App.add_api which then passes this on to the Operation class constructor. This value is then used in a simple if statement to decide if the response decorator should be introduced into the chain.

All unit tests currently passing in python 2 and 3. Feedback appreciated :)

hjacobs:
@danballance I probably can't review it today, but maybe @jmcs can and I will do later..

danballance:
No rush at all. Whenever you have some time!

jmcs:
Thanks for the Pull Request.

Issue #85, opened 2015-11-05
Allow adding headers to problem responses


jmcs:
It should be possible to add HTTP headers to a problem response to easy debugging.

PR #86, opened 2015-11-05
#85 Allow adding headers to problem responses

jmcs:


connexion/problem.py
```diff
@@ -33,12 +34,21 @@ def problem(status, title, detail, type='about:blank', instance=None):
...
 
-    return flask.current_app.response_class(json.dumps(problem_response),
-                                            mimetype='application/problem+json',
-                                            status=status)
+    body = json.dumps(problem_response)
+    response = flask.current_app.response_class(body, mimetype='application/problem+json',
+                                                status=status)  # type: flask.Response
+
+    if headers:
+        for key, value in headers.items():
```
hjacobs: can't you use dict.update(..) here?

jmcs: unfortunately response.headers is not a dict and doesn't have an update method.

danballance: I noticed that too :+1:

Issue #83, opened 2015-10-29
Reconsider logging client errors (HTTP status 4xx) with "ERROR" log level


hjacobs:
Connexion is currently logging some client failures (HTTP 4xx) as errors (logging.error or logging.exception), we should reconsider/discuss this.

Client errors should not spam the server log file. We could (depending on the severity) consider INFO or WARN log levels.

```
grep -nr logger.error connexion
connexion/decorators/security.py:57: logger.error("... No auth provided. Aborting with 401.")
connexion/decorators/security.py:74: logger.error("... User scopes (%s) don't include one of the allowed scopes (%s). Aborting with 401.",
connexion/decorators/validation.py:132: logger.error("Unrecognized collectionFormat, cannot validate: %s", col_fmt)
connexion/decorators/validation.py:194: logger.error("'%s' is not a '%s'", data, expected_type_name)
connexion/decorators/validation.py:211: logger.error("Missing parameter '%s'", required_key, extra=log_extra)

hjacobs:
Let's discuss this before changing any code :smile:

hjacobs:
I would strongly suggest to either remove "... No auth provided. Aborting with 401." or log it with INFO or DEBUG (crawlers, spiders etc will always pass by your API without any token...)

PR #87, opened 2015-11-05
#83 change log severity levels in the decorators

jmcs:


connexion/decorators/security.py
```diff
@@ -54,7 +53,7 @@ def wrapper(*args, **kwargs):
         logger.debug("%s Oauth verification...", request.url)
         authorization = request.headers.get('Authorization')  # type: str
         if not authorization:
-            logger.error("... No auth provided. Aborting with 401.")
+            logger.info("... No auth provided. Aborting with 401.")
```
jmcs: I decided to go for info here because it's the same severity used on the successful case.

hjacobs:
:+1: I think the new log levels make sense

PR #90, opened 2015-11-09
Fixed an issue with query parameters not properly handling arrays

danballance:
Hi,

I think I have hit a small issue this morning with query parameters passed as a "csv array". So my GET request would look something like this:

/v1.0/test_array_csv_query_param?items=one,two,three

And I was getting parameter items like this:

[u'o', u'n', u'e', u',', u't', u'w', u'o', u',', u't', u'h', u'r', u'e', u'e']

(A string cast to a list).

Please take a look at my suggested fix for this in this pull request. I have only supported the "csv" and "pipes" collectedFormat attributes from the swagger spec currently (http://swagger.io/specification/#parameterObject). Tabs and spaces don't make much sense to me as query parameters.

Let me know what you think guys. I've added a couple of tests that hopefully show the behaviour.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -29,6 +29,24 @@ def get_function_arguments(function):  # pragma: no cover
         return inspect.getargspec(function).args
 
 
+def make_type(value, type):
+    type_func = TYPE_MAP[type]  # convert value to right type
+    return type_func(value)
+
+
+def get_val_from_param(value, query_param):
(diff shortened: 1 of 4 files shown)


hjacobs:
@danballance could you check the decreasing code coverage?

hjacobs:
Thanks for contributing :bow:

danballance:
Yes, no problem, I should have a gap tomorrow :)

hjacobs:
:+1:

PR #91, opened 2015-11-09
added flask-compress flag to app

kashif:


jmcs:
You can do this on your app without adding extra dependencies to connexion by calling `compress.init_app(connexion_app.app)` before calling `connexion_app.run`.

kashif:
cool thanks @jmcs

PR #95, opened 2015-11-18
added logic for detecting x-swagger-router-controller

### LINKED ISSUE #94: change use of `operationId`
The spec field `operationId` is used for client generation method names.  While connexion uses `operationId` for routing directly to a controller, it will make for some very ugly client SDK methods.

For example:

```yaml
operationId: my.controller.default_controller.getUser
```

Will generate into this for ObjectiveC:

```objc
-(NSNumber*) myControllerDefaultControllerGetUserWithCompletionBlock
```

It would be nice if connexion followed the same strategy as [swagger-inflector](https://github.com/swagger-api/swagger-inflector#locating-the-target-method) and [swagger-node](https://github.com/swagger-api/swagger-node#4-write-controller-code-in-nodejs) for locating the routing controller/method.

Specifically, I'd propose the following:

```yaml
operationId: getUser
x-swagger-router-controller: my.controller.default_controller
```

Which allows for a friendly name for the client SDK and server controller method, and uses the vendor extensions for the controller location.

fehguy:
closes #94 

adds logic for detecting `x-swagger-router-controller` and if set, will join the `operationId` with this value.

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -76,7 +76,13 @@ class Operation:
         self.validate_responses = validate_responses
 
         self.operation = operation
-        self.operation_id = operation['operationId']
+        operation_id = operation['operationId']
+        router_controller = None
+
+        if 'x-swagger-router-controller' in operation:
+            router_controller = operation['x-swagger-router-controller']
+
+        self.operation_id = self.detect_controller(operation_id, router_controller)
         # todo support definition references
         # todo support references to application level parameters
         self.parameters = list(self.resolve_parameters(operation.get('parameters', [])))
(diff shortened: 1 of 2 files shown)


hjacobs:
OK, that sounds like a reasonable approach being backwards-compatible :sunny:

hjacobs:
@jmcs what's your opinion on this?

connexion/operation.py
```diff
@@ -76,7 +76,13 @@ def __init__(self, method, path, operation, app_produces, app_security,
         self.validate_responses = validate_responses
 
         self.operation = operation
-        self.operation_id = operation['operationId']
+        operation_id = operation['operationId']
+        router_controller = None
```
jmcs: You can reduce the amount of code by doing `router_controller = operation.get('x-swagger-router-controller')` as this will return None if `x-swagger-router-controller`is not defined.

jmcs:
In general I like the approach, maybe this could also be implemented at API level (like `basePath`). I think this would be a cool addiction to swagger spec to ensure consistency between frameworks like connexion and swagger1st (@fehguy  :wink:).

PR #100, opened 2015-11-25
updated swagger-ui

kashif:


hjacobs:
@jmcs let's review this first.
@kashif is there anything specific you needed to update Swagger UI for?

kashif:
@hjacobs i was mostly interested in getting the ui working with the oauth tokens... but yes looking at the diffs there are a few fixes in the swagger ui js etc. so i thought it might be useful...

hjacobs:
OAuth integration in Swagger UI is not as easy as you might think --- but there is a great "workaround" for you: https://github.com/zalando/chrome-oauth-bearer-plugin

Basically the Chrome OAuth Bearer Plugin automatically injects tokens (from IAM's Token Service) into requests for your (configured) URLs. That's how you can use Swagger UI in your REST service without worrying about OAuth.

kashif:
@hjacobs i see thanks!

jmcs:
I don't see any reason for not updating.

Issue #104, opened 2015-12-04
Typecasting?


kleijnweb:
connexxion seems to pass Swagger integers as unicode strings. At least a note in the docs would be a start.

hjacobs:
Depends on what you mean, type casting "partly" works, see e.g. https://github.com/hjacobs/connexion-example

Can you describe your situation / use case where it fails?

kleijnweb:
To reproduce:

`swagger.yml`:

```yaml
swagger: '2.0'
host: localhost
paths:
  '/foo/{id}':
    get:
      operationId: api.get
      parameters:
        - in: path
          name: id
          type: integer
          format: int64
```

`api.py`:

```python
#!/usr/bin/env python2.7
import connexion

def get(id):
    return str(type(id))

app = connexion.App(__name__)
app.add_api('swagger.yml')
application = app.app

if __name__ == '__main__':
    app.run(port=8080)
```

```bash
http http://localhost:8080/foo/1
```

Actual:

```
HTTP/1.0 200 OK
Content-Length: 18
Content-Type: application/json
Date: Sat, 05 Dec 2015 17:53:49 GMT
Server: Werkzeug/0.11.2 Python/2.7.6

"<type 'unicode'>"
```

Expected:

```
HTTP/1.0 200 OK
Content-Length: 18
Content-Type: application/json
Date: Sat, 05 Dec 2015 17:53:49 GMT
Server: Werkzeug/0.11.2 Python/2.7.6

"<type 'int'>"
```

kleijnweb:
Sidenote, this also works (`"<type 'unicode'>"`):

```yml
swagger: '2.0'
host: localhost
paths:
  '/foo/{id}':
    get:
      operationId: api.get
      parameters:
        - in: path
          name: id
          type: fantasia
```

It would be nice if connexion was a little more defensive, this way typos can lead to hard to track bugs.

hjacobs:
True, that's a missing feature: Flask route converters should be supported http://flask.pocoo.org/docs/0.10/api/#url-route-registrations

kleijnweb:
In the mean time, a small note in the README might save new users  some time. I myself was baffled when a `pymongo.find_one()` operation did not yield any results because the argument was not an integer. Naive, perhaps, but still.

hjacobs:
@kleijnweb my personal recommendation (off topic): stay away from MongoDB :smirk: (we replaced a 5 node MongoDB cluster with a single node PostgreSQL :elephant:)

[...]

PR #105, opened 2015-12-05
Add type casting (and type-based routing) for int/float path params

hjacobs:
Adds type casting for path parameters (only int/float), see #104


kleijnweb:
LGTM :)

hjacobs:
Thanks, I will wait for @jmcs' approval :hourglass_flowing_sand:

jmcs:
I only saw this now, sorry. Here's a retroactive :+1:

Issue #98, opened 2015-11-20
Allow for optional operationId


fehguy:
The swagger spec states that operationId is optional.  We should auto-generate one if it does not exist--there is a simple algorithm that can be used here:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1086-L1107

hjacobs:
@fehguy what would be the use case for this? Without a proper operationId we cannot resolve to any Python function, i.e. it would be like "operation not implemented".

fehguy:
The ideal behavior would be to send mock responses for inimplemented methods. This is how both the node and Java frameworks operate, and it has been extremely well received.

valgog:
One could define some special function, that is used if operationId is not present, and it would default to returning "operation not implemented".

kleijnweb:
@fehguy I strongly disagree. The ideal behavior would be provide a "sensible default", eg something matching the HTTP method (as an aside, you are referring to stubbing, not mocking).

hjacobs:
@kleijnweb could you elaborate on your "sensible default" proposal? I also dislike implementing stubbing/mocking as default behavior -- "not implemented" responses look more like the way to go (if one would really want to start the server without an operationId mapping)

valgog:
What speaks against returning "Not implemented" error on invocation and putting a WARNING in the logs during initialization for the methods without defined `operationId` ?

kleijnweb:
@hjacobs, @valgog 

Well, that's not it being optional. That's deferring the handling of the fact that it is required. With a Swagger spec you know all the endpoints, and you can know in advance that the Operation does not resolve. As a dev (-opper) I'd rather know sooner than later.

As I said, a sensible default would be the method name. HTTP methods are perfectly good verbs to use as operation names.

Then you need to know which module to look in, which I figure can look like this:

```python
def detect_controller(self):

    operation_id = self.operation['operationId']

    # Since operationId should be a unique identifier, never combine it with x_router_controller
    # See https://github.com/apigee-127/swagger-tools/issues/221
    if operation_id:
        return operation_id

    x_router_controller = self.operation['x-swagger-router-controller']

    if x_router_controller:
        mod_name = x_router_controller
    else:
        mod_name = self.default_router_controller_name

    return mod_name + '.' + self.method
```

Where `default_router_controller_name` should be passed down from `connexxion.App`, defaulting to `import_name`.

[...]

PR #107, opened 2015-12-06
Sensible defaults for controller resolution

kleijnweb:
Adds a `default_controller_name` and uses HTTP method name in absence of `operationId`. See #98

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -14,6 +14,7 @@ Unless required by applicable law or agreed to in writing, software distributed 
 import functools
 import logging
 import os
+import re
 from .decorators.parameter import parameter_to_arg
 from .decorators.produces import BaseSerializer, Produces, Jsonifier
 from .decorators.security import security_passthrough, verify_oauth
@@ -32,7 +33,7 @@ class Operation:
     """
 
     def __init__(self, method, path, operation, app_produces, app_security, security_definitions, definitions,
-                 parameter_definitions, resolver, default_controller_name='', validate_responses=False):
+                 parameter_definitions, resolver, default_module_name='', validate_responses=False):
         """
         This class uses the OperationID identify the module and function that will handle the operation
 
@@ -74,7 +75,7 @@ class Operation:
             'parameters': self.parameter_definitions
         }
         self.validate_responses = validate_responses
-        self.default_router_controller_name = default_controller_name
+        self.default_module_name = default_module_name
         self.operation = operation
 
         self.operation_id = self.detect_controller()
@@ -96,12 +97,24 @@ class Operation:
                 return x_router_controller + '.' + operation_id
             return operation_id
 
+        function = self.method.lower()
+
         if x_router_controller:
             mod_name = x_router_controller
         else:
-            mod_name = self.default_router_controller_name
+            if not self.default_module_name:
+                raise InvalidSpecification(
+                    "Neither operationId or x-swagger-router-controller was "
+                    + "configured and no default module set for Api"
+                )
 
-        return mod_name + '.' + self.method.lower()
+            mod_name = self.default_module_name
+            match = re.search('^/(.+?)(/.?|$)', self.path)
+            if match:
+                mod_name += '.' + match.group(1)
+                function = ('search' if function == 'get' and match.group(2).strip('/') == '' else function)
+
+        return mod_name + '.' + function
 
     def resolve_reference(self, schema):
         schema = schema.copy()  # avoid changing the original schema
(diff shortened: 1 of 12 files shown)


kleijnweb:
Giving this a test drive, revealed that as is using the main module for 'default_controller_name' still has some issues.

connexion/operation.py
```diff
@@ -89,10 +87,34 @@ def __init__(self, method, path, operation, app_produces, app_security,
...
+                raise InvalidSpecification(
+                    "Neither operationId or x-swagger-router-controller was "
+                    + "configured and no default module set for Api"
+                )
+
+            mod_name = self.default_module_name
+            match = re.search('^/(.+?)(/.?|$)', self.path)
+            if match:
+                mod_name += '.' + match.group(1)
+                function = ('search' if function == 'get' and match.group(2).strip('/') == '' else function)
```
hjacobs: That looks like too much "magic" to me --- one could argue that the collection endpoint (e.g. "GET /pets/) is a "list" operation rather than a "search".

hjacobs:
I'm not sure if we should go that way of supporting "operationId-less" semantics.

connexion/operation.py
kleijnweb: Well, that's what you get when you use defaults over configuration. Sure it's arguable, but nobody is stopping no one from configuring an operationId where you want something different, even if that's 'everywhere'. It just becomes *optional* exactly as the Swagger spec requires.

kleijnweb:
@hjacobs Please see the example in #98. Personally I think this is a sensible, RESTy, default.

connexion/operation.py
kleijnweb: @hjacobs What if the resolver returned both the function *and* the operationId? Then the whole algorithm would become pluggable by passing the resolver to `add_api()` as is already supported in master. `detect_controller` would become obsolete (as it would be part of the default resolver), the default, purely `operationId` based algorithm could stay the default, *but* you could easily replace it with this more 'automagic' algorithm. Eg:

```python
app.add_api('swagger.yaml', default_module_name='api', resolver=connexion.resolvers.resty)
```

hjacobs:
@kleijnweb please check your PR, it won't work for Python 3 (see Travis CI logs) as you are using implicit relative imports (default for Python 2).

kleijnweb:
@hjacobs Yeah I noticed that, I set up a virtualenv and tested it locally before pushing again. Should pass now.

kleijnweb:
Okay, so what I did was wrap the default and new logic into `Resolver` classes, which was already more-or-less supported. So the default behavior of connexion is unchanged, but if you want REST-like fallback resolution, you can do:

```python
from connexion.resolver import RestyResolver

app = connexion.App(__name__)
app.add_api('swagger.yaml', resolver=RestyResolver('api'))
```

hjacobs:
OK, sounds good, obviously the resolver semantics changed, i.e. if somebody already used a custom resolver it will break. Maybe we can still support passing a simple callable (a function) to the "resolver" keyword argument?

kleijnweb:
You're right of course: that's a BC break. I can wrap it if the argument is a function. Will get on that shortly.

kleijnweb:
Btw, is it just me or is the current implementation of ''x-swagger-router-controller' pretty pointless?

```json
{
   "operationId": "post_greeting",		
   "'x-swagger-router-controller": "fakeapi.hello"
}
```

I mean, it's somewhat useful in combination with defaulting to HTTP method names, but when you already need to specify the operationId, why bother? Also it takes more effort to ensure the operationId is unique globally, as required by the spec.

I'd just ignore it when you have an operationId and require operationIds be fully quantified.

kleijnweb:
Related to that, there is an inconsistency in `RestyResolver`: when `x-swagger-router-controller` is encountered, it skips the special case for the collection endpoint. Will fix.

hjacobs:
@kleijnweb thanks for all your contributions so far :smile: --- I will check tomorrow (and you wanted to fix something minor?)..

kleijnweb:
Yes, I probably wont get around to till tonight (CET) though. And you're welcome, it's good exercise as Python is not my dayjob but I find myself having to help some sysadmins expose their Python scripts using an API. Enter Connexion.) This feature will give them a little more structure, too.

TBC :)

kleijnweb:
Once you guys have decided to merge this, I'll commence the less glorious task of including a README update ;)

kleijnweb:
Bug squashed.

kleijnweb:
@hjacobs I did a little refactoring. Also it doesn't break on missing starting slashes anymore. Default implementation doesn't either, that should be properly validated somewhere else.

PR #112, opened 2015-12-08
Handle dashes in resource names in RestyResolver

kleijnweb:
Minor improvement which will replace dashes to underscores in the resource name in the URI path.

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -103,19 +103,21 @@ class RestyResolver(Resolver):
         :type operation: connexion.operation.Operation
         """
         path_match = re.search(
-            '^/?(?P<resource_name>(\w(?<!/))*)(?P<trailing_slash>/*)(?P<extended_path>.*)$', operation.path
+            '^/?(?P<resource_name>([\w\-](?<!/))*)(?P<trailing_slash>/*)(?P<extended_path>.*)$', operation.path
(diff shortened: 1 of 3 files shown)


connexion/resolver.py
```diff
@@ -103,19 +103,21 @@ def resolve_operation_id_using_rest_semantics(self, operation):
...
             name = self.default_module_name
+            resource_name = path_match.group('resource_name')
 
             if x_router_controller:
                 name = x_router_controller
 
-            elif path_match.group('resource_name'):
-                name += '.' + path_match.group('resource_name')
+            elif resource_name:
+                resource_controller_name = resource_name.replace('-', '_')
```
hjacobs: I think this should be more generic, i.e. please use something to replace all illegal Python function chars with underscores (otherwise I foresee the next PR for the next single illegal character already :smirk: )

hjacobs: i.e. maybe use a utility function "make_python_identifier" (or similar)

kleijnweb: Is going to be less common, but fair enough. Also applies to the standard operationId resolution, nothing is stopping me from throwing illegal chars in there either. I'll apply it to both.

hjacobs:
OK, OK, you win :smile:

PR #113, opened 2015-12-08
Updated endpoint routing docs

kleijnweb:
Added documentation for endpoint routing customization and the use of x-router-controller.


hjacobs:
:+1:

PR #114, opened 2015-12-10
Use jsonschema

jmcs:


hjacobs:
Could you raise the code coverage? Such an important change should have good test coverage :smile: 

Also we need to test and review it very thoroughly.

jmcs:
The coverage went down because I removed lots of code. The validation decorator is actually 100% tested.

PR #116, opened 2015-12-28
Allow True/False for query parameters

cyprien-g:
I think 'True' / 'False' should be allowed for query parameters. And that would also make swagger client like Bravado compatible with connexion (as it send query parameters with a capital letter).

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -131,9 +131,9 @@ def boolean(s):
     >>> boolean('false')
     False
     '''
-    if s == 'true':
+    if s in ['true', 'True']:
         return True
-    elif s == 'false':
+    elif s in ['false', 'False']:
         return False
     else:
         raise ValueError('Invalid boolean value')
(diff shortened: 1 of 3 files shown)


hjacobs:
@cyprien-g thanks for the pull request :smile: 

Just an idea: if we relax the boolean parsing anyway, what about using lower() to also allow all uppercase values?

cyprien-g:
That's a good idea, I've just made the change.

connexion/utils.py
```diff
@@ -131,9 +131,9 @@ def boolean(s):
     >>> boolean('false')
     False
     '''
-    if s == 'true':
+    if s.lower() == 'true':
```
hjacobs: IMHO we should allow passing None values, i.e. I would check for None to make sure we always get proper ValueError (and not AttributeError).

traxair:
When do you think this will be released @hjacobs ?

hjacobs:
@traxair I'm on it, but I have to get our CD server (which does the release) working again first :smirk:

traxair:
Cool! Looking forward to it :wink:

hjacobs:
done: https://github.com/zalando/connexion/releases/tag/1.0.29

PR #117, opened 2015-12-28
Support for default fields in operation parameters

arjunrn:
In this PR I have implemented support for defaults for query and body parameters. Support for header parameters is not implemented because headers and not treated as arguments to functions. If this changes in the future then adding the support is not difficult. Finally there is no support for _default_ in individual fields of object types. I think this is not directly related to the parameters it can be implemented in another PR.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -59,6 +59,8 @@ def parameter_to_arg(parameters, function):
     query_types = {parameter['name']: parameter
                    for parameter in parameters if parameter['in'] == 'query'}  # type: dict[str, str]
     arguments = get_function_arguments(function)
+    default_query_params = {param['name']: param['default'] for param in parameters if param['in'] == 'query'
+                            and 'default' in param}
 
     @functools.wraps(function)
     def wrapper(*args, **kwargs):
@@ -67,6 +69,9 @@ def parameter_to_arg(parameters, function):
         try:
             request_body = flask.request.json
         except exceptions.BadRequest:
+            request_body = None
+
+        if default_body and not request_body:
             request_body = default_body
 
         # Add body parameters
@@ -78,7 +83,7 @@ def parameter_to_arg(parameters, function):
                 kwargs[body_name] = request_body
 
         # Add query parameters
-        for key, value in flask.request.args.items():
+        for key, value in list(flask.request.args.items()) + list(default_query_params.items()):
             if key not in arguments:
                 logger.debug("Query Parameter '%s' not in function arguments", key)
             else:
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -50,11 +50,12 @@ def parameter_to_arg(parameters, function):
     Pass query and body parameters as keyword arguments to handler function.
 
     See (https://github.com/zalando/connexion/issues/59)
-    :type body_schema: dict|None
+    :type parameters: dict|None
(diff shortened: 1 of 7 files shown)


hjacobs:
@arjunrn thanks for the pull request, we will review it :smile:

connexion/decorators/parameter.py
```diff
@@ -50,14 +50,17 @@ def parameter_to_arg(parameters, function):
     Pass query and body parameters as keyword arguments to handler function.
 
     See (https://github.com/zalando/connexion/issues/59)
-    :type body_schema: dict|None
+    :type parameters: dict|None
     :type parameters: dict|None
```
jmcs: This line is repeated

connexion/decorators/parameter.py
```diff
@@ -77,14 +83,14 @@ def wrapper(*args, **kwargs):
                 kwargs[body_name] = request_body
 
         # Add query parameters
-        for key, value in flask.request.args.items():
+        for key, value in list(flask.request.args.items()) + list(default_query_params.items()):
```
jmcs: You can create a set from two `dict_items` with `flask.request.args.items() | default_query_params.items()`.
It would also be cleaner and cleaner if you created a variable before, for example, `query_parameters = flask.request.args.items() | default_query_params.items()`

connexion/decorators/parameter.py
```diff
@@ -77,14 +83,14 @@ def wrapper(*args, **kwargs):
...
             if key not in arguments:
                 logger.debug("Query Parameter '%s' not in function arguments", key)
             else:
                 logger.debug("Query Parameter '%s' in function arguments", key)
                 query_param = query_types[key]
                 logger.debug('%s is a %s', key, query_param)
-                kwargs[key] = get_val_from_param(value, query_param)
+                kwargs[key] = get_val_from_param(value, query_param) or query_param.get('default')
```
jmcs: If the value is "falsy" (like `0`) it would be replaced with the default value.

connexion/decorators/parameter.py
arjunrn: Fixed this.

connexion/decorators/parameter.py
arjunrn: When I looked at it more closely I saw there is a problem with which value is used. I have fixed this issue by using a python `dict` and then `update()`ing it. Seemed a better solution on the whole.

connexion/decorators/parameter.py
arjunrn: Fixed this issue and added a test for it.

hjacobs:
Released in https://github.com/zalando/connexion/releases/tag/1.0.31

@arjunrn cool, I changed my example to use the default param, see https://github.com/hjacobs/connexion-example

pkjg:
In my case, the default values dont work. Here is swagger code snippet

```
 parameters:
        - name: limit
          in: query
          description: some description
          required: false
          type: integer
          format: int32
          default: 10
```

in the controller, the **request.args** gets **None** if no value is given while calling the endpoint.

rafaelcaricio:
@pkjg `required: false` change to `required: true` and it should work fine.

pkjg:
@rafaelcaricio Thanks for the quick response. I thought whats "required" doesn't need default value. The value must come with request. 
on the other hand, whats optional should have some default value if not provided in the request. Curious

PR #121, opened 2016-01-13
fix bug in restyresolver which was causing collision between .search and .get methods

dfeinzeig:


dfeinzeig:
Test passed.  Unless you have a preference on .upper() == 'GET' or .lower() == 'get' I think this is good to go.

hjacobs:
:+1: thanks

PR #125, opened 2016-01-30
RestyResolver Example

kleijnweb:


examples/restyresolver/api/pets.py
```diff
@@ -0,0 +1,44 @@
...
+
+pets = {}
+
+
+def post(pet):
+    count = len(pets)
+    pet['id'] = count + 1
+    pet['registered'] = rfc3339(time.time())
+    pets[pet['id']] = pet
+    return Response(json.dumps(pet), status=201, mimetype='application/json')
```
hjacobs: any reason you are manually doing JSON serialization?

I would expect to just use `return pet`

kleijnweb: Ah, I can do "return pet, 201", no?

hjacobs: yes, exactly, see https://github.com/zalando/connexion/blob/master/connexion/decorators/decorator.py#L41

hjacobs:
:+1: thanks!

PR #126, opened 2016-02-01
Added support for circular references and additionalProperties

jbryan:
Connexion doesn't currently support references in many places explicitly supported by the spec.  A reference in a typed map (e.g. the example under Model with Map/Dictionary Properties at http://swagger.io/specification/) would fail.  Also, circular references, like those used in a tree structure, would also fail.  Though spec doesn't specifically claim to support them, this [bug](https://github.com/swagger-api/swagger-codegen/issues/728) seems to suggest that most implementations should.  

The solution attached leaves most of the reference resolution up to jsonschema, and just copies the definitions as necessary to individual schema objects.

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -146,6 +146,16 @@ class Operation:
         return schema
 
     def check_references(self, schema):
+        """
+        Searches the keys and values of a schema object for json references.
+        If it finds one, it attempts to locate it and will thrown an exception
+        if the reference can't be found in the definitions dictionary.
+
+        :param schema: The schema object to check
+        :type schema: dict
+        :raises InvalidSpecification: raised when a reference isn't found
+        """
+
         stack = [schema]
         visited = set()
         while stack:
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -150,7 +150,7 @@ class Operation:
         visited = set()
         while stack:
             schema = stack.pop()
-            for k,v in schema.items():
+            for k, v in schema.items():
                 if k == "$ref":
                     if v in visited:
                         continue
@@ -162,7 +162,6 @@ class Operation:
                             stack.append(item)
                 elif hasattr(v, "items"):
                     stack.append(v)
-
 
     def _retrieve_reference(self, reference):
         if not reference.startswith('#/'):
@@ -184,8 +183,6 @@ class Operation:
                 definition_name=definition_name, method=self.method, path=self.path))
 
         return definition
-
-
 
     def get_mimetype(self):
         if produces_json(self.produces):  # endpoint will return json
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -123,43 +123,69 @@ class Operation:
 
     def resolve_reference(self, schema):
         schema = deepcopy(schema)  # avoid changing the original schema
-        # find the object we need to resolve/update
+        self.check_references(schema)
+
+        # find the object we need to resolve/update if this is not a proper SchemaObject
+        # e.g a response or parameter object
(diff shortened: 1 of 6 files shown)


hjacobs:
@jbryan thanks for the PR, it will probably take some time to check it :smile:

hjacobs:
LGTM, but I would give it some more time to review (takes me always some time to get back into this schema stuff :smirk: )

@jmcs @rafaelcaricio maybe you can also help reviewing this...

connexion/operation.py
```diff
@@ -123,44 +123,67 @@ def validate_defaults(self):
...
+
+        # if there is a schema object on this param or response, then we just
+        # need to include the defs and it can be validated by jsonschema
+        if 'schema' in schema:
+            schema['schema']['definitions'] = self.definitions
+            return schema
+
         return schema
 
+    def check_references(self, schema):
```
hjacobs: What is this method doing? It's not immediately clear, the reader would have to see the call to `_retrieve_reference` which can throw exceptions... --- a docstring would help :smile:

hjacobs:
I guess we need some more unit tests to really test the schema logic itself (i.e. not the request processing etc)...

connexion/operation.py
```diff
@@ -123,44 +123,67 @@ def validate_defaults(self):
...
+        stack = [schema]
+        visited = set()
+        while stack:
+            schema = stack.pop()
+            for k, v in schema.items():
+                if k == "$ref":
+                    if v in visited:
+                        continue
+                    visited.add(v)
+                    stack.append(self._retrieve_reference(v))
```
jmcs: Besides the docstring I would also comment here that this can raise an exception

tests/test_app.py
```diff
@@ -397,6 +397,79 @@ def test_schema_list(app):
...
+                                  data=json.dumps(invalid_object))  # type: flask.Response
+    assert wrong_items.status_code == 400
+    assert wrong_items.content_type == 'application/problem+json'
+    wrong_items_response = json.loads(wrong_items.data.decode())  # type: dict
+    assert wrong_items_response['title'] == 'Bad Request'
+    assert wrong_items_response['detail'].startswith("42 is not of type 'object'")
+
+    right_type = app_client.post('/v1.0/test_schema_recursive', headers=headers,
+                                  data=json.dumps(valid_object))  # type: flask.Response
+    print(right_type.data.decode())
```
rafaelcaricio: Please remove this `print`.

connexion/operation.py
jbryan: Fair point, I can add docs ... I was actually am of a split mind on whether I should include it ... on the one hand it will check most references and thus preserve the current behavior of failing on load ... on the other hand, jsonschema will check refs when it validates, and will give a somewhat more meaningful message.

connexion/operation.py
jbryan: Added a comment in 11fa3a64a357.

hjacobs:
LGTM

JohnGalt1717:
This still hasn't been fixed in Swagger-UI. It still crashes the browser in an endless loop on cicular references which happen all of the time in any basic application that has a structure like this:

Organization
--> Users

Your users Dtos are going to reference the organization and the Organization is going to reference the users.

jbryan:
@JohnGalt1717, this PR was to specifically solve the issue in the connexxion swagger parser ... the issues in swagger-ui should probably be fixed upstream and pulled in. It appears there is at least one open issue related to it https://github.com/swagger-api/swagger-ui/issues/1646, but seems to be slightly different then the scenario you are describing.

PR #129, opened 2016-02-01
Fix #128 - document `x-swagger-router-controller`

### LINKED ISSUE #128: Documentation - relative routing with router controller
The endpoint routing section incorrectly states that the `x-router-controller` property may be set for an operation to make the operationId property relative. However, the correct property name is `x-swagger-router-controller` (see [connexion.resovler:60](https://github.com/zalando/connexion/blob/master/connexion/resolver.py#L60)).

milo-minderbinder:
Minor documentation fix for issue #128. Documentation incorrectly referred to the operation property `x-router-controller`, which should actually be `x-swagger-router-controller`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ For example:
           operationId: myapp.api.hello_world
 
 If you provided this path in your specification POST requests to ``http://MYHOST/hello_world`` would be handled by the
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1: thanks

Issue #127, opened 2016-02-01
Issue with Boolean types in code generated by swagger-codegen


scolby33:
Here is a minimal (non-)working example needed to reproduce this issue.

test.yml:
```
---
swagger: '2.0'

info:
    version: '0'
    title: 'Boolean Test'
    
paths:
    /test:
        get:
            description: A test endpoint.
            parameters:
                - name: truthiness
                  in: query
                  description: A test parameter.
                  type: boolean
                  default: true
            responses:
                204:
                    description: Do some magic.
```

Process this file via swagger-codegen:

`java -jar /usr/local/Cellar/swagger-codegen/2.1.5/libexec/swagger-codegen-cli.jar generate -i ./test.yml -l python-flask -o .`

Here is the resulting swagger.yml:
```
---
swagger: "2.0"
info:
  version: "0"
  title: "Boolean Test"
paths:
  /test:
    get:
      tags:
      - "default_controller"
      description: "A test endpoint."
      operationId: "controllers.default_controller.test_get"
      parameters:
      - name: "truthiness"
        in: "query"
        description: "A test parameter."
        required: false
        type: "boolean"
        default: true
      responses:
        204:
          description: "Do some magic."
definitions: {}
```

Running this with `python app.py` results in the following error:

```
Traceback (most recent call last):
  File "~/.virtualenvs/swagger/lib/python3.5/site-packages/connexion/decorators/validation.py", line 75, in validate_type
    return make_type(value, param_type)
  File "~/.virtualenvs/swagger/lib/python3.5/site-packages/connexion/decorators/validation.py", line 53, in make_type
    return type_func(value)
  File "~/.virtualenvs/swagger/lib/python3.5/site-packages/connexion/utils.py", line 135, in boolean
    raise ValueError('Invalid boolean value')
ValueError: Invalid boolean value

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/.virtualenvs/swagger/lib/python3.5/site-packages/connexion/operation.py", line 118, in validate_defaults
    validation.validate_type(param, param['default'], 'query', param['name'])
  File "~/.virtualenvs/swagger/lib/python3.5/site-packages/connexion/decorators/validation.py", line 77, in validate_type
    raise TypeValidationError(param_type, parameter_type, parameter_name)
connexion.decorators.validation.TypeValidationError: Wrong type, expected 'boolean' for query parameter 'truthiness'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #130, opened 2016-02-01
Handle correctly boolean values in default property

rafaelcaricio:
This PR fixes the issue #127. 

[Here](http://www.yaml.org/spec/1.2/spec.html#id2803629) in the spec, it seems that Booleans can be true or false, without quotes.

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -131,7 +131,9 @@ def boolean(s):
     >>> boolean('false')
     False
     '''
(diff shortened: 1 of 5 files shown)


hjacobs:
@rafaelcaricio can you add a unit test for that case? (boolean in Swagger YAML)

jmcs:
:+1:

rafaelcaricio:
@hjacobs added.

PR #140, opened 2016-02-07
Use the JSON/Swagger aware boolean constructor for constructing boole…

### LINKED ISSUE #139: All Boolean values passed in via query strings evaluate to True
I have an issue with an endpoint defined like so:
```YAML
paths:
  /locations:
    get:
      tags:
        - locations
      summary: List locations
      operationId: location_controllers.locations_get
      parameters:
        - name: per_page
          in: query
          description: The maximum number of entries returned. Default `10` max `200`.
          type: integer
          minimum: 1
          maximum: 200
          default: 10
        - name: page
          in: query
          description: The page to return. Default `1`.
          type: integer
          default: 1
        - name: reverse_chronological
          in: query
          description: Return the results in reverse chronological order (newest first) if `true`. In chronological order if `false`. Default `true`.
          type: boolean
          default: false
      responses:
        '204':
          description: Success.
```
(slightly edited to only include salient parts)

The test controller function I've been using goes like this:
```Python
def locations_get(per_page=10, page=1, reverse_chronological=False) -> str:
    if reverse_chronological:
        print('reverse')
```

This request doesn't print `reverse` to the console:
```HTTP
GET /locations HTTP/1.1
```

And this one does
```HTTP
GET /locations?reverse_chronological=True HTTP/1.1
```

So far, so good. But this request *also* prints `reverse`:
```HTTP
GET /locations?reverse_chronological=False HTTP/1.1
```

After several hours of pulling my hair out and single-stepping through the request code, I believe I've found the issue, which lies in `connexion.decorators.parameter`. I've pasted the relevant bits of code here.
```Python
        # Add query parameters
        query_arguments = copy.deepcopy(default_query_params)
        query_arguments.update(flask.request.args.items())
        for key, value in query_arguments.items():
            if key not in arguments:
                logger.debug("Query Parameter '%s' not in function arguments", key)
            else:
                logger.debug("Query Parameter '%s' in function arguments", key)
                query_param = query_types[key]
                logger.debug('%s is a %s', key, query_param)
                kwargs[key] = get_val_from_param(value, query_param)
```
```Python
def get_val_from_param(value, query_param):
    if query_param["type"] == "array":  # then logic is more complex
        if query_param.get("collectionFormat") and query_param.get("collectionFormat") == "pipes":
            parts = value.split("|")
        else:  # default: csv
            parts = value.split(",")
        return [make_type(part, query_param["items"]["type"]) for part in parts]
    else:
        return make_type(value, query_param["type"])
```
```Python
def make_type(value, type):
    type_func = TYPE_MAP[type]  # convert value to right type
    return type_func(value)
```
```Python
# https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
TYPE_MAP = {'integer': int,
            'number': float,
            'string': str,
            'boolean': bool,
            'array': list,
            'object': dict}  # map of swagger types to python types
```

In the first snippet, everything is cool until the line `kwargs[key] = get_val_from_param(value, query_param)`. The value is stored as a string ("True" or "False") up until this point, and here you mean to cast it to a python `bool`. `get_val_from_param()` is called, which calls `make_type("False", "boolean")`, in this case.  `make_type()` uses the default Boolean constructor: `bool("False")`. [But, in Python, all strings are `True`!](https://docs.python.org/3/library/stdtypes.html#truth) So, we get `bool(True)` returned and everything goes wrong from there.

You have a utility function `connexion.utils.boolean` that takes care of this for you. Using that in place of the standard `bool()` constructor seems to solve the problem.

scolby33:
…an values from query parameters.

Should close #139.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -6,13 +6,15 @@ import inspect
 import logging
 import six
 
+from ..utils import boolean
+
 logger = logging.getLogger(__name__)
 
 # https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
 TYPE_MAP = {'integer': int,
             'number': float,
             'string': str,
-            'boolean': bool,
+            'boolean': boolean,
             'array': list,
             'object': dict}  # map of swagger types to python types
 
--- a/tests/fakeapi/api.yaml
+++ b/tests/fakeapi/api.yaml
@@ -684,6 +684,17 @@ paths:
           in: query
           default: false
 
+  /test-bool-array-param:
+    get:
+      summary: Test usage of an array of booleans value
+      operationId: fakeapi.hello.test_bool_array_param
+      parameters:
+        - name: thruthiness
+          in: query
+          type: array
+          items:
+            type: boolean
+
   /test-required-param:
     get:
       summary: Test required param without default value
@@ -724,7 +735,7 @@ definitions:
     properties:
       children:
         type: array
-        items: 
+        items:
           $ref: "#/definitions/simple_tree"
         description: Docker image version to deploy
     additionalProperties: false
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -238,6 +238,10 @@ def test_bool_default_param(thruthiness):
     return thruthiness
 
 
+def test_bool_array_param(thruthiness):
+    return all(thruthiness)
+
+
 def test_required_param(simple):
     return simple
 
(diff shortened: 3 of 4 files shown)


scolby33:
This should also fix a similar (and hypothetical) issue with Booleans in arrays passed in through the query string. I tested for neither the existence of this issue nor the efficacy of the fix for it. Someone may want to give that a shot before merging.

hjacobs:
LGTM, thanks for the PR!

hjacobs:
@scolby33 could you add a unit test to reproduce the problem (and test the fix)?

scolby33:
Added tests. Confirmed that before the fix, the tests failed (always returning true values) and that the fix mitigates the issue.

PR #143, opened 2016-02-08
Auth 404 errors option

rafaelcaricio:
Add option to authenticate even when the path is not found. That prevents some scripts to scan for valid paths.


connexion/handlers.py
```diff
@@ -0,0 +1,27 @@
...
+        """
+        security_decorator = self.security_decorator
+        logger.debug('... Adding security decorator (%r)', security_decorator, extra=vars(self))
+        return security_decorator(self.handle)
+
+    def handle(self, *args, **kwargs):
+        """
+        Actual handler for the execution after authentication.
+        """
+        return '', self.status_code
```
jmcs: This should return a [problem object](https://github.com/zalando/connexion/blob/master/connexion/problem.py#L17) to be consistent with the rest of the framework.

PR #147, opened 2016-02-13
#146 allow using TOKENINFO_URL env var (OAuth2 token info)

### LINKED ISSUE #146: Use "TOKENINFO_URL" environment variable as fallback
We decided to standardize the OAuth 2 token info URL configuration: we are now using the `TOKENINFO_URL` environment variable which should be supported by Connexion.

See also https://github.com/zalando-stups/taupage/issues/177

hjacobs:
Fixes #146

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -16,6 +16,7 @@ Unless required by applicable law or agreed to in writing, software distributed 
 from flask import request
 import functools
 import logging
+import os
 import requests
 from ..problem import problem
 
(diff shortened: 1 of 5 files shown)


connexion/decorators/security.py
```diff
@@ -28,6 +29,20 @@
...
+def get_tokeninfo_url(security_definition):
+    '''
+    :type security_definition: dict
+    :rtype: str
+
+    >>> get_tokeninfo_url({'x-tokenInfoUrl': 'foo'})
+    'foo'
+    '''
+    token_info_url = security_definition.get('x-tokenInfoUrl') or \
+        os.environ.get('HTTP_TOKENINFO_URL') or \
```
rafaelcaricio: Maybe we should log a warning message notifying the user -- if the `HTTP_TOKENINFO_URL` is set -- that they should use `TOKENINFO_URL`?

hjacobs: Probably a good idea, so we can deprecate it in some later version.

hjacobs:
@rafaelcaricio updated with warning log :smile:

rafaelcaricio:
Great! LGTM! :smile:

PR #149, opened 2016-02-15
Recommend usage of `decorator` package in Connexion projects

### LINKED ISSUE #142: get_function_arguments() behavior changes between python 2.7 and python 3.5
Function get_function_parameters from connexion/decorators/parameters.py changes its behavior from python 2.7 to python 3.5:

```python
def get_function_arguments(function):  # pragma: no cover
    """
    Returns the list of arguments of a function
    :type function: Callable
    :rtype: list[str]
    """
    if six.PY3:
        return list(inspect.signature(function).parameters)
    else:
        return inspect.getargspec(function).args
```

If the parameter function is a decorated function (i.e basicauth example) with parameters in python 2.7 the result is always an empty list while in python 3.5 the result is the list with the actual parameters of the decorated function.

rafaelcaricio:
## The problem

The `functools.wraps` does not pass the information about parameters of the original function to the decorated instance. This makes Connexion fail to get the argument list defined in the original function. So in case you create a decorator with `functools.wraps` and use in your Connexion view function with parameters it will fail on **Python 2.x**.

## Solution

Due to that problem we should advise the users of Connexion to use the [`decorator`](https://github.com/micheles/decorator) package while creating decorators for their Connexion view functions. This will keep Connextion working as expected specially in case of handing endpoints with parameters in decorated view functions.

## What was changed in this PR

Updates our `/basicauth` example project to use the `decorator` package so it drives the users to the recommended usage. Also a test was added to make sure that Connexion will keep working if followed the recommendation.

Fixes #142

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ htmlcov/
 .eggs
 .cache/
 *.swp
+.tox/
--- a/examples/basicauth/app.py
+++ b/examples/basicauth/app.py
@@ -3,11 +3,21 @@
 Connexion HTTP Basic Auth example
 
 Most of the code stolen from http://flask.pocoo.org/snippets/8/
+
+Warning: It is recommended to use 'decorator' package to create decorators for
+         your view functions to keep Connexion working as expected. For more
+         details please check: https://github.com/zalando/connexion/issues/142
 '''
 
 import connexion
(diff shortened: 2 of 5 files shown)


hjacobs:
:+1:

PR #152, opened 2016-02-16
Support parameters definitions in path level

rafaelcaricio:
OpenAPI Specification allow [parameters definitions in path level](http://swagger.io/specification/#pathItemObject) of the API specification, but connexion did not support that. This PR adds support for that in Connexion.

### Additional changes
More clear error message when there is no parameter defined for a endpoint and the function expects an argument. Before user was just seeing a `TypeError` and it was very obscure to identify what was going on underneath.

Related to #48

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -102,7 +102,7 @@ class Api:
         if auth_all_paths:
             self.add_auth_on_not_found()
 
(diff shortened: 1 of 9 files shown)


jmcs:
Can you fix the flake8 error?

rafaelcaricio:
@jmcs Done. :+1:

PR #157, opened 2016-02-22
Update README.rst

LappleApple:
Rafael and I chatted on Friday and I've included these changes as per his input -- please check for errors and improper phrasings (a weekend away from my notes => potential of forgetting some context)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,14 +25,47 @@ Connexion
    :target: https://github.com/zalando/connexion/blob/master/LICENSE
    :alt: License
 
-Connexion is a framework on top of Flask_ to automagically handle your REST API requests
-based on `OpenAPI 2.0 Specification`_ (formerly known as Swagger Spec) files
-in YAML.
+Connexion is a Flask_ application/framework that automagically handles your REST API requests based on `OpenAPI 2.0 Specification`_ (formerly known as Swagger Spec) files in YAML. Written in Python, Connexion enables you to write your Swagger specification, then maps the specification to Python functions. This is what makes it unique from other tools. We built Connexion this way in order to:
+- simplify the development process
+- reduces misinterpretation about what an API is going to look like
+- save time
 
-How to use
+Connexion Features:
 ----------
+- validates requests and endoint parameters automatically, based on your specification  
+- provides the Swagger Console UI so that you can call your API through it
+- handles OAuth 2
+- supports API versioning
+- supports automatic serializization. If your specification defines that an endpoint returns JSON, Connexion will automatically serialize the return value for you and set the right content type in the HTTP header.
 
-Put your API YAML inside a folder in the root path of your application (e.g ``swagger/``) and then do
+Why Connexion
+----------
+With Connexion, you write the spec first. It then calls your Python code—handling the mapping from the specification to the code. This incentivizes you to write the specification so that all of your developers can understand what your API does—even before you write a single line of code. If multiple teams depend on your APIs, you can already send them documentation of your API. This guarantees that your API will follow the specification that you wrote.  
+it’s not like Hug_, which generates the specification based on what code you write. 
+
+Other Sources/Mentions
+----------
+- Zalando Tech blog post `API First`_
+- We're honored that Connexion is Swagger_'s official framework of choice for Python.
+
+How to Use
+=========
+Prerequisites
+----------
+Python 2.7 - most recent version (3.5)
+
+Installing It
+----------
+In your command line, type this:
+
+.. code-block:: python
+$ pip install Connexion
+$ python hello.py
+ * Running on http://localhost:5000/
+ 
+Running It
+----------
+Place your API YAML inside a folder in the root path of your application (e.g ``swagger/``). Then run:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -84,8 +119,8 @@ function ``hello_world`` in ``myapp.api``. Optionally you can include ``x-swagge
           x-swagger-router-controller: myapp.api
           operationId: hello_world
 
-
-To customize this behavior, Connexion can use alternative ``Resolvers``, for example ``RestyResolver``. The ``RestyResolver`` will compose an ``operationId`` based on the path and HTTP method of the endpoints in your specification:
+**Automatic Routing**
```
jmcs: You need `:` and a empty line here, or maybe this should a subsection ([with `~~~~~~~~~` below the title](http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections))

README.rst
```diff
@@ -188,57 +229,49 @@ to ``tornado``:
...
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-http://www.apache.org/licenses/LICENSE-2.0
+TODOs
+--------------
+If you'd like to become a more consistent contributor to Connexion, we'd love your help working on these:
+- Additional ways to handle OAuth 2 authentications
+- Silencing ImportErrors raised in modules referenced by operationId
+- Overriding default validation error message
```
jmcs: Maybe you can also include the link to the issues page or even the [waffle board](https://waffle.io/zalando/lizzy).

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
    :target: https://github.com/zalando/connexion/blob/master/LICENSE
    :alt: License
 
-Connexion is a framework on top of Flask_ to automagically handle your REST API requests
-based on `OpenAPI 2.0 Specification`_ (formerly known as Swagger Spec) files
-in YAML.
+Connexion is a Flask_ application/framework that automagically handles your REST API requests based on `OpenAPI 2.0 Specification`_ (formerly known as Swagger Spec) files in YAML. Written in Python, Connexion enables you to write your Swagger specification, then maps the specification to Python functions. This is what makes it unique from other tools. We built Connexion this way in order to:
```
rafaelcaricio: I would just say it is a "framework" because it cannot handle automagically your REST API requests without you actually writing some code.

rafaelcaricio: I would add like that "files in YAML format." just to be explicit about what it is about and not to mislead the reader to think YAML could be something else. Maybe even a link to the spec would be interesting https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#format

rafaelcaricio: I would remove the "Written in Python," part since it is mentioned twice.

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
+- simplify the development process
```
rafaelcaricio: We need a newline before it so it will be formatted as bullet lists.

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
+- reduces misinterpretation about what an API is going to look like
+- save time
 
-How to use
+Connexion Features:
 ----------
+- validates requests and endoint parameters automatically, based on your specification  
+- provides the Swagger Console UI so that you can call your API through it
```
rafaelcaricio: "Provides a Swagger Console UI..."

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
+- handles OAuth 2
```
rafaelcaricio: Handles OAuth 2 token based.

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
-Put your API YAML inside a folder in the root path of your application (e.g ``swagger/``) and then do
+Why Connexion
+----------
+With Connexion, you write the spec first. It then calls your Python code—handling the mapping from the specification to the code. This incentivizes you to write the specification so that all of your developers can understand what your API does—even before you write a single line of code. If multiple teams depend on your APIs, you can already send them documentation of your API. This guarantees that your API will follow the specification that you wrote.  
+it’s not like Hug_, which generates the specification based on what code you write. 
+
+Other Sources/Mentions
+----------
+- Zalando Tech blog post `API First`_
+- We're honored that Connexion is Swagger_'s official framework of choice for Python.
```
rafaelcaricio: I think this is a bit misleading. Connexion is one of the Python frameworks listed in Swagger website. But it is not the official framework of choice.

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
+
+How to Use
+=========
+Prerequisites
+----------
+Python 2.7 - most recent version (3.5)
```
rafaelcaricio: I would just add as:

"Python 2.7 up to 3.x"

README.rst
```diff
@@ -25,14 +25,47 @@ Connexion
...
+
+Installing It
+----------
+In your command line, type this:
+
+.. code-block:: python
```
rafaelcaricio: It should be `bash` instead of `python` and should add a newline after that.

README.rst
jmcs: I don't want to support python 3.{0..3}

jmcs: So it should be something like `Python 2.7 or Python 3.4+`

rafaelcaricio: :+1:

README.rst
LappleApple: I'm adding "token-based authentication" as I think there's a word (authentication) missing here

rafaelcaricio:
:+1:

PR #159, opened 2016-02-23
App breaks if error adding operation raise exception

### LINKED ISSUE #155: Silencing ImportErrors raised in modules referenced by operationId
To reproduce simply create an incorrect import that will raise an ImportError in a module that is referenced by operationId of an endpoint specified in a swagger API.

```
project/swagger/api.yaml
---------------------------------
...
paths:
    /scms:
         get:
             operationId:
                  api.scms.test
...



project/src/api/scms.py
---------------------------------
#broken import
import something_broken

def test():
    return {'test': 'ok'}, 200

```

When URL `/api/scms` is requested the module raises an `ImportError` and the test method is not reachable, but connexion proceeds and it says:
```{"status": 404, "type": "about:blank", "detail": "The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.", "title": "Not Found"}```
without re-raising the exception so it is not clear what failed.

rafaelcaricio:
Fixes #155.

Question is should we change that behaviour in Connexion? It is a breaking change since older versions would keep going even if the `operationId` is invalid.

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -150,11 +150,7 @@ class Api:
             for method, endpoint in methods.items():
                 if method == 'parameters':
                     continue
-                try:
-                    self.add_operation(method, path, endpoint, path_parameters)
-                except Exception:  # pylint: disable= W0703
-                    logger.exception('Failed to add operation for %s %s%s',
-                                     method.upper(), self.base_url, path)
+                self.add_operation(method, path, endpoint, path_parameters)
 
     def add_auth_on_not_found(self):
         """
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -156,10 +156,13 @@ class Api:
                 try:
                     self.add_operation(method, path, endpoint, path_parameters)
                 except Exception:  # pylint: disable= W0703
-                    logger.exception('Failed to add operation for %s %s%s',
-                                     method.upper(), self.base_url, path)
-                    if not self.debug:
+                    error_msg = 'Failed to add operation for {} {}{}'.format(
+                        method.upper(), self.base_url, path)
+                    if self.debug:
+                        logger.exception(error_msg)
+                    else:
                         import sys
+                        logger.error(error_msg)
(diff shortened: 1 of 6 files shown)


jmcs:
The current behavior was added (I think by @hjacobs) to allow a partial implementation of the API during development.

rafaelcaricio:
I think it actually hurts the development. As an example @jkozlowicz lost around 2 hours trying to figure out what was going on.

I think maybe we should add a flag to explicitly force the application to load even if some endpoint are not going to be available.

hjacobs:
@rafaelcaricio obviously we need to output errors, but I like a more iterative development approach where "something" is running even if not everything is implemented yet. We could also tie the behavior to the `debug` flag which also prints errors/stacktraces.

rafaelcaricio:
I like the idea of letting the `debug=True` allow the application to load even if there is endpoints not completely implemented. And only break if the `debug=False` is set. Sounds good then? If so, I will move forward and implement that change.

jkozlowicz:
Well I would like to be able to see what, where and why fails. Being able to do that by turning on `debug` is a solution. But anyway I would simply raise that `ImportError` in this case.

@hjacobs I don't think that in this case 

> "something" is running even if not everything is implemented yet

but it is rather "something" is running even if something is broken, because an `ImportError` was raised. You can reference an empty method by operationId and it will work even though it is not implemented. That would be "something" is running even if not everything is implemented yet. But if the module that contains a method raises `ImportError` that's an error and why should the rest of the  application proceed? 

That is my opinion, but well I don't know what other projects depend on connexion and what outcomes could arise from changing this here. It is just the most reasonable solution to me.

rafaelcaricio:
@hjacobs please check now if that fulfil your use case.

rafaelcaricio:
@hjacobs @jmcs please review. I made the requested changes.

jmcs:
:+1:

hjacobs:
:+1:

PR #160, opened 2016-02-25
Only support valid OpenAPI/Swagger 2.0 specifications

rafaelcaricio:
I have added the package [`swagger_spec_validation`](https://github.com/Yelp/swagger_spec_validator/) because we were supporting some invalid swagger specification definitions so that change will guarantee that we will only keep supporting valid definitions. Along with that I have made the required changes in the code to fix how we handled the specification (mostly related to `body` parameters).

Changes:
- Only supports valid OpenAPI/Swagger 2.0 specifications
- Support correctly the body parameter. Before we were taking then default value from the parameter level, but since parameters of type "body" require schema the [default value should come from the schema definition object](http://swagger.io/specification/#parameterObject).
- Split the file `tests/test_app.py` in several smaller contextual test files under `tests/api/test_*.py` so it makes easier to find and extend our tests with more contextualisation on what they aim to cover.


connexion/operation.py
```diff
@@ -169,28 +168,12 @@ def __init__(self, method, path, path_parameters, operation, app_produces,
...
-                break
-        else:
-            self.default_body = None
-
         self.validate_defaults()
 
     def validate_defaults(self):
         for param in self.parameters:
             try:
-                if param['in'] == 'body' and 'default' in param:
```
rafaelcaricio: This validation is handled by the "swagger_spec_validation" pkg, we do not need to do it anymore.

connexion/operation.py
```diff
@@ -92,9 +91,6 @@ def security_decorator(self):
                 logger.debug(
                     "... Security type '%s' not natively supported by Connexion; you should handle it yourself",
                     security_definition['type'], extra=vars(self))
-            else:
-                logger.warning("... Security type '%s' unknown. **IGNORING SECURITY REQUIREMENTS**",
-                               security_definition['type'], extra=vars(self))
```
rafaelcaricio: An invalid security definition in the swagger specification will throw now an error while validating against swagger_spec_validator so this code won't be reachable.

jmcs:
:+1:

PR #162, opened 2016-02-29
make api.add_paths() exception handler work in python 2.7

TNGPS:
The re-raising of the exception in this version fails in python 2.7raising an AttributeError since  "with_traceback" is know known.
using the ` ei.__traceback__ = tb` and than just raising `ie` again works fine in Python 2.7.
Note: I did not test this in Python 3, but I assume this should work just fine there to !

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -170,7 +170,7 @@ class Api:
                             raise et, "Fatal error found while addding '{}': {}".format(
                                         endpoint.get('operationId'),
                                         str(ei)), tb
-                                        
+
     def add_auth_on_not_found(self):
         """
         Adds a 404 error handler to authenticate and only expose the 404 status if the security validation pass.
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -165,12 +165,10 @@ class Api:
                         import sys
                         import six
                         et, ei, tb = sys.exc_info()
-                        if six.PY3:
-                            raise ei.with_traceback(tb)
-                        else:
-                            raise et, "Fatal error found while addding '{}': {}".format(
+                        message = "Fatal error found while addding '{}': {}".format(
                                         endpoint.get('operationId'),
-                                        str(ei)), tb
+                                        str(ei))
+                        six.reraise(et, message, exc_traceback=tb)
 
     def add_auth_on_not_found(self):
         """
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -168,7 +168,7 @@ class Api:
                         message = "Fatal error found while addding '{}': {}".format(
                                         endpoint.get('operationId'),
                                         str(ei))
-                        six.reraise(et, message, exc_traceback=tb)
+                        six.reraise(et, message, tb)
 
     def add_auth_on_not_found(self):
         """
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -169,7 +169,8 @@ class Api:
                         else:
                             raise et, "Fatal error found while addding '{}': {}".format(
                                         endpoint.get('operationId'),
-                                        str(ei) ), tb
+                                        str(ei)), tb
+                                        
     def add_auth_on_not_found(self):
         """
         Adds a 404 error handler to authenticate and only expose the 404 status if the security validation pass.
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -161,12 +161,15 @@ class Api:
                     if self.debug:
                         logger.exception(error_msg)
                     else:
-                        import sys
                         logger.error(error_msg)
+                        import sys
                         et, ei, tb = sys.exc_info()
-                        ei.__traceback__ = tb
-                        raise ei
-
+                        if sys.version_info[0] == 3:
+                            raise ei.with_traceback(tb)
+                        else:
+                            raise et, "Fatal error found while addding '{}': {}".format(
+                                        endpoint.get('operationId'),
+                                        str(ei) ), tb
     def add_auth_on_not_found(self):
         """
         Adds a 404 error handler to authenticate and only expose the 404 status if the security validation pass.
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -168,7 +168,7 @@ class Api:
                         message = "Fatal error found while addding '{}': {}".format(
                                         endpoint.get('operationId'),
                                         str(ei))
-                        six.reraise(et, message, tb)
+                        six.reraise(et, AttributeError(message), tb)
 
     def add_auth_on_not_found(self):
         """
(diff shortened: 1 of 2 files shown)


rafaelcaricio:
Could you please update the [test](https://github.com/zalando/connexion/blob/master/tests/test_api.py#L28) and verify if the `AttributeError` that is being thrown is the correct one (just check if the correct missing method is in the exception message)?

TNGPS:
I looked into it, and found that indeed the original code raises an `AttributeError` in python 2.x as expected in the test, however it is not the Attribute Error you expected. I will change the test  (and the code).

However wonder with module you would expect in the test. The api loading fails because of in deep `ImportError` in `fakeapi.module_with_error` it would be good to see that error.

connexion/api.py
```diff
@@ -161,11 +161,15 @@ def add_paths(self, paths=None):
                     if self.debug:
                         logger.exception(error_msg)
                     else:
-                        import sys
                         logger.error(error_msg)
+                        import sys
                         et, ei, tb = sys.exc_info()
-                        raise ei.with_traceback(tb)
-
+                        if sys.version_info[0] == 3:
```
rafaelcaricio: Please use six for that as in https://github.com/zalando/connexion/blob/33f7457d1a4851da77f24bc20948a55190543038/connexion/decorators/parameter.py#L29

tests/test_api.py
```diff
@@ -25,13 +25,16 @@ def test_template():
...
+
 def test_invalid_operation_does_stop_application_to_setup():
-    with pytest.raises(AttributeError):
-        Api(TEST_FOLDER / "fakeapi/op_error_api.yaml", "/api/v1.0",
-            {'title': 'OK'})
+    with pytest.raises(AttributeError) as exc_info:
+        Api(TEST_FOLDER / "fakeapi/op_error_api.yaml", "/api/v1.0", {'title': 'OK'})
+
+    assert 'fakeapi.module_with_error.something' in str(exc_info.value)
+    # TODO: I would be nice to show the root cause or the import error of 'module_with_error' like this ?
```
rafaelcaricio: Not sure if I understand what you wanted to mean with this `TODO` comment/question.

Anyway, sure -- it is nice to show the errors as clear as possible.

rafaelcaricio:
Please also fix the `flake8` formatting errors.

tests/test_api.py
TNGPS: I was not able to get the traceback to to original cause, so I added this TODO comment, maybe better to make a new issue ?
I stubled upon [this article](http://code.activestate.com/recipes/578252-python-exception-chains-or-trees/?in=user-4182236) with may come in handy here ?

TNGPS:
flake 8 should be OK now

rafaelcaricio:
Still flake8 does not pass. Please also review the previous comments and I will be happy to move your PR forward.

connexion/api.py
```diff
@@ -161,11 +161,16 @@ def add_paths(self, paths=None):
...
+                            raise ei.with_traceback(tb)
+                        else:
+                            raise et, "Fatal error found while addding '{}': {}".format(
+                                        endpoint.get('operationId'),
+                                        str(ei)), tb
+
```
jmcs: According to the tests this line (173) has unnecessary spaces, that's what's making flake8 fail.

srlopez:
width the simple setup for the example: 

> import connexion
> 
> app = connexion.App(__name__, specification_dir='swagger/')
> app.add_api('api.yaml')
> app.run(port=8080)

And the envoronment show, and i have the following error message:

> (env) iMac:miserver$ pip freeze
> connexion==1.0.76
> Flask==0.10.1
> functools32==3.2.3.post2
> httpie==0.9.3
> itsdangerous==0.24
> Jinja2==2.8
> jsonschema==2.5.1
> MarkupSafe==0.23
> pathlib==1.0.1
> Pygments==2.1.3
> PyYAML==3.11
> requests==2.9.1
> six==1.10.0
> strict-rfc3339==0.6
> swagger-spec-validator==2.0.2
> Werkzeug==0.11.4


> (env) iMac:miserver$ python api.py
> No handlers could be found for logger "connexion.api"
> Traceback (most recent call last):
>   File "api.py", line 4, in <module>
>     app.add_api('api.yaml')
>   File "~/.pyenv/versions/miserver/lib/python2.7/site-packages/connexion/app.py", line 133, in add_api
>     debug=self.debug)
>   File "~/.pyenv/versions/miserver/lib/python2.7/site-packages/connexion/api.py", line 122, in __init__
>     self.add_paths()
>   File "~/.pyenv/versions/miserver/lib/python2.7/site-packages/connexion/api.py", line 187, in add_paths
>     raise ei.with_traceback(tb)
> AttributeError: 'exceptions.AttributeError' object has no attribute 'with_traceback'
> 

I think this is the same issue you are talking about

¿How to solve it? 
Thanks in advance

rafaelcaricio:
@srlopez That's exactly the problem this PR is trying to solve. The problem is that one of your `operationId` is pointing to a inexistent function in your python path. The actual error is being shadowed by this `AttributeError` with the `with_traceback`.

I am hopping to hear from @TNGPS, if not I will move this PR forward myself.

TNGPS:
Looks like all checks are OK now, this should do it.

Sorry for the delay, had a view deadlines to catch the last view weeks.

connexion/api.py
```diff
@@ -161,10 +161,14 @@ def add_paths(self, paths=None):
...
-                        import sys
                         logger.error(error_msg)
+                        import sys
+                        import six
                         et, ei, tb = sys.exc_info()
-                        raise ei.with_traceback(tb)
+                        message = "Fatal error found while addding '{}': {}".format(
+                                        endpoint.get('operationId'),
+                                        str(ei))
+                        six.reraise(et, AttributeError(message), tb)
```
rafaelcaricio: You should not create a new exception object here.

rafaelcaricio:
@TNGPS Sorry, this issue was solved by the PR #186. We needed a simple implementation and it was getting more complicated here.

For your improvement I will list the problems I had with this PR:
- Too many commits to handle a small change; How to do better? When sending a PR try to [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits that do not add any useful information or concise block of code that would make it easier to understand the code.
- Obscure comments around the code and commented code. Commented code is dead code and just confuse whoever is trying to understand the codebase.
- Long iteration loops. Do not need to work immediately since we are working usually in our free time and in things we care, but it was also an issue here.

I appreciate your investment in time and effort here and I hope to see more PR's from you in the future.

TNGPS:
thanks for the time taken anyways. Let me point out that this was indended to be a single line change, one commit and all should be fine, however the Python2 vs Python3 made this a bit more complicated than I hoped.  What we can take away from this (as I see it) that the test for the orginal change, did fail to see the problem in python 2 by not failing the test. 

Changing the test to prevent future problems here may still be a good idee ?

PR #163, opened 2016-02-29
add custom security decorator attribute to the app.add_api() call so …

TNGPS:
writing your own security logic without changing the connexion code itself, is not easy.

In this branch I added the possibility to add a custom security decorator to the add_api() method.

Now it is easy to write curtom security logic in you're own decorator, and pass it to the connection app as it is created. I updated the readme  with an example how to use this.


README.rst
```diff
@@ -152,6 +152,110 @@ URL to use to validate and get the token information.
...
+      :param token_info_url: Url to get information about the token
+      :type token_info_url: str
+      :param allowed_scopes: Set with scopes that are allowed to access the endpoint
+      :type allowed_scopes: set
+      :type function: types.FunctionType
+      :rtype: types.FunctionType
+      """
+
+      @functools.wraps(function)
+      def wrapper(*args, **kwargs):
```
rafaelcaricio: Maybe you could use a way simpler example code than that? It would be better to have as example the very minimal work necessary to have a custom security decorator working. So it could be used as bootstrap code for the reader's code.

connexion/api.py
```diff
@@ -164,7 +171,8 @@ def add_paths(self, paths=None):
                         import sys
                         logger.error(error_msg)
                         et, ei, tb = sys.exc_info()
-                        raise ei.with_traceback(tb)
+                        ei.__traceback__ = tb
```
rafaelcaricio: That change was made in another PR. So it will cause a merge conflict for this branch after the #162 gets merged or when it get merged depending on what comes first.

rafaelcaricio: TL;TR: you should undo this change and let your PR's focused in every change you want to make.

connexion/app.py
```diff
@@ -129,6 +132,7 @@ def add_api(self, swagger_file, base_path=None, arguments=None, auth_all_paths=N
                   swagger_url=swagger_url,
                   resolver=resolver,
                   validate_responses=validate_responses,
+                  security_decorator=security_decorator,
```
rafaelcaricio: If you pass the security_decorator variable into the constructor of `App` class it should be considered here. Just like the other global configs like `swagger_ui`, `swagger_path`, etc.

Precisely, I think you should add something like this line:
```python
security_decorator = security_decorator if security_decorator is not None else self.security_decorator
```

connexion/app.py
```diff
@@ -87,7 +87,8 @@ def common_error_handler(exception):
         return problem(title=exception.name, detail=exception.description, status=exception.code)
 
     def add_api(self, swagger_file, base_path=None, arguments=None, auth_all_paths=None, swagger_ui=None,
-                swagger_path=None, swagger_url=None, validate_responses=False, resolver=Resolver()):
+                swagger_path=None, swagger_url=None, validate_responses=False, resolver=Resolver(),
+                security_decorator=None):
```
rafaelcaricio: Probably this option should be added to the constructor of `App` class too, in case you want to add this security decorator globally in your APIs.

TNGPS: the app constructor need not know about the decorator here. The decorator will be specific to an API that is added using the `app.add_api()` AFTER the app was instantiated using 

`app = connection.App(....)`

connexion/api.py
TNGPS: clear, I will remove this line here

TNGPS:
any suggestions how we can add this new feature to the test suite ? Looks like the coverage is now down with 0.2 % ....

LappleApple:
@jmcs @rafaelcaricio @hjacobs Hi guys, can we respond to the person who made this PR and then resolve/close as appropriate?

rafaelcaricio:
No follow up. No work in response of code review. Closing...

Issue #153, opened 2016-02-18
Log details of response validation errors?


danballance:
Hi guys,

I'm not sure if I'm missing an option here - or maybe I can add something in for you? So I haven't upgraded in a while and now when I run my code I get this error message:

500, "Response body does not conform to specification"

Tracing this back through the code, this seems to be generated at lines 54-56 of decorators/response.py. If I examine the error returned, there is plenty of information to help me fix my broken response, but there doesn't seem to be any way for me to discover what the underlying issue is without inspecting the code.

Shall I submit a pull request with some extra debug logging there? IMO it would be useful to check your logs at this point and see exactly what the validation problem is - even if we are only returning the generic error to the end user.

Let me know your thoughts!

hjacobs:
@danballance I guess that sounds reasonable :smile:

rafaelcaricio:
Hi @danballance! It makes total sense what you wrote. If you can send a PR with your suggestions would be even better. Anyway, submitting an example of what your are experiencing would help to learn how we can improve.

danballance:
Hi guys,

I got sidetracked with other work, but hope to get a pull request in this week. It's only quite a simple change to add the extra logging.

rafaelcaricio:
Solution for this issue being discussed in https://github.com/zalando/connexion/pull/186

PR #171, opened 2016-03-04
Attempt to better log the validation error responses receive in

danballance:
So this is my proposal for enhancement  #153 

So the main difficulty is that RequestBodyValidator.validate_schema() returns a problem like so:

problem(400, 'Bad Request', str(exception))

So the exception is basically cast to a string - even though this is quite a complex python data structure in the case of these validation errors.

So my small tweak is to use the ast module to safely convert this string back to a python data structure. I then attempt a couple of sanity checks - it's a list with at least one item - and then pull the error dictionary out. Once I have this object I'm able to log the error in a way that is easier to read (for me at least!) than just logging the returned string representation of the error.

Do you think this is worth doing? Is it safe? I'm not too sure about what RequestBodyValidator is doing - handling the error from json-schema I assume?

I've also not been able to test this properly under Python 3. It seems to work fine with 2.7 in my tests so far. Feedback welcomed :)


rafaelcaricio:
Hi @danballance , first of all -- thank you very much for working in that issue.

Before I start digging into your PR. Please update your master branch (your fork) and then create a new branch and cherry-pick the most important commit https://github.com/danballance/connexion/commit/ceda46f3ecfa011c31afdcec40f929f142942311 and send a new PR with that new branch. The thing is that I rather not merge this commit https://github.com/danballance/connexion/commit/d6b88c3ba5cc9d88fd6ccb3910b95730fc843987 which is useless.

rafaelcaricio:
TL;TR:
Please do it:
```bash
git checkout master
git remote add original git@github.com:zalando/connexion.git
git pull original master
git checkout -b new-pr
git cherry-pick ceda46f
```
And send a new Pull Request from your new branch `new-pr`.
 :)

danballance:
Apologies for the delay here - I missed the Github email notification. I have created a new pull request #175 that is hopefully a cleaner commit and what you are looking or :)

rafaelcaricio:
replaced by #175.

PR #175, opened 2016-03-08
Attempt to convert string representation of the error back to a pytho…

danballance:
So this is my proposal for enhancement #153

This is a second attempt - hopefully a clean commit following #171.

So the main difficulty is that RequestBodyValidator.validate_schema() returns a problem like so:

problem(400, 'Bad Request', str(exception))

So the exception is basically cast to a string - even though this is quite a complex python data structure in the case of these validation errors.

So my small tweak is to use the ast module to safely convert this string back to a python data structure. I then attempt a couple of sanity checks - it's a list with at least one item - and then pull the error dictionary out. Once I have this object I'm able to log the error in a way that is easier to read (for me at least!) than just logging the returned string representation of the error.

Do you think this is worth doing? Is it safe? I'm not too sure about what RequestBodyValidator is doing - handling the error from json-schema I assume?

Maybe the better way to solve this issue would be to alter the behaviour in RequestBodyValidator.validate_schema() to not blindy convert the retuned exception to a string? 
 
I've also not been able to test this properly under Python 3. It seems to work fine with 2.7 in my tests so far. Feedback welcomed :)

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -14,6 +14,7 @@ Unless required by applicable law or agreed to in writing, software distributed 
 # Decorators to change the return type of endpoints
 import functools
 import logging
+import ast
 from ..exceptions import NonConformingResponse
 from ..problem import problem
 from .validation import RequestBodyValidator
@@ -51,6 +52,9 @@ class ResponseValidator(BaseDecorator):
             v = RequestBodyValidator(schema)
             error = v.validate_schema(data, schema)
             if error:
+                if isinstance(error.response, list) and len(error.response) > 0:
+                    error_dict = ast.literal_eval(error.response[0])
+                    logger.debug('Validation error in response body was:\n{0}'.format(error_dict["detail"]))
                 raise NonConformingResponse("Response body does not conform to specification")
 
         if response_definition and response_definition.get("headers"):


rafaelcaricio:
Hi @danballance ,

Thank you very much for your PR. I also got in the same situation as you described, wanting a better error message in the case of my response does not match the specification. I think that is a valid issue to handle in a better/more clear way.

I think that this PR indeed solves the problem showing a better feedback to the developer. Though it looks like a hack over Connexion's code itself. So I would like to suggest that we could maybe remove that `str(exception)` and pass this precious information in a better way.

Maybe we could extract the proper "detail" information from the exception itself and then pass it to the `problem` function?

dfeinzeig:
I think a question is whether you want to expose that level of information in the problem response.  There are a ton of places that log debug like info, so I think it's interesting to determine when you want to just log something vs. return it in an http response.

danballance:
I'd agree we don't want to expose that validation information in the problem response to the end user. I think a simple debug log message would be better.

I don't seem to get all of Github's notifications for some reason, so I unfortunately missed @rafaelcaricio response to me earlier in the month. I am off to the states for a week now, but could take a look at this when I get back if needed.

I agree with what has been said, my PR is really a hack. Getting the clean error string and simply debug logging that would be the best solution IMHO. However it seems like @dfeinzeig is already solving this issue here:

https://github.com/zalando/connexion/pull/185

Perhaps my PR could just be closed and enahcement #153 would also be solved by that work?

rafaelcaricio:
Agree with you guys on not exposing the error to users. I was not suggesting to add the exception error message to the `problem` object, but using the "detail" key in the logging message. @dfeinzeig already sent a #186 which do exactly what I was going to suggest. :)

rafaelcaricio:
Closing this in favor to #186.

PR #177, opened 2016-03-08
Remove assignment to prevent unexpected keyword argument 'port' error

JoshStutts:
When using the tornado web server, the following error was thrown: 

TypeError: _log() got an unexpected keyword argument 'port'

This is resolved by removing the assignment in the logging statement in the tornado section of the app run function.


hjacobs:
:+1: thanks!

PR #181, opened 2016-03-11
Add support for responses as references

ja-b:
Fixes # .180

# Adds support for global response definitions

 -  Support global response definitions in the same way model definitions and parameter definitions are supported.
 -  Allow for validation of responses when `validate_response=True` for specifications that use global response definitions.


ja-b:
Will reopen with changes to tests

PR #185, opened 2016-03-17
Updated body validation error detail to be more easily human readable

JoshStutts:
The class documentation in problem.py indicates that the detail parameter should be "An human readable explanation specific to this occurrence of the problem."  I've updated the body validation code to issue a problem containing just the jsonschema ValidationError message instead of the entire exception object converted to a string.  

When using a large schema for a model that is being validated, this entire exception object renders the detail portion of the error message unusable for quick and easy debugging.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -117,7 +117,7 @@ class RequestBodyValidator(object):
         try:
             validate(data, self.schema, format_checker=draft4_format_checker)
         except ValidationError as exception:
-            return problem(400, 'Bad Request', str(exception))
+            return problem(400, 'Bad Request', str(exception.message))
 
         return None
 


dfeinzeig:
Does this actually help?  Won't it get swallowed by:

https://github.com/zalando/connexion/blob/master/connexion/decorators/response.py#L54

I was actually looking at this today and was going to just settle for a debug statement to make finding the actual cause of a validation error easier to find.

JoshStutts:
I'll take another look tomorrow, but in my environment, I was receiving the
full ValidationError, which led me to the line I modified.

dfeinzeig:
Hmm, OK. I'd be curious to know how you have your environment configured so
I can do the same.

danballance:
Hi folks,

There's a related issue here:

https://github.com/zalando/connexion/issues/153

If the validation decorator could be made to set a simple problem string
then my pull request would simply need to log that problem string and not
need to mess around trying to get hold of the origin exception object.

It didn't log anything there originally because of the crazy object cast to
a string and being new to the framework I didn't feel confident enough to
start making more widespread changes.

dfeinzeig:
@danballance , thanks for pointing it out!

at a minimum, it feels like logging is a good thing, even if/when the details are included in the response.

i was going to do

```
    def validate_schema(self, data):
        """
        :type schema: dict
        :rtype: flask.Response | None
        """
        try:
            # ADDED THIS DEBUG
            logger.debug("%s validation error: %s" % (flask.request.url, exception))
            return problem(400, 'Bad Request', str(exception.message))

        return None
```
in validation.py, since it seems consistent with other logging.

Maybe we have NonConformingResponseBody and NonConformingResponseHeaders exceptions, so we can raise those from ResponseValidator.validate_response() in response.py with the details of the exception as the message, and then in __call__() catch them separately so we can stuff in the different titles for each, passing in the exception message as description?  If that sounds reasonable, I can whip something up.

dfeinzeig:
Here's something quick for discussion.  I'll clean it up if we like it.

https://github.com/zalando/connexion/pull/186/files

danballance:
PR https://github.com/zalando/connexion/pull/186 by @dfeinzeig looks a good solution to me :)

rafaelcaricio:
@JoshStutts I am closing this PR in favor to #186. Could you please check if that satisfies your concerns?

@dfeinzeig Please take a look in the comments on your PR (#186) so we can move it forward.

JoshStutts:
The approach in #186 looks like a more complete solution, my primary concern was to eliminate logging of the entire exception converted to a string and instead to log the exception's message.

rafaelcaricio:
@JoshStutts Great! So let's move our efforts on #186 then. :+1:

PR #186, opened 2016-03-17
Better error handling for non-conforming responses

### LINKED ISSUE #153: Log details of response validation errors?

dfeinzeig:
Fixes #153 .

# Description

 - let the validation error be a validation error
 - catch the validation error when validating the response
 - create exceptions for nonconforming body and headers and use them to differentiate between the different validation errors and to populate a problem response

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -16,8 +16,10 @@ import jinja2
 import json
 import logging
 import pathlib
-import yaml
+import six
+import sys
 import werkzeug.exceptions
+import yaml
 from swagger_spec_validator.validator20 import validate_spec
 from .operation import Operation
 from . import utils
@@ -181,10 +183,8 @@ class Api(object):
                     if self.debug:
                         logger.exception(error_msg)
                     else:
-                        import sys
                         logger.error(error_msg)
-                        et, ei, tb = sys.exc_info()
-                        raise ei.with_traceback(tb)
+                        six.reraise(*sys.exc_info())
 
     def add_auth_on_not_found(self):
         """
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -14,10 +14,11 @@ Unless required by applicable law or agreed to in writing, software distributed 
 # Decorators to change the return type of endpoints
 import functools
 import logging
-from ..exceptions import NonConformingResponse
+from ..exceptions import NonConformingResponseBody, NonConformingResponseHeaders
 from ..problem import problem
-from .validation import RequestBodyValidator
+from .validation import ResponseBodyValidator
 from .decorator import BaseDecorator
+from jsonschema import ValidationError
 
 
 logger = logging.getLogger('connexion.decorators.response')
@@ -48,14 +49,18 @@ class ResponseValidator(BaseDecorator):
 
         if response_definition and response_definition.get("schema"):
             schema = response_definition.get("schema")
-            v = RequestBodyValidator(schema)
-            error = v.validate_schema(data)
-            if error:
-                raise NonConformingResponse("Response body does not conform to specification")
+            v = ResponseBodyValidator(schema)
+            try:
+                v.validate_schema(data)
+            except ValidationError as e:
+                raise NonConformingResponseBody(message=str(e))
 
         if response_definition and response_definition.get("headers"):
-            if not all(item in headers.keys() for item in response_definition.get("headers").keys()):
-                raise NonConformingResponse("Response headers do not conform to specification")
+            response_definition_header_keys = response_definition.get("headers").keys()
+            if not all(item in headers.keys() for item in response_definition_header_keys):
+                raise NonConformingResponseHeaders(
+                    message="Keys in header don't match response specification. Difference: %s"
+                    % list(set(headers.keys()).symmetric_difference(set(response_definition_header_keys))))
         return True
 
     def __call__(self, function):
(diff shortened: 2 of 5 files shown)


connexion/decorators/validation.py
```diff
@@ -117,7 +117,8 @@ def validate_schema(self, data):
         try:
             validate(data, self.schema, format_checker=draft4_format_checker)
         except ValidationError as exception:
-            return problem(400, 'Bad Request', str(exception))
+            logger.debug("%s validation error: %s" % (flask.request.url, exception))
+            raise
```
rafaelcaricio: :+1: This method was doing too much creating this `problem` object.

rafaelcaricio:
Please clean it up and I will be happy to merge it. :)

jmcs:
Thanks for the PR. The unit tests are failing so you need to fix them before we can merge it.

connexion/decorators/validation.py
rafaelcaricio: Maybe better to use it https://pythonhosted.org/six/#six.reraise
So it is compatible across Python versions.

connexion/decorators/response.py
```diff
@@ -69,8 +71,10 @@ def wrapper(*args, **kwargs):
             try:
                 data, status_code, headers = self.get_full_response(result)
                 self.validate_response(data, status_code, headers)
-            except NonConformingResponse as e:
-                return problem(500, 'Internal Server Error', e.reason)
+            except NonConformingResponseBody as e:
+                return problem(500, 'Response body does not conform to specification', str(e))
```
rafaelcaricio: If you move those default error messages to their respective classes (`NonConformingResponseBody` and `NonConformingResponseHeaders`) this code could be simplified to be only:
```python
except NonConformingResponse as e:
    return problem(500, e.message, str(e))
```

connexion/decorators/response.py
```diff
@@ -49,13 +50,14 @@ def validate_response(self, data, status_code, headers):
...
-                raise NonConformingResponse("Response body does not conform to specification")
+            try:
+                v.validate_schema(data)
+            except ValidationError as e:
+                raise NonConformingResponseBody(str(e))
 
         if response_definition and response_definition.get("headers"):
             if not all(item in headers.keys() for item in response_definition.get("headers").keys()):
-                raise NonConformingResponse("Response headers do not conform to specification")
+                raise NonConformingResponseHeaders()
```
rafaelcaricio: Maybe we could add an argument here to be more clean:
```python
raise NonConformingResponseHeaders("Missing key in response header")
```

connexion/exceptions.py
```diff
@@ -44,3 +44,11 @@ def __str__(self):  # pragma: no cover
 
     def __repr__(self):  # pragma: no cover
         return '<NonConformingResponse: {}>'.format(self.reason)
+
+
+class NonConformingResponseBody(NonConformingResponse):
+    pass
```
rafaelcaricio: Here to handle better all scenarios and be clear on what I am suggesting we could add:
```python
def __init__(self, reason):
   super(NonConformingResponseBody, self).__init__(
         'Response body does not conform to specification')
   self.reason = reason
```

And similar code just changing the message to `NonConformingResponseHeaders` exception.

rafaelcaricio:
Please [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) :scissors: this commit https://github.com/zalando/connexion/pull/186/commits/259adb5d7e8e6d26d1a60432391583e7014571a7 to leave the PR clean. :shower:

connexion/decorators/validation.py
dfeinzeig: good catch!  (sorry for the bad pun  ;)  )

connexion/decorators/response.py
dfeinzeig: good idea

connexion/exceptions.py
dfeinzeig: makes sense

dfeinzeig:
Thanks for the feedback.  I'll make the requested changes and fix the tests today.

dfeinzeig:
Ok, I've finally figured this out.  Should be good to go.

I ended up separating response body validation from request body validation, because request body validation needs the returned problem at the moment whereas response body validation needs to raise exceptions.  Also, these could diverge as they are meant for different things.

connexion/api.py
```diff
@@ -181,10 +183,8 @@ def add_paths(self, paths=None):
                     if self.debug:
                         logger.exception(error_msg)
                     else:
-                        import sys
                         logger.error(error_msg)
-                        et, ei, tb = sys.exc_info()
-                        raise ei.with_traceback(tb)
+                        six.reraise(*sys.exc_info())
```
dfeinzeig: unrelated to this PR, but I hit this the other day on python2, so figured I'd fix it

rafaelcaricio:
:+1:

connexion/api.py
rafaelcaricio: There is a PR hanging for that https://github.com/zalando/connexion/pull/162
But I am willing to accept this here since there is no activity, from the author, in that PR for quite some time.

dfeinzeig:
great!  anything else you'd like me to update before merging?

rafaelcaricio:
@dfeinzeig no, that is good to go. Thank you for taking care of this issue.

dfeinzeig:
No problem, glad to do it

PR #187, opened 2016-03-18
Response definitions can be defined globally in spec

### LINKED ISSUE #180: Global response definitions not supported
When response validation is enabled, we trigger the `$ref needs to point to definitions or parameters` when using global response definitions. The swagger [spec](http://swagger.io/specification/) contains the responses field as global and should support references to global responses.

This is a matter of copying the behavior for definitions and parameters to handle responses. I can add a pull request for this.

rafaelcaricio:
Fixes #180 .

# Description
As described in [Swagger Specification](http://swagger.io/specification/#responsesDefinitionsObject) it is possible to define global responses to be used in endpoint specs.
 
 - Added possibility to reference (exp.: `$ref: '#/responses/Something'`) responses defined globally in specification.

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -103,6 +103,7 @@ class Api(object):
 
         self.definitions = self.specification.get('definitions', {})
         self.parameter_definitions = self.specification.get('parameters', {})
+        self.response_definitions = self.specification.get('responses', {})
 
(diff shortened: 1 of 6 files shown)


ja-b:
Tested with a global response spec, works as expected.

PR #189, opened 2016-03-21
problem response extension

lukyoung:
Fixes # .

This fix will help to extend response dict, if you need to return more data.


rafaelcaricio:
Hi @lukyoung,
Thank you for the PR proposal.

Using Flask app config to change Connexion behaviour was never used before. It sounds like a good way to allow users to extend some parts of Connexion.

I think if we are going this way we should write some documentation on what are the options allowed. Also it needs some automated test. I am not willing to accept PR's without tests.

rafaelcaricio:
Since there was no response from the author of this PR, I am closing it.

PR #193, opened 2016-03-24
Make JSON printing a bit nicer

gdb:
This change makes JSON generation more consistent and human-friendly:

- Adding a trailing newline is nicer to people using `curl` from the command line, without making the machine case worse.
- Add `indent=2` to errors so they are pretty printed as well.

### DIFF
--- a/connexion/problem.py
+++ b/connexion/problem.py
@@ -46,7 +46,7 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
     if ext:
         problem_response.update(ext)
 
-    body = json.dumps(problem_response)
+    body = [json.dumps(problem_response, indent=2), '\n']
     response = flask.current_app.response_class(body, mimetype='application/problem+json',
                                                 status=status)  # type: flask.Response
     if headers:
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -124,7 +124,7 @@ class Jsonifier(BaseSerializer):
                 logger.debug('Endpoint returned an empty response (204)', extra={'url': url, 'mimetype': self.mimetype})
                 return '', 204, headers
 
-            data = json.dumps(data, indent=2)
+            data = [json.dumps(data, indent=2), '\n']
             response = flask.current_app.response_class(data, mimetype=self.mimetype)  # type: flask.Response
             response = self.process_headers(response, headers)
 
--- a/tests/api/test_bootstrap.py
+++ b/tests/api/test_bootstrap.py
@@ -100,13 +100,13 @@ def test_single_route(simple_app):
 def test_resolve_method(simple_app):
     app_client = simple_app.app.test_client()
     resp = app_client.get('/v1.0/resolver-test/method')  # type: flask.Response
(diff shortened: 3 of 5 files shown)


connexion/problem.py
```diff
@@ -46,7 +46,7 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
     if ext:
         problem_response.update(ext)
 
-    body = json.dumps(problem_response)
+    body = json.dumps(problem_response, indent=2) + "\n"
```
jmcs: The `+ "\n"` means you create 3 string every time instead of 1.

jmcs:
The unit tests are also failing.

gdb:
Whoops, sorry yes I'd meant to ask about whether the implementation and functionality was ok before fixing up the tests. (On the one hand, copies always sound bad. On the other hand, there was just a `json.dumps` there, which I'd expect would always be much more expensive than the copy. You also mention 3 strings — are you concerned about creating the '\n'?)

I just switched the implementation to not do the copy (but instead pass a list to the Response — I'm less familiar with the Python WSGI ecosystem than the Ruby Rack one, but I believe this should be supported regardless of `response_class`?), and fixed the tests — is this any better?

ainmosni:
Is it really the frameworks responsibility to pretty print it? I mean, everyone working with REST has tools to pretty print it on their end. At the very least they can pipe it to python -m json.tool or jq.

gdb:
At Stripe, one of our core observations was that the little details are what changes an API from acceptable to actually a pleasure to use. (And particularly, the most important details happen during interactive development: when you're first `curl`ing around a new API. We would do things like pre-fill your test API key in the API docs, so you can just copy-paste the examples directly without having to figure out where your API key lives, etc.)

Certainly developers can pipe the output of a curl to `jq`. But even better would be anticipating this use-case and returning the JSON already in a human-friendly format.

Regardless of the philosophy, this framework already does `indent=2` in one case; here I've made it consistent and added trailing newlines. I'd also be equally happy if this functionality were easily pluggable. But I think it's actually a pretty good thing for the framework to provide little finishing touches that otherwise most people probably wouldn't bother with.

hjacobs:
@gdb maybe you can consider using HTTPie, https://github.com/jkbrzt/httpie

>  .. user-friendly curl replacement with intuitive UI, JSON support, syntax highlighting, wget-like downloads, extensions, etc

I use it everyday to do REST calls :smile:

gdb:
HTTPie is great. Unfortunately, it's less about me developing my own API, and more about the external developers who are going to integrate against my API.

hjacobs:
@jmcs @rafaelcaricio what is now the status here (apart from merge conflicts)? I think the trailing newline is OK (we already have indent as @gdb points out).

rafaelcaricio:
@hjacobs I guess it is fine to add the new-line. In the future we could add configuration to Connexion to enable "human readable JSON responses", so it would enable the indentation + newline.

hjacobs:
@gdb can you resolve the merge conflict to get this PR into master? Thanks!

gdb:
Ok!

Issue #182, opened 2016-03-11
Add flag for validation disabling


cansik:
In my swagger definition I have a lot of parameters which contain a valid datatype which then is needed for our automated query builder:

```yaml
  PropertyGroup:
    properties:
      time_start:
        description: Start time of flare
        type: string
        format: date-time

      nar:
        description: NOAA active region number
        type: number
        format: int32

      lat_hg:
        description: Stonyhurst heliographic latitude of event
        type: number
        format: int32
```

Now the problem we encounter is, that sometimes the validation of these types is not possible because we need for example **nullable types** as well.

So my question is, if it is possible to add an option which disables the validation or a method to provide an own validator class for some datatypes?

hjacobs:
@rafaelcaricio @jmcs could you check this?

rafaelcaricio:
I was reading about [this issue](https://github.com/OAI/OpenAPI-Specification/issues/229) which cover this topic about Swagger/OpenAPI Spec. I would be nice to support `x-nullable` in Connexion. But I also agree we should have a way to make possible more personalisation of custom types/formats. I will invest some work on that next week.

rafaelcaricio:
@cansik Please check if #197 solves partially your problem.

Allowing customisation for validation is the next step.

cansik:
@rafaelcaricio Yes this looks very promising! I'll give it a try this week. Thank you!

rafaelcaricio:
@cansik It might interest you. I just added an example how to create custom formats to be used with Connexion https://github.com/zalando/connexion/pull/216

cansik:
@rafaelcaricio Great!

cansik:
@rafaelcaricio So it is possible to write a validator for `format`, is it also possible to write one for `type`?

rafaelcaricio:
@cansik You validate types using formats, no? Do you wanna define new types?
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types

cansik:
@rafaelcaricio Yes your right, but as described in my initial issue, we need to support nullable types. This isn't possible with `format` validators because e.g. as `type` we use **number**. Even if I now create my own format validator `nullableint32`, the type validator won't let `None` pass.

So the only possible option I have would be to send the data as string and validate it then with a formatter. Am I right?

rafaelcaricio:
@cansik you can define many types for a single field. So in your case you can do:
```json
{
[...]

PR #197, opened 2016-03-25
Support nullable parameters

rafaelcaricio:
Partially solves #182.

Add support to `x-nullable` vendor extension in Connexion. It is a [ongoing discussion](https://github.com/OAI/OpenAPI-Specification/issues/229) in OpenAPI spec, but requested by users of Connexion and acknowledged as a valid use case.

 - Add support for attribute `x-nullable` flag in parameter definitions.
 - Add documentation for `x-nullable` flag usage.

### DIFF
--- a/docs/request.rst
+++ b/docs/request.rst
@@ -110,8 +110,12 @@ definitions. It's usage would be:
 It is supported by Connexion in all parameter types: `body`, `query`,
 `formData`, and `path`. Nullable values are the strings `null` and `None`.
 
+.. warning:: Be careful on nullable paramenters for sensitive data where the
+             strings "null" or "None" can be _`valid values`.
+
 .. _`nullable parameters`: https://github.com/zalando/connexion/issues/182
 .. _`not support`: https://github.com/OAI/OpenAPI-Specification/issues/229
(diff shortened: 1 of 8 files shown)


docs/request.rst
```diff
@@ -89,6 +89,34 @@ supports collection formats "pipes" and "csv". The default format is "csv".
 
 .. _jsonschema: https://pypi.python.org/pypi/jsonschema
 
+Nullable parameters
+^^^^^^^^^^^^^^^^^^^
+
+Sometimes your API should explicitly accept _`nullable parameters`. However
+OpenAPI specification currently does _`not support` officially a way to serve
+this use case, Connexion adds the `x-nullable` vendor extension to parameter
+definitions. It's usage would be:
```
jmcs: I would add a note saying than if/when a future version of OpenAPI defines an option for this, we will deprecate this extension.

rafaelcaricio: Sounds good. I will add that.

jmcs:
:+1:

rafaelcaricio:
Can we merge this PR?

PR #201, opened 2016-03-31
Real version is generated during release process

### LINKED ISSUE #194: version-mismatch: v1.0.33 in master, while v1.0.76 was already released
On pypi I see as latest release v1.0.76
https://pypi.python.org/pypi/connexion/1.0.76

yet  in github-master I see 1.0.33
https://github.com/zalando/connexion/blob/master/connexion/__init__.py

this breaks my setup when using addon packages like swagger-tester
https://github.com/Trax-air/swagger-tester

this depends on connexion >=1.0.37, so when I switch to master, 
my setup stops working because of broken dependencies.

The fact that the version-number is incorrect seems to indicate a broken release-process?

rafaelcaricio:
Fixes #194

Add placeholder and comment to avoid misunderstanding about versions in development mode.

 - Change placeholder version to be `2016.0.dev1`
 - Add note to inform that versions are generated during release process.

### DIFF
--- a/connexion/__init__.py
+++ b/connexion/__init__.py
@@ -13,13 +13,13 @@ Unless required by applicable law or agreed to in writing, software distributed 
  language governing permissions and limitations under the License.
 """
 
-from flask import abort, request, send_file, send_from_directory, render_template, render_template_string, \
-    url_for
-import werkzeug.exceptions as exceptions
-from .app import App
-from .api import Api
-from .problem import problem
-from .decorators.produces import NoContent
-from .resolver import Resolution, Resolver, RestyResolver
+from flask import (abort, request, send_file, send_from_directory,  # NOQA
+                   render_template, render_template_string, url_for)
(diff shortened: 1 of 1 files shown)


WouterVH:
:-1: 
this version-number compares incorrectly with all released connexions-packages=,

if a third-party package specifies "use connexions v1.0.37  **and higher**",
the trunk-version /master-dev will not match.

jmcs:
@rafaelcaricio regarding @WouterVH comment see https://www.python.org/dev/peps/pep-0440/

Maybe we can use the current year (`2016`) as the major version in the master.

rafaelcaricio:
I changed the placeholder version to be `2016.0.dev1`. Just a reminder that this is a **placeholder** and never gets published to public in our official released version on *pypi*. It could be anything.

hjacobs:
:+1:, let's see what @WouterVH thinks of it :smirk:

WouterVH:
many thanks.

PR #205, opened 2016-04-04
Validate accordingly with content-type defined in spec

### LINKED ISSUE #203: File download doesn't work
I'm trying to use connexion to return a requested file, such that the 200 response has a binary string as its body. I've tried two ways so far and neither have worked. The first looks like a bug in swagger_spec_validator, but the second looks like a Connexion issue.

Swagger yaml fragment:
```
  /deployments/{deployment_id}/avatars/{avatar_filename}:
    get:
      tags:
      - "Avatars"
      summary: "Download an avatar"
      operationId: "mypackage.controller.get_avatar"
      produces:
      - "application/octet-stream"
      parameters:
      - $ref: "#/parameters/deployment_id"
      - $ref: "#/parameters/avatar_filename"
      responses:
        200:
          description: Avatar file
          # Choose one of the schema's below
          # Problem 1 occurs if schema has type: file
          schema:
            type: file
          # Problem 2 occurs if schema has type: string
          schema:
            type: string
            format: binary
        201:
          description: "Avatar uploaded"
```

### Problem 1
I raised https://github.com/Yelp/swagger_spec_validator/issues/46 to cover this, but actually this is fixed in latest master. Essentially swagger_spec_validator didn't like the "file" type in a response schema.

### Problem 2
To work around problem 1, we instead try to return a binary string (octet-stream). My handler code is:
```
def get_avatar(deployment_id, avatar_filename):
    avatar_path = os.path.join(settings.AVATAR_DIR, avatar_filename)

    with open(avatar_path, 'r') as f:
        return f.read()
```
This string clearly isn't necessarily utf-8 encoded (it's the contents of a binary file), but Connexion assumes it is at https://github.com/zalando/connexion/blob/master/connexion/decorators/response.py#L57, when it tries to encode it as JSON.
Full error:
```
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "myeggsdir/eggs/connexion-1.0.84-py2.7.egg/connexion/decorators/validation.py", line 245, in wrapper
    response = function(*args, **kwargs)
  File "myeggsdir/eggs/connexion-1.0.84-py2.7.egg/connexion/decorators/produces.py", line 85, in wrapper
    data, status_code, headers = self.get_full_response(function(*args, **kwargs))
  File "myeggsdir/eggs/connexion-1.0.84-py2.7.egg/connexion/decorators/response.py", line 82, in wrapper
    self.validate_response(data, status_code, headers)
  File "myeggsdir/eggs/connexion-1.0.84-py2.7.egg/connexion/decorators/response.py", line 57, in validate_response
    data = json.dumps(data)
  File "myeggsdir/eggs/Flask-0.10.1-py2.7.egg/flask/json.py", line 126, in dumps
    rv = _json.dumps(obj, **kwargs)
  File "/usr/lib/python2.7/json/__init__.py", line 250, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 201, in encode
    return encode_basestring_ascii(o)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa0 in position 4050: invalid start byte
```
It seems like the only solution is to base-64 encode my file before sending, which I shouldn't have to do.

rafaelcaricio:
Fixes #203

Connexion should validate the schema of responses that have content-type as JSON.

 - Validate using json-schema only responses of operations that produces JSON.
 - Makes it possible for Connexion applications to return binary data in operations.

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -49,9 +49,7 @@ class ResponseValidator(BaseDecorator):
         response_definition = self.operation.resolve_reference(response_definition)
         # TODO handle default response definitions
 
-        if response_definition and "schema" in response_definition \
-           and (produces_json([self.mimetype]) or
-                self.mimetype == 'text/plain'):  # text/plain can also be validated with json schema
+        if self.is_json_schema_compatible(response_definition):
             schema = response_definition.get("schema")
             v = ResponseBodyValidator(schema)
             try:
@@ -71,6 +69,23 @@ class ResponseValidator(BaseDecorator):
                     message="Keys in header don't match response specification. Difference: %s"
                     % list(set(headers.keys()).symmetric_difference(set(response_definition_header_keys))))
         return True
+
+    def is_json_schema_compatible(self, response_definition):
+        """
+        Verify if the specified operation responses are JSON schema
+        compatible.
+
+        All operations that specify a JSON schema and have content
+        type "application/json" or "text/plain" can be validated using
+        json_schema package.
+
+        :type response_definition: dict
+        :rtype bool
+        """
+        if not response_definition:
+            return False
+        return ('schema' in response_definition and
+                (produces_json([self.mimetype]) or self.mimetype == 'text/plain'))
 
     def __call__(self, function):
         """
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -17,6 +17,7 @@ import functools
 import logging
 from ..exceptions import NonConformingResponseBody, NonConformingResponseHeaders
 from ..problem import problem
+from ..utils import produces_json
 from .validation import ResponseBodyValidator
 from .decorator import BaseDecorator
 from jsonschema import ValidationError
@@ -48,7 +49,9 @@ class ResponseValidator(BaseDecorator):
         response_definition = self.operation.resolve_reference(response_definition)
         # TODO handle default response definitions
 
-        if response_definition and response_definition.get("schema"):
+        if response_definition and "schema" in response_definition \
+           and (produces_json([self.mimetype]) or
+                self.mimetype == 'text/plain'):  # text/plain can also be validated with json schema
             schema = response_definition.get("schema")
             v = ResponseBodyValidator(schema)
             try:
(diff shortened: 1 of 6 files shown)


hjacobs:
:see_no_evil: decreasing code coverage..

rafaelcaricio:
Going up again :point_up:

connexion/decorators/response.py
```diff
@@ -48,7 +49,9 @@ def validate_response(self, data, status_code, headers):
         response_definition = self.operation.resolve_reference(response_definition)
         # TODO handle default response definitions
 
-        if response_definition and response_definition.get("schema"):
+        if response_definition and "schema" in response_definition \
```
jmcs: Using parenthesis looks better than using line continuations ('\')

connexion/decorators/response.py
```diff
@@ -48,7 +49,9 @@ def validate_response(self, data, status_code, headers):
...
+           and (produces_json([self.mimetype]) or
```
jmcs: Using parenthesis here would look better than using line continuations ('\').

connexion/operation.py
```diff
@@ -264,16 +264,17 @@ def _retrieve_reference(self, reference):
         return definition
 
     def get_mimetype(self):
+        # if the endpoint as no 'produces' then the default is 'application/json'
+        default_mimetype = 'application/json'
```
jmcs: As this is a constant you should put it a the module level to avoid creating a string every time get_mimetype is called.

jmcs:
:+1:

hjacobs:
:+1:

PR #209, opened 2016-04-07
Add minimum versions of dependencies

### LINKED ISSUE #195: specify minimum requirements
the connexion-requirements are listed without any minimum version-numbers,
this can lead to broken setups, when older dependencies are already installed.

eg. connexion is broken when using jsonschema v2.4.0
so the requirement should be:
jsonschema >= v2.5.0

rafaelcaricio:
Fixes #195

I ran a series of tests using different minimal versions of each requirement of Connexion and found the minimum we need to run.

 - Add minimum versions of each requirement of Connexion that have it working across the supported Python versions/implementations.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-jsonschema
-flask
-PyYAML
-requests
-six
-strict-rfc3339
-swagger_spec_validator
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

PR #211, opened 2016-04-13
Do not provide too much detailed information on unauthorized access

rafaelcaricio:
Connexion is giving too much information to unauthorized entities trying to access the Connexion secured applications. 

In my opinion a malicious user could investigate, with "debugging" level detail, how the authentication is being handled inside the applications secured by Connexion.

 - Removed the debugging messages related to authentication of requests.
 - Moved the messages to proper debug logging.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -1,18 +1,3 @@
-"""
-Copyright 2015 Zalando SE
-
-Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
-License. You may obtain a copy of the License at
-
-http://www.apache.org/licenses/LICENSE-2.0
-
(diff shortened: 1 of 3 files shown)


hjacobs:
@rafaelcaricio actually I would disagree here: telling the user whether user *or* password are wrong *is* a security issue, but here we only tell the client that a "token" is required. Knowing that the application requires an OAuth Bearer token is not really an issue --- only if you consider "security by obscurity" a valid counter measure :smirk:

hjacobs:
We have a lot of Open Source services where it is public knowledge that they require OAuth Bearer tokens, changing this response message in Connexion would not change this fact...

rafaelcaricio:
Indeed it is security by obscurity. It tries to hide the application's security logic and even that it is Connexion code.

I wanted to rise the discussion on this topic. If it is not a concern, I am good with having this PR closed.

PR #212, opened 2016-04-18
Add __call__ method for WSGI-compliance

tdpsk:

WSGI standard requires the handler to be callable - so either be a method or implement __call__. Currently, in order to use Connexion apps with a WSGI handler, one would have to directly pass the MyApp.app attribute instead of the MyApp instance of the connexion.App class. 

Such behaviour should be discouraged as it can lead to unwanted modification of the attribute and limits the amount of backwards-compatible changes possible in the future. Also, the current implementation leads to unwanted side-effects on IIS / wfastcgi.py which is used on the Azure platform.

This small change makes the App class WSGI-compliant and reduces the use of directly referencing the app attribute from outside the class, which should be considered unsafe.

### DIFF
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -271,8 +271,8 @@ class App(object):
             http_server.serve_forever()
         else:
             raise Exception('Server %s not recognized', self.server)
- 
-    def __call__(self, environ, start_response):
+
+    def __call__(self, environ, start_response):  # pragma: no cover
         """
         Makes the class callable to be WSGI-compliant. As Flask is used to handle requests,
         this is a passthrough-call to the Flask callable class.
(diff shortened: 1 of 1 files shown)


hjacobs:
@tdpsk thanks for the PR. I like the idea in general :smile:, but sure what @jmcs and @rafaelcaricio think about it.
Please check the failing Travis CI build and add a test..

hjacobs:
@jmcs @rafaelcaricio any opinion on this PR?

jmcs:
:+1:

rafaelcaricio:
👍

ingiulio:
👍

PR #215, opened 2016-04-21
Copy spec before validation occurs

### LINKED ISSUE #214: swagger_spec_validator adds in x-scope property to definitions
When validating the swagger document the dictionary that is passed to `swagger_spec_validator.validator20` is modified. It is possible for that library to add in any properties it wants to the spec.

Expected behavior, validation doesn't add in extra properties like `x-scope` to references.

Actual behavior, validation adds in extra properties like `x-scope` to references which can cause issues when using swagger clients that rely on the swagger_spec validator (like bravado). If the validator sets `x-scope` to the empty string then the bravado client can't properly understand where those references point to.

This was in a new connexion project, the swagger document had parameters defined as references.  

### Additional info:

Output of the commands:

- `python --version`
`Python 3.5.1`
`Version: 1.0.89`

This is related to this issue though allowing only a copy of the spec to be passed into the validator is a good idea on this side as well. https://github.com/Yelp/swagger_spec_validator/pull/48

wasbazi:
Fixes #214 

 - Copies the spec and passes the copy to validator instead of passing the original spec used later in the code

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -11,11 +11,11 @@ Unless required by applicable law or agreed to in writing, software distributed 
  language governing permissions and limitations under the License.
 """
 
+import copy
 import json
 import logging
 import pathlib
 import sys
-import copy
 
 import flask
 import jinja2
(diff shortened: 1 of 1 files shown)


jmcs:
Thanks for the pull request.

Do you have any bug caused by not deep copying the specification?

Please note that we can only merge your pull request after the failing unit test is fixed.

rafaelcaricio:
@wasbazi  Why would you like to change that? Could you please elaborate in the problem you are having?

rafaelcaricio:
@wasbazi Sounds good, but you have to do:

```python
self.specification = compatibility_layer(spec)
```

Because Connexion expects that the status codes to be integer value, probably that is why the tests are failing now.

wasbazi:
Thanks for pointing that out, just noticed the failing build and will fix that!

rafaelcaricio:
👍

PR #224, opened 2016-05-11
Cleanup decorators

jmcs:
See https://www.quantifiedcode.com/app/project/6ca6bc4af4224f4586a74658f551a64f?groups=code_patterns%3A4ACGxFj1

I replaced the old style `%` format with the new (? is 9 years still "new") style format and made the header comparison easier to read and understand, while avoiding needless conversions to `set`.

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -66,10 +66,12 @@ class ResponseValidator(BaseDecorator):
 
         if response_definition and response_definition.get("headers"):
             response_definition_header_keys = response_definition.get("headers").keys()
-            if not all(item in headers.keys() for item in response_definition_header_keys):
-                raise NonConformingResponseHeaders(
-                    message="Keys in header don't match response specification. Difference: %s"
-                    % list(set(headers.keys()).symmetric_difference(set(response_definition_header_keys))))
+            missing_keys = response_definition_header_keys - headers.keys()
+            if missing_keys:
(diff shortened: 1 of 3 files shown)


rafaelcaricio:
👍

rafaelcaricio:
Are we having some problem with travis? It is taking long to run the tests.

connexion/decorators/response.py
```diff
@@ -65,11 +65,15 @@ def validate_response(self, data, status_code, headers):
                 raise NonConformingResponseBody(message=str(e))
 
         if response_definition and response_definition.get("headers"):
-            response_definition_header_keys = response_definition.get("headers").keys()
-            if not all(item in headers.keys() for item in response_definition_header_keys):
-                raise NonConformingResponseHeaders(
-                    message="Keys in header don't match response specification. Difference: %s"
-                    % list(set(headers.keys()).symmetric_difference(set(response_definition_header_keys))))
+            # converting to set is needed to support python 2.7
```
rafaelcaricio: Do we need this comment? If it is removed the tests are going to break, plus your commit message is clear to say why it was added.

jmcs: It's just a reminder to remove the casting as soon as we decide to drop python 2.7 (worse case scenario in [4 years](https://www.python.org/dev/peps/pep-0373/)).

PR #227, opened 2016-05-16
Fix array conversion during validation

gredin:
I noticed this bug while testing a swagger.yaml that included `minItems` and `maxItems` properties on an `array` parameter.
It is quite obvious and self-explanatory.


jmcs:
:+1:

jmcs:
Thanks for the PR.

gredin:
Thanks for the release. Any idea when a release will include this fix?

rafaelcaricio:
@gredin we should get a new version later today.

hjacobs:
@gredin @rafaelcaricio released here: https://github.com/zalando/connexion/releases/tag/1.0.95

rafaelcaricio:
Faster than I thought. 👍

PR #228, opened 2016-05-17
Fix for when parameter spec is unordered.

### LINKED ISSUE #219: Skipped Parameter Validation
In the _validation.py_ file there is a _ParameterValidator_ class with follow line in it's init method:
```python
self.parameters = {k: list(g) for k, g in itertools.groupby(parameters, key=lambda p: p['in'])}
```

Here `parameters` is a list of parameters for an endpoint from the swagger specification. The expected behaviour is that the parameters get sorted into a dictionary based on their type. But the _groupby_ has a caveat which is mentioned in the documentation.

>  The operation of groupby() is similar to the `uniq` filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function). That behavior differs from SQL’s GROUP BY which aggregates common elements regardless of their input order.

When the parameters are sorted there is no issue. However when the parameters are mixed when the groupby() function is run it results in:

```python
('query', <itertools._grouper object at 0x10c6a4690>)
('path', <itertools._grouper object at 0x10c6a4a10>)
('query', <itertools._grouper object at 0x10c6a4690>)
```

So the first set of query parameters are overwritten by the second set of parameters. This means that the first set of parameters are not validated at all.

I can submit a PR with a fix in two ways. Either sort the parameters like so:
```python
self.parameters = {k: list(g) for k, g in itertools.groupby(parameters.sort(key=lambda p:p['in']), key=lambda p: p['in'])}
```
Or create a dictionary like so:
```python
self.parameters = defaultdict(list)
for p in parameters:
    self.parameters[p['in]].append(p)
```

arjunrn:
Fixes #219  .

 - Parameter validation dictionary is built with a default dict instead of `groupby` from `itertools`.
 - Tests to verify that the validation now really works.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -11,9 +11,9 @@ Unless required by applicable law or agreed to in writing, software distributed 
  language governing permissions and limitations under the License.
 """
 
+import collections
 import copy
 import functools
-import itertools
 import logging
 import sys
 
@@ -157,7 +157,9 @@ class ResponseBodyValidator(object):
 
 class ParameterValidator(object):
     def __init__(self, parameters):
-        self.parameters = {k: list(g) for k, g in itertools.groupby(parameters, key=lambda p: p['in'])}
+        self.parameters = collections.defaultdict(list)
+        for p in parameters:
+            self.parameters[p['in']].append(p)
 
     @staticmethod
     def validate_parameter(parameter_type, value, param):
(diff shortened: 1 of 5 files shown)


hjacobs:
Thanks for the PR.

hjacobs:
:+1:

PR #236, opened 2016-06-01
Don't send back entire API schema to clients after validation failures

### LINKED ISSUE #234: Don't send back entire schema after validation errors
### Description

When Connexion validates incoming data and finds a property is missing, it will dutifully report this back to the caller. However, it will include the entire swagger spec/schema as part of the response. For large APIs, this can be tens of KB of text, which is not ideal for debugging or logging.

### Expected behaviour

Expect to only receive the error portion, eg: "Missing required property 'email'"

### Actual behaviour

I receive the error portion following by a long JSON string that is my Swagger spec.

### Steps to reproduce

The actual generation of this error appears to occur in *jsonschema*, but this can be solved within Connexion by only sending back the `.message` attribute in the `problem()` call, rather than the string-representation of the entire exception object.

    def validate_schema(self, data):
        """
        :type schema: dict
        :rtype: flask.Response | None
        """
        if self.is_null_value_valid and is_null(data):
            return None

        try:
            logger.error("%s validation error: %s" % (flask.request.url, exception))
            return problem(400, 'Bad Request', str(exception))

        return None

Instead, the second-to-last line could read:

            return problem(400, 'Bad Request', exception.message)

Which will return the expected (and succinct) error message without the Swagger spec included.

### Additional info:

Output of the commands:

- `python --version`
  - Python 3.5.1

  - Version: 1.0.88

jvinet:
Fixes #234  .

 - For request body validations, instead of sending back the entire API schema, Connexion will only send back the error message (eg: "'value' is a required property"
 - For response validations, Connexion will continue to log the entire API schema via the `logger.error` call, as it did before.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -127,8 +127,8 @@ class RequestBodyValidator(object):
             validate(data, self.schema, format_checker=draft4_format_checker)
         except ValidationError as exception:
             logger.error("{url} validation error: {error}".format(url=flask.request.url,
-                                                                  error=exception))
-            return problem(400, 'Bad Request', str(exception))
+                                                                  error=exception.message))
+            return problem(400, 'Bad Request', str(exception.message))
 
         return None
 


rafaelcaricio:
Thank you for the PR. Could you please add a simple test for that?

focusend:
Please make this move forward. Makes no sense to spam production logs with full schemas.

rafaelcaricio:
👍

jmcs:
:+1:

PR #238, opened 2016-06-02
SQLAlchemy based example

nmusatti:
This is a simple example of how SQLAlchemy might be used to provide a backing store for a Connexion based application. It is based on Henning Jacobs's [Connexion Example REST Service](https://github.com/hjacobs/connexion-example).

### DIFF
--- a/examples/sqlalchemy/orm.py
+++ b/examples/sqlalchemy/orm.py
@@ -1,0 +1,35 @@
+from sqlalchemy import create_engine, Column, DateTime, String
+from sqlalchemy.orm import scoped_session, sessionmaker
+from sqlalchemy.ext.declarative import declarative_base
+
+
+Base = declarative_base()
+
+
+class Pet(Base):
+    __tablename__ = 'pets'
+    id = Column(String(20), primary_key=True)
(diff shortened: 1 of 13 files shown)


hjacobs:
Cool, thanks for the PR!

examples/sqlalchemy/orm.py
```diff
@@ -0,0 +1,35 @@
...
+            created=None):
+        if name is not None:
+            self.name = name
+        if animal_type is not None:
+            self.animal_type = animal_type
+        if created is not None:
+            self.created = created
+        
+    def dump(self):
+        return dict([(k,v) for k,v in self.__dict__.items() if k[0] != '_'])
```
jmcs: You can replace `self.__dict__` with `vars(self)` to avoid explicitly getting a dunder property.

hjacobs: and `k[0] != ...` is better replaced with `k.startswith(...)`

nmusatti:
I applied the suggested changes and corrected import order, as signalled by the first test run.

hjacobs:
:+1:

PR #243, opened 2016-06-07
Add link to official spec

rafaelcaricio:
I want to add a link to direct users to the proper part of the spec where is possible to find more information about parameters specially the `collectionFormat` definitions.

 - Better docs.


hjacobs:
:+1:

PR #247, opened 2016-06-10
Stricter requirements.txt

### LINKED ISSUE #246: Stricter requirements.txt?
The recent release of flask 0.11.0 required Connexion changes, per #240.

Since your `requirements.txt` says [`flask>=0.10.1`](https://github.com/zalando/connexion/blob/1.0.88/requirements.txt), the result was that all versions of Connexion that do not include that fix are broken.

As it happens this isn't a big deal this time; upgrading has just been a matter of figuring out what has happened and moving to a more recent version of Connexion.  But if, for instance, Connexion had changed more in the meantime then it might easily have been awkward.

This project can't predict what breaking changes might be released by upstream projects, so it would be more prudent to make upgrade of dependencies be explicit and not, as this time, accidental.

Therefore suggest that requirements.txt should contain exact rather than minimal requirements.  

(Probably it's OK to allow variation in minor releases so long as we can trust upstream to have some adherence to semantic versioning so eg something like `flask>=0.10.1,<0.10.99` would be reasonable).

dimbleby:
Fixes #246: avoid being exposed to API changes in dependencies.

- provide upper bounds on dependencies
- remove `strict-rfc3339` dependency - doesn't seem to be required
- explicitly specify `jinja2` and `werkzeug` dependencies
- I haven't added dependencies for optional imports `gevent`, `tornado`, `uwsgi_metrics`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,9 @@
-jsonschema>=2.5.1
-flask>=0.10.1
-PyYAML>=3.11
-requests>=2.9.1
-six>=1.7
-strict-rfc3339>=0.6
-swagger_spec_validator>=2.0.2
+jsonschema>=2.5.1,<3
+flask>=0.10.1,<0.12
+PyYAML>=3.11,<4
+requests>=2.9.1,<3
+six>=1.7,<2
+strict-rfc3339>=0.6,<0.7
+swagger_spec_validator>=2.0.2,<3
+jinja2>=2.8,<3
+werkzeug>=0.11.10,<0.12


dimbleby:
Apparently `strict-rfc3339` really is needed, so I've reinstated it.

requirements.txt
```diff
@@ -1,7 +1,9 @@
...
-flask>=0.10.1
-PyYAML>=3.11
-requests>=2.9.1
-six>=1.7
-strict-rfc3339>=0.6
-swagger_spec_validator>=2.0.2
+jsonschema>=2.5.1,<3
+flask>=0.10.1,<0.12
+PyYAML>=3.11,<4
+requests>=2.9.1,<3
```
rafaelcaricio: How does it prevent that the next version of "requests" that could be `2.10.0` from braking Connexion? It applies to all other requirements.

This PR does not make much difference for the problem you are trying to solve.

dimbleby: `requests` [uses semantic versioning](http://docs.python-requests.org/en/master/dev/philosophy), so if the major version number does not change then that's an explicit promise that the code is back-compatible.

rafaelcaricio: Have you checked if all dependencies follow that? Have you installed each combination to check which version ranges Connexion is compatible?

requirements.txt
```diff
@@ -1,7 +1,9 @@
...
+six>=1.7,<2
+strict-rfc3339>=0.6,<0.7
```
rafaelcaricio: How is your approach better than pointing to a specific version?

requirements.txt
dimbleby: I admit that I haven't, though it is a very common convention and quite likely to be the case - especially for the large and well-known projects that `connexion` depends on.  

If `connexion` does have dependencies that are willing to break APIs without signaling this in the release number, then there isn't a lot that can be done to protect ourselves from that.

However let's not make the perfect the enemy of the good...

requirements.txt
dimbleby: - It allows `connexion` to benefit from fixes eg `flask` have now released `0.11.1`.  
- Generally flexibility is desirable (where it doesn't break things!) eg maybe my application really needed the bug fix in `flask 0.11.1`: then I'd have been annoyed if `connexion` had insisted on exactly `0.11.0`.

requirements.txt
rafaelcaricio: I am asking that because I actually interacted to each of those packages to set the minimal versions not so long ago.

PR #252, opened 2016-06-25
Open spec file as binary and then handle problems with encoding.

### LINKED ISSUE #249: UnicodeDecodeError: 'gbk' codec can't decode byte 0xaf in position
### Description
swagger-editer generator server python flask based connexion, when i run app.
### Expected bahaviour

  File "E:\eclipsephp\plugins\org.python.pydev_4.5.4.201601292234\pysrc\pydevd.py", line 1524, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "E:\eclipsephp\plugins\org.python.pydev_4.5.4.201601292234\pysrc\pydevd.py", line 931, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "E:\eclipsephp\plugins\org.python.pydev_4.5.4.201601292234\pysrc\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "E:\pworkspace\testrest\examples\python-flask-server\app.py", line 6, in <module>
    app.add_api('swagger.yaml', arguments={'title': ''})
  File "E:\pythonvenv\flaskvenv\lib\site-packages\connexion\app.py", line 146, in add_api
    debug=self.debug)
  File "E:\pythonvenv\flaskvenv\lib\site-packages\connexion\api.py", line 89, in __init__
    swagger_template = swagger_yaml.read()
UnicodeDecodeError: 'gbk' codec can't decode byte 0xaf in position 266: illegal multibyte sequence

### Steps to reproduce
orig:
        with swagger_yaml_path.open() as swagger_yaml:
            swagger_template = swagger_yaml.read()
change:
        with swagger_yaml_path.open(encoding='utf-8') as swagger_yaml:
            swagger_template = swagger_yaml.read()

and then is run success.
### Additional info:

Output of the commands:

- `python --version`3.5.1
connexion-1.0.97
in windows10 64 run

rafaelcaricio:
Fixes #249 .

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -84,12 +84,17 @@ class Api(object):
                             'swagger_url': swagger_url,
                             'auth_all_paths': auth_all_paths})
         arguments = arguments or {}
-        with swagger_yaml_path.open() as swagger_yaml:
-            swagger_template = swagger_yaml.read()
+        with swagger_yaml_path.open(mode='rb') as swagger_yaml:
+            contents = swagger_yaml.read()
+            try:
+                swagger_template = contents.decode()
+            except UnicodeDecodeError:
+                swagger_template = contents.decode('utf-8', 'replace')
+
             swagger_string = jinja2.Template(swagger_template).render(**arguments)
             self.specification = yaml.load(swagger_string)  # type: dict
 
-        logger.debug('Read specification', extra=self.specification)
+        logger.debug('Read specification', extra={'spec': self.specification})
 
         self.specification = compatibility_layer(self.specification)
         # Avoid validator having ability to modify specification
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -36,6 +36,9 @@ logger = logging.getLogger('connexion.api')
 
 def compatibility_layer(spec):
     """Make specs compatible with older versions of Connexion."""
+    if not isinstance(spec, dict):
+        return spec
+
     # Make all response codes be string
     for path_name, methods_available in spec.get('paths', {}).items():
         for method_name, method_def in methods_available.items():
@@ -92,7 +95,7 @@ class Api(object):
                 swagger_template = contents.decode('utf-8', 'replace')
 
             swagger_string = jinja2.Template(swagger_template).render(**arguments)
-            self.specification = yaml.load(swagger_string)  # type: dict
+            self.specification = yaml.safe_load(swagger_string)  # type: dict
 
         logger.debug('Read specification', extra={'spec': self.specification})
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -5,6 +5,7 @@ import tempfile
 
 from connexion.api import Api
 from swagger_spec_validator.common import SwaggerValidationError
+from yaml import YAMLError
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


tech-cd:
We should keep the old behaviour but receive the encoding as an optional
argument (with utf8 as default).

rafaelcaricio:
The old behaviour is kept with this PR and there is no need for one more option, IMHO.

connexion/api.py
```diff
@@ -84,12 +84,17 @@ def __init__(self, swagger_yaml_path, base_url=None, arguments=None,
...
                             'auth_all_paths': auth_all_paths})
         arguments = arguments or {}
-        with swagger_yaml_path.open() as swagger_yaml:
-            swagger_template = swagger_yaml.read()
+        with swagger_yaml_path.open(mode='rb') as swagger_yaml:
+            contents = swagger_yaml.read()
+            try:
+                swagger_template = contents.decode()
+            except UnicodeDecodeError:
+                swagger_template = contents.decode('utf-8', 'replace')
```
jmcs: What it it's an encoding that's completely incompatible with ASCII/UTF-8?

connexion/api.py
```diff
@@ -84,12 +84,17 @@ def __init__(self, swagger_yaml_path, base_url=None, arguments=None,
                             'swagger_url': swagger_url,
                             'auth_all_paths': auth_all_paths})
         arguments = arguments or {}
-        with swagger_yaml_path.open() as swagger_yaml:
-            swagger_template = swagger_yaml.read()
+        with swagger_yaml_path.open(mode='rb') as swagger_yaml:
```
jmcs: It's better to receive an optional encoding and then pass it to open.

connexion/api.py
```diff
@@ -84,12 +84,17 @@ def __init__(self, swagger_yaml_path, base_url=None, arguments=None,
...
+            contents = swagger_yaml.read()
+            try:
+                swagger_template = contents.decode()
+            except UnicodeDecodeError:
+                swagger_template = contents.decode('utf-8', 'replace')
+
             swagger_string = jinja2.Template(swagger_template).render(**arguments)
             self.specification = yaml.load(swagger_string)  # type: dict
```
jmcs: Unrelated to this issue, but it would be better to use safe_load instead of load.

connexion/api.py
rafaelcaricio: Then the characters will be replaced by ?.

jmcs: Which would mean the yaml.load would fail for those encodings

rafaelcaricio: No, because the keys need to be valid by Swagger/OpenAPI spec anyway. Please check the test with a invalid character. The yaml.load does not fail.

rafaelcaricio: Could you please provide an "encoding that's completely incompatible with ASCII/UTF-8" so we can add a test case for that and adapt the code to handle this situation?

jmcs:
:+1:

jmcs:
:+1:

PR #253, opened 2016-06-29
Fixed a wrong example of README

pistatium:
Because the document had a trifling mistake, I made modifications.


rafaelcaricio:
@pistatium Looks like [the default is "search"](https://github.com/zalando/connexion/blob/master/connexion/resolver.py#L80) so the docs should be right. Could you provide more details on your issue?

pistatium:
I see.I misunderstood it. sorry

PR #257, opened 2016-07-27
Fix security check

### LINKED ISSUE #256: OAuth token should have all scopes listed in swagger spec to grant access
Hi, I have a question regarding implementation:
https://github.com/zalando/connexion/blob/master/connexion/decorators/security.py#L88
checks that access token must have any scope from the list of scopes.

Should the check be more strict, by checking that declared scopes is a subset of token's scopes?

rafaelcaricio:
Fixes #256

 - Check for all defined scopes.


hjacobs:
:+1:

PR #271, opened 2016-09-01
Support operation functions with **kwargs

### LINKED ISSUE #270: Support *args and **kwargs in operation function signatures
### Description

This is a follow up to #268 to allow decorating operation functions where the decorator changes the signature to generic `*args`, `**kwargs`.

### Expected behaviour

Parameters should be mapped for operation functions with `*args` and/or `**kwargs`. 

### Actual behaviour

Parameters not directly matching the function signature are silently dropped (`kwargs` are not populated).

### Steps to reproduce

* Create a simple Swagger YAML with one operation having at least one query parameter "foo". 
* Define the operation function with only `*args` and `**kwargs`.
* The Swagger query parameter "foo" is not passed in `**kwargs`.

hjacobs:
Fixes #270.

* support operation functions with `**kwargs`

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -25,12 +25,15 @@ def get_function_arguments(function):  # pragma: no cover
     Returns the list of arguments of a function
 
     :type function: Callable
-    :rtype: list[str]
+    :rtype: tuple[list[str], bool]
     """
     if six.PY3:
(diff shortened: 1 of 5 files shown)


hjacobs:
:+1:

victorzhao1990:
Does this only work for python3?

PR #272, opened 2016-09-01
Fix setup.py to work with non-numeric Python patchlevel

### LINKED ISSUE #260: python setup.py install fails with: ValueError: invalid literal for int() with base 10: '6rc1'
### Description

pip install fails

### Expected bahaviour

Install complete

### Actual behaviour

Russells-MacBook-Pro-OLD-506:connexion-1.0.109 rjurney$ python setup.py install
  File "setup.py", line 23, in <module>
    py_major_version, py_minor_version, _ = (int(v.rstrip('+')) for v in platform.python_version_tuple())
  File "setup.py", line 23, in <genexpr>
    py_major_version, py_minor_version, _ = (int(v.rstrip('+')) for v in platform.python_version_tuple())
ValueError: invalid literal for int() with base 10: '6rc1'

### Steps to reproduce

pip install connexion

### Additional info:

Output of the commands:

- `python --version` Python 2.7.6rc1
- `pip show connexion | grep "^Version\:"` Nothing

hjacobs:
Fixes #260 .

* Ignore `patchlevel` (which might be non-numeric)
* Simplify comparison

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,13 +20,13 @@ def read_version(package):
 
 version = read_version('connexion')
 
-py_major_version, py_minor_version, _ = platform.python_version_tuple()
+py_major_minor_version = tuple(int(v.rstrip('+')) for v in platform.python_version_tuple()[:2])
 
 
 def get_install_requirements(path):
     content = open(os.path.join(__location__, path)).read()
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

PR #273, opened 2016-09-04
Resolve JSON pointers in swagger spec

patrickw276:
Add support for resolving local file references in swagger specs (#254).

Edit: I'm having some trouble getting the build to pass... I think it's just isort complaining about import order. I ran isort on the two modules in my PR and it's still not happy.

### DIFF
--- a/connexion/swagger.py
+++ b/connexion/swagger.py
@@ -1,0 +1,197 @@
+"""
+Copyright 2015 Zalando SE
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the
+License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ language governing permissions and limitations under the License.
+"""
+
+import os.path
+from collections import deque, namedtuple
+from copy import deepcopy
+from pathlib import Path
+
+import yaml
+from connexion.exceptions import InvalidSpecification
+from jinja2 import Template
+
+
+def load(path, template_args):
+    """
+    Load a swagger spec.
+
+    :param path: path to spec
+    :type path: pathlib.Path
+    :param template_args: jinja template args
+    :type template_args: dict
+
+    :rtype dict
+    """
+    resolver = Resolver(template_args)
+    pointers = inventory(path, resolver)
+
+    root_data = deepcopy(resolver.get(path, []))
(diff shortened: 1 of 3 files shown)


hjacobs:
@patrickw276 thanks for the PR! We'll look into it...

connexion/swagger.py
```diff
@@ -0,0 +1,197 @@
+"""
```
hjacobs: you can remove the license header as the `LICENSE` file is enough (and we should probably remove all existing file headers in a later PR)

patrickw276:
I just realized that this code creates recursive dictionaries when resolving circular pointers contained in any file except the root. These recursive dictionaries will crash the swagger validator. My use case doesn't include circular references but I guess it's feasible that someones will. I don't mind fixing this but it may take me some time before I can get around to it. 

Would it be acceptable to raise an exception stating that circular references outside of root are not currently supported?

jfinkhaeuser:
One of the annoying things with this is that the jsonschema validator underlying the validation does in fact (partially) resolve references.

I've got a parser that resolves references. I have this in an as-yet unreleased package related to API testing. I could factor it into its own package, if you'd like to use it.

patrickw276:
@jfinkhaeuser Yeah, I knew going into writing this code that the jsonschema validator resolves references but I don't think the swagger spec validator does. Do you think a better approach would be leaving the references in and having the various libraries resolve them on their own? It seems like at some point the schema will need to be resolved into a single dictionary for connexion to pass around for routing, etc...

Is your parser on github somewhere? I'd be happy to take a look at it/potentially use it.

jfinkhaeuser:
I'll make it available in the next few days, then, and link it here. It's on github, but the repo is private for the moment - I need to move it, package it, etc.

jfinkhaeuser:
Try https://pypi.python.org/pypi/prance/0.1.0 and/or https://github.com/jfinkhaeuser/prance

It might not be what you're looking for. If you have additional test cases you need to work, let me know.

patrickw276:
After more thought, I don't like my approach in this PR. I looked more into swagger UI and it can resolve relative references. I'm going to see what it would take to resolves references as needed and not bundle everything into a single schema on load.

jfinkhaeuser:
IMHO the problem with resolving references as needed is that for validation, you need to resolve them anyway. If you want to validate on load, then you have to resolve on load. If you want to validate at run-time later on, then of course resolving later is just fine.

That's just my 2c, of course.

patrickw276:
@jfinkhaeuser I'll have to dig into it a bit more but I think resolving on load would be better/safer. There's a good chance a different approach will use a lot of the same code (or your library, ty for making it available btw). Once I realized that the swagger UI supports relative references, bundling all documents into one seemed like too hacky of a solution.

Issue #250, opened 2016-06-15
Raise original exception while importing modules


tomas-fp:
### Description

Currently, when importing Python modules dynamically while binding routes to Flask, if something is wrong with the loaded code there's only a generic `connexion` exception raised, so the original error is hidden and therefore there's not easy way to see what was the real error.

### Expected behaviour

Catch the original Python exception and raise it properly exposing it to higher layers instead of raising a generic error who tells you nothing.

### Actual behaviour

This is the error stack trace, which is not exposing the real problem:
```
ERROR:connexion.api:Failed to add operation for DELETE /api/reports/{report}
Traceback (most recent call last):
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/api.py", line 196, in add_paths
    self.add_operation(method, path, endpoint, path_parameters)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/api.py", line 169, in add_operation
    resolver=self.resolver)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/operation.py", line 172, in __init__
    resolution = resolver.resolve(self)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/resolver.py", line 51, in resolve
    return Resolution(self.resolve_function_from_operation_id(operation_id), operation_id)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/resolver.py", line 72, in resolve_function_from_operation_id
    return self.function_resolver(operation_id)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/utils.py", line 109, in get_function_from_name
    function = deep_getattr(module, attr_path)
  File "/Users/taparicio/.pyenv/versions/3.5.1/lib/python3.5/site-packages/connexion/utils.py", line 91, in deep_getattr
    return functools.reduce(getattr, attr.split('.'), obj)
AttributeError: module 'api' has no attribute 'reports'
```

### Steps to reproduce

Just try breaking your dynamically loaded code.

### Additional info:

Output of the commands:

- Python `3.5.1`.
- Version: 1.0.98

rafaelcaricio:
[...]

PR #274, opened 2016-09-07
Fix operation ID handling

jfinkhaeuser:
Fixes error handling when dealing with bad operation IDs. Any of the following now lead to a ResolverError, which is easier to understand in the server logs:

- A module that fails to import cleanly (e.g `fakeapi/module_with_error.py`)
- A completely missing operationId
- An operationId that specifies an unresolvable module/function

Mostly, the last class leads to very bad error messages about NoneType not having an rsplit function - completely true, but also devoid of any useful information.

I'll follow up with another PR that uses this change, too.

(Note: the change to setup.py is technically out of scope, but it's a one-liner. It ensures that when you're using virtualenv, `python setup.py test` doesn't fail because some library in the virtualenv path breaks the tests. If you insist, I'll make it a separate PR).


jfinkhaeuser:
FWIW, the failed travis test doesn't look like it's got anything to do with my changes. Let me know if it does and I just don't get it.

hjacobs:
Does this conflict with #250?

jfinkhaeuser:
@hjacobs Interesting. It might; however, the original error should still get logged.

My main issue was that errors about NoneType not having an rsplit function are really obscure for the user and only reflect internals of connexion. That should IMHO not happen.

jfinkhaeuser:
Mhh, looking at #250 again, it really shouldn't conflict, but it doesn't solve that other issue either.

But since I'm at it anyway, I might as well look into a PR for that issue, too.

jfinkhaeuser:
That last change should work better with the complaint raised in #250

rafaelcaricio:
@jfinkhaeuser Since this PR was for some time without update and still with failing checks I stepped forward and sent to you a new PR (https://github.com/jfinkhaeuser/connexion/pull/1) to cover issues I could see in this PR. Please consider accepting it so we can merge your PR to Connexion. :zap:

rafaelcaricio:
I think we are experiencing a intermittent error in our tests. I do not know yet what is causing it. 😞

tox.ini
```diff
@@ -3,7 +3,7 @@ max-line-length=120
 exclude=connexion/__init__.py
 
 [tox]
-envlist=pypy,py27,py34,py35,py35-flask0.10.1,isort-check,flake8
+envlist=py27,py34
```
rafaelcaricio: 😨

rafaelcaricio: Please merge this fix https://github.com/jfinkhaeuser/connexion/pull/2.

jfinkhaeuser:
Hmm, the isort errors I *think* came from my merges from master.

This last error appears to appear more consistently, no? I'm looking into it right now.

rafaelcaricio:
It is not consistent. But if you use this seed, it will happen:

```properties
[testenv]
...
setenv=
    PYTHONHASHSEED=1584359276
```
add to `tox.ini` file.

jfinkhaeuser:
I'll try that.

I'm also getting errors for flake8 and py34 parsing README.rst as ascii in setup.py

rafaelcaricio:
Check if you are using the latest Python versions of the interpreter. It influences the flake8 and isort results.

jfinkhaeuser:
I am, I'm also using tox.

README.rst does contain utf-8, so it makes sense for encoding-aware file I/O to fail. With that fixed and the `tox.ini` changes, I'm getting the same errors as on Travis.

rafaelcaricio:
Good! Yep, I also have the same error as on Travis. You can see that every time the tests either pass or return error without code changes.

jfinkhaeuser:
This is really weird, though. It's raising an ImportError, catching it immediately after, and the tests fail with an uncaught ImportError?

rafaelcaricio:
Seems like so, I am also trying to understand what is going on. :disappointed:

jfinkhaeuser:
See, I don't get the logic in that bit.

First off, the ImportError is re-raised. It's absolutely correct that it was raised in the first place, because the module in question doesn't exist.

But what we're getting in the error handling is that if a nested module import fails, it tries the next parent, etc. I don't understand why.

rafaelcaricio:
I think I get what is going on here. The problem is iterating over unordered data structure, the swagger.yaml file, specifically the methods list. To solve that and have a consistent error I split the swagger specs.

jfinkhaeuser:
That's likely what's happening, yes.

But I still do not understand the reasoning behind https://github.com/zalando/connexion/pull/274/commits/7804fe13da3758be445da2d6d44cf46067a61798#diff-39dbdb8a5aaada24b36a6285b29318cdL107 - do you have any thoughts on this?

rafaelcaricio:
https://github.com/jfinkhaeuser/connexion/pull/3

rafaelcaricio:
@jfinkhaeuser This code is what imports the module of the operation. I think I do not understand what is your question.

jfinkhaeuser:
My question is, why if `module.submodule.function` is to be resolved does it make sense to try importing `module` and looking for a `submodule.function` attribute within.

Is that a py27/py3.x compatibility thing? I would expect that if `module.submodule` import fails, that's all we need to know.

jfinkhaeuser:
Ah... you're trying to find static functions?

jfinkhaeuser:
You know... reading the doc string actually helps (doh).

rafaelcaricio:
So the idea here is to all of those cases:
- A module that fails to import cleanly (e.g `fakeapi/module_with_error.py`)
- A completely missing operationId
- An operationId that specifies an unresolvable module/function

To result in a `ResolverError`?

jfinkhaeuser:
Correct. The main purpose here was that previously, it would either result in `ImportError` or `AttributeError`, which isn't very good for handling the errors later on.

jfinkhaeuser:
So, here's the issue:

```patch
diff --git a/tests/test_api.py b/tests/test_api.py
index 8398ef7..f9fffa6 100644
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -41,7 +41,7 @@ def test_invalid_operation_does_stop_application_to_setup():
         Api(TEST_FOLDER / "fixtures/missing_op_id/swagger.yaml", "/api/v1.0",
             {'title': 'OK'})
 
-    with pytest.raises(ResolverError):
+    with pytest.raises(ImportError):
         Api(TEST_FOLDER / "fixtures/module_not_implemented/swagger.yaml", "/api/v1.0",
             {'title': 'OK'})
``` 
That change would fix the tests.

What I tried to do - because I'm aware of backwards compatibility issues - is not hide the original error  in non-debug mode. Someone else complained about that. So the `ResolverError` carries information about the original error, which is then raised in non-debug mode.

In debug mode, ResolverError is just ignored (or rather, a later PR handles them as 501 response codes).

rafaelcaricio:
Ok, makes sense to me.

rafaelcaricio:
:+1:

PR #275, opened 2016-09-07
Add ability to add random string to endpoint names

jfinkhaeuser:
Swagger is clear on the fact that operationIds must be unique across a spec. That leads to connexion assuming that there is no reason for the same Python function to be mapped to multiple paths, and using the output of `flaskify_endpoint` for unique flask view names.

Now this is correct as connexion is currently implemented. I'm preparing another PR based on #274 that would no longer make this assumption always true.

 - Add the ability to *optionally* add random characters to a flask view name

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -45,9 +45,9 @@ def flaskify_endpoint(identifier, randomize=None):
     if randomize is None:
         return result
 
-    def generator(size=randomize, chars=string.ascii_uppercase + string.digits):
-        return ''.join(random.SystemRandom().choice(chars) for _ in range(size))
-    return result + '|' + generator()
+    chars = string.ascii_uppercase + string.digits
+    return result + '|' + ''.join(
+        random.SystemRandom().choice(chars) for _ in range(randomize))
 
 
 def convert_path_parameter(match, types):
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -28,14 +28,25 @@ PATH_PARAMETER_CONVERTERS = {
 }
 
 
-def flaskify_endpoint(identifier):
+def flaskify_endpoint(identifier, randomize=None):
     """
     Converts the provided identifier in a valid flask endpoint name
 
     :type identifier: str
+    :param randomize: If specified, add this many random characters (upper case
+        and digits) to the endpoint name, separated by a pipe character.
+    :type randomize: int | None
     :rtype: str
     """
-    return identifier.replace('.', '_')
+    result = identifier.replace('.', '_')
+    if randomize is None:
+      return result
+
+    import random
+    import string
+    def generator(size=randomize, chars=string.ascii_uppercase + string.digits):
+      return ''.join(random.SystemRandom().choice(chars) for _ in range(size))
+    return  result + '|' + generator()
 
 
 def convert_path_parameter(match, types):
@@ -47,7 +58,7 @@ def convert_path_parameter(match, types):
                                 name.replace('-', '_'))
 
 
-def flaskify_path(swagger_path, types=None):
+def flaskify_path(swagger_path, types=None, method=None):
     """
     Convert swagger path templates to flask path templates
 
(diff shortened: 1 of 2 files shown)


rafaelcaricio:
@jfinkhaeuser thanks for the PR. But could you please explain what is a user case for this PR?

connexion/utils.py
```diff
@@ -28,14 +28,26 @@
...
+    result = identifier.replace('.', '_')
+    if randomize is None:
+        return result
+
+    import random
+    import string
+
+    def generator(size=randomize, chars=string.ascii_uppercase + string.digits):
+        return ''.join(random.SystemRandom().choice(chars) for _ in range(size))
+    return result + '|' + generator()
```
rafaelcaricio: This function supposed to return a valid python function name and `|` is not a valid character.

connexion/utils.py
```diff
@@ -28,14 +28,26 @@
...
+    :type randomize: int | None
     :rtype: str
     """
-    return identifier.replace('.', '_')
+    result = identifier.replace('.', '_')
+    if randomize is None:
+        return result
+
+    import random
+    import string
```
rafaelcaricio: Imports should be in the top level and sorted properly.

connexion/utils.py
jfinkhaeuser: That's the point of this change. There's no need for it to return a valid python. See #278 for how it's used.

connexion/utils.py
jfinkhaeuser: Yeah, that's easily enough fixed. I'll put it on my list.

rafaelcaricio:
Ok, I see what is your suggestion here.

jfinkhaeuser:
Since that got lost (outdated diff and all that), the use case is #278.

`flaskify_endpoint` returns a string that's supposed to be used by flask as a view name. Flask itself cares about unique view names, but multiple view names can map to the same Python function.

Swagger is a bit stricter, in that each operationId must be unique, and operationIds are interpreted as function names. So the previous implementation makes perfect sense from a Swagger point of view, but not so much from a Flask point of view.

So what's the use case?

In order not to replicate a huge amount of code, #278 installs the same error handler for all operations with a missing operationId, which generates a 501. The PR hooks into things at a point where connexion still interprets the operationId *a little*, i.e. expects a unique one, but then just passes it on as a flask view name.

#278 *could* be refactored. But when I looked into that, the code changes would have been much bigger, and re-use far less of what's already there.

rafaelcaricio:
:+1:

connexion/utils.py
```diff
@@ -28,14 +31,23 @@
...
+        and digits) to the endpoint name, separated by a pipe character.
+    :type randomize: int | None
     :rtype: str
     """
-    return identifier.replace('.', '_')
+    result = identifier.replace('.', '_')
+    if randomize is None:
+        return result
+
+    def generator(size=randomize, chars=string.ascii_uppercase + string.digits):
```
hjacobs: any reason why this is a local function (and not just a local variable holding the random string)?

jfinkhaeuser: No good one, mostly this was too much for me for a single line:

```python
''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(randomize))
```

But there's a few other ways it could be broken up, sure.

rafaelcaricio: I am not 100% sure but I assumed it is like that for code clarity, everything inside that is related to generating a random string. The code wouldn't be very clear without this contextualized part.

jfinkhaeuser: See 246d566 (I can revert that again, of course).

hjacobs:
:+1:

connexion/utils.py
```diff
@@ -28,14 +31,23 @@
...
+    :type randomize: int | None
     :rtype: str
     """
-    return identifier.replace('.', '_')
+    result = identifier.replace('.', '_')
+    if randomize is None:
+        return result
+
+    chars = string.ascii_uppercase + string.digits
+    return result + '|' + ''.join(
```
rafaelcaricio: Since you changed, maybe then move this concatenation (` + '|' + `) to the `''.join(...` ?

PR #276, opened 2016-09-07
It's bad practice to derive custom exception classes from BaseException.

jfinkhaeuser:

 - Derive from Exception rather than BaseException (see https://www.python.org/dev/peps/pep-0352/)
 - Change the expectation that `InvalidSpecification` is only raised outside debug mode.

Note: the latter change is in line with how exceptions are generally handled in connexion. It'd be easy enough to pass this particular error through anyway.


rafaelcaricio:
:+1:

PR #277, opened 2016-09-07
Fix tests

jfinkhaeuser:
- Not specifying `tests` as the test directory will lead to pytest trying to run test code in dependencies installed in the project directory via `virtualenv`.
- `README.rst` is in utf-8 encoding; reading it fails on py34


jfinkhaeuser:
Probably merge this before #274 - it contains a change that the other PR also contains.

rafaelcaricio:
👍

PR #284, opened 2016-09-12
Run an API spec, no boilerplate needed

rafaelcaricio:
Currently creating or developing an API with Connexion means that you need to Copy&Paste® a little boilerplate (small, but still) code.

```
import connexion

app = connexion.App(__name__, specification_dir='swagger/')
app.add_api('my_api.yaml')
app.run(port=8080)
```

Plus would be nice to configure some logging, more Copy&Paste®.

```
import logging
logging.basicConfig(level=logging.DEBUG)
```

Then would be nice to have a command line to run the project. Let me Copy&Paste® this little CLI tool boilerplate code structure.

```
import click
@click.command()
def run():
    ...
```

**Alternatively**, we could provide this minimal set of common practices of creating apps using Connexion bundled already within the installation of Connexion itself. To create a Microservice using Connexion, developers would only need to worry about the OpenAPI specification and the actual business logic/Python code. Then run the server using:

```
$ connexion run my_api.yaml
```
or
```
$ cnx run my_api.yaml -p 8080 --debug
```

Other use case is when during the development we want to run the server with partially implemented endpoints. Just run it:

```
$ cnx run my_unfinished_api.yaml --stub
```

Endpoints not implemented yet will return an error message saying that the operation is not yet implemented.

Those changes will make Connexion sail smoothly and providing fast satisfactory results will definitely improve the success of Connexion.

***Important to notice***
This is not a breaking change and all other functionality continues to work the same.

 - Adds a the class `StubResolver` to provide a easy way to stub not implemented or not found operations;
 - Adds a command line tool for running specifications (`connexion` or `cnx`).

I hope you will like it! ❤️

### DIFF
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -13,26 +13,17 @@ main = AliasedGroup(context_settings=dict(help_option_names=[
     '-h', '--help']))
 
 
-def _operation_not_implemented():
-    return problem(
-        title='Not Implemented Yet',
-        detail='The requested functionality is not implemented yet.',
-        status=400)
-
-
 def validate_wsgi_server_requirements(ctx, param, value):
     if value == 'gevent':
         try:
-            import gevent.wsgi  # NOQA
+            import gevent  # NOQA
         except:
             fatal_error('gevent library is not installed')
     elif value == 'tornado':
         try:
-            import tornado.wsgi  # NOQA
-            import tornado.httpserver  # NOQA
-            import tornado.ioloop  # NOQA
+            import tornado  # NOQA
         except:
-            fatal_error('tornado library not installed')
+            fatal_error('tornado library is not installed')
 
 
 @main.command()
@@ -91,7 +82,10 @@ def run(spec_file,
 
     resolver = None
     if stub:
-        resolver = StubResolver(_operation_not_implemented)
+        resolver = StubResolver(lambda: problem(
+            title='Not Implemented Yet',
+            detail='The requested functionality is not implemented yet.',
+            status=400))
 
     app = App(__name__)
     app.add_api(path.abspath(spec_file), resolver=resolver)
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -34,22 +34,28 @@ def validate_wsgi_server_requirements(ctx, param, value):
               type=click.Choice(['flask', 'gevent', 'tornado']),
               callback=validate_wsgi_server_requirements,
               help='Which WSGI server container to use.')
-@click.option('--stub', '-s',
+@click.option('--stub',
               help='Returns status code 400, and `Not Implemented Yet` payload, for '
               'the endpoints which handlers are not found.',
               is_flag=True, default=False)
 @click.option('--hide-spec',
               help='Hides the API spec in JSON format which is by default available at `/swagger.json`.',
-              is_flag=True, default=True)
+              is_flag=True, default=False)
 @click.option('--hide-console-ui',
               help='Hides the the API console UI which is by default available at `/ui`.',
-              is_flag=True, default=True)
+              is_flag=True, default=False)
 @click.option('--console-ui-url', metavar='URL',
               help='Personalize what URL path the API console UI will be mounted.')
 @click.option('--console-ui-from', metavar='PATH',
               help='Path to a customized API console UI dashboard.')
 @click.option('--auth-all-paths',
               help='Enable authentication to paths not defined in the spec.',
+              is_flag=True, default=False)
+@click.option('--validate-responses',
+              help='Enable validation of response values from operation handlers.',
+              is_flag=True, default=False)
+@click.option('--strict-validation',
+              help='Enable strict validation of request payloads.',
               is_flag=True, default=False)
 @click.option('--debug', '-d', help='Show debugging information.',
               is_flag=True, default=False)
@@ -63,6 +69,8 @@ def run(spec_file,
         console_ui_url,
         console_ui_from,
         auth_all_paths,
+        validate_responses,
+        strict_validation,
         debug):
     """
     Runs a server compliant with a OpenAPI/Swagger 2.0 Specification file.
@@ -89,4 +97,14 @@ def run(spec_file,
 
     app = App(__name__)
     app.add_api(path.abspath(spec_file), resolver=resolver)
-    app.run(port=port, server=wsgi_server)
+    app.run(
+        port=port,
+        server=wsgi_server,
+        swagger_json=hide_spec or None,
+        swagger_ui=hide_console_ui or None,
+        swagger_path=console_ui_from or None,
+        swagger_url=console_ui_url or None,
+        strict_validation=strict_validation,
+        validate_responses=validate_responses,
+        auth_all_paths=auth_all_paths,
+        debug=debug)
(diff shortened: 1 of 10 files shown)


hjacobs:
:+1: for having `connexion` as a runnable console script --- but I would only keep one, i.e. remove the `cnx` alias.

NOTE: I haven't reviewed the whole PR yet...

connexion/cli.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    logging_level = logging.ERROR
+    if debug:
+        logging_level = logging.DEBUG
+    logging.basicConfig(level=logging_level)
+
+    sys.path.insert(1, path.abspath(base_path or '.'))
+
+    resolver = None
+    if stub:
```
jmcs: As a future improvement we could implement a resolver that mocks a response based on the specification.

jmcs:
clickclick only supports python3, so we can't use it while we still support python 2.7.

rafaelcaricio:
@jmcs That is :unamused:. I will try to make https://github.com/zalando/python-clickclick work with Python 2.7. Shouldn't be difficult.

connexion/cli.py
rafaelcaricio: That is definitely interesting. There are some open source tools that do that, but bringing together the possibility to have partially mocked API's, that can morph in a fully implemented one, is a new approach.

jfinkhaeuser: What I'm looking to do is use my resolving swagger parser [prance](https://pypi.python.org/pypi/prance/) to generate test cases for an API. That's coming from realizing that swagger-tester isn't quite what I'd like it to be.

Some of that code might well be useful for generating the mocked response. One way or another it's going through the spec and generating stuff.

Just a thought. But I can't guarantee I'll be able to spend time on that.

rafaelcaricio:
Clickclick supports Python 2.7 now in zalando/python-clickclick#9

requirements.txt
```diff
@@ -5,3 +5,4 @@ requests>=2.9.1
 six>=1.7
 strict-rfc3339>=0.6
 swagger_spec_validator>=2.0.2
+clickclick>=1.1
```
hjacobs: you can now require `clickclick>=1.2` for Python 2.7 support

setup.py
```diff
@@ -88,5 +88,6 @@ def readme():
         'Topic :: Software Development :: Libraries :: Application Frameworks'
     ],
     include_package_data=True,  # needed to include swagger-ui (see MANIFEST.in)
-
+    entry_points={'console_scripts': ['connexion = connexion.cli:main',
+                                      'cnx = connexion.cli:main']}
```
hjacobs: as already mentioned: I would remove the `cnx` alias --- I think it adds no value here (users can always create as many aliases as they wish via their CLI...)

hjacobs:
@rafaelcaricio we also need to add documentation for this CLI feature :smile:

hjacobs:
The tests still fail for Python 2.7...

rafaelcaricio:
Python behavior changes. I need to see how to normalize this between versions. This PR will be updated as soon as I figure it out.

connexion/cli.py
```diff
@@ -0,0 +1,104 @@
...
+    - SPEC_FILE: specification file that describes the server endpoints.
+
+    - BASE_PATH (optional): filesystem path where the API endpoints handlers are going to be imported from.
+    """
+    logging_level = logging.ERROR
+    if debug:
+        logging_level = logging.DEBUG
+    logging.basicConfig(level=logging_level)
+
+    sys.path.insert(1, path.abspath(base_path or '.'))
```
rafaelcaricio: I think instead of using `'.'` (current this where `connexion` command is run) we should probably use `path.dirname(path.abspath(spec_file))`. Probably that will be the expected behavior by users.

rafaelcaricio:
The problem was caused the missing `__init__.py` files in some modules in the tests tree. 🍭 😞

rafaelcaricio:
@hjacobs Basic docs added. 👍

docs/cli.rst
```diff
@@ -0,0 +1,50 @@
+Commnad-Line Interface
```
hjacobs: typo

rafaelcaricio: Fixed.

connexion/cli.py
```diff
@@ -0,0 +1,112 @@
...
+    py_module_path = base_module_path or path.dirname(spec_file_full_path)
+    sys.path.insert(1, path.abspath(py_module_path))
+    logger.debug('Added {} to system path.'.format(py_module_path))
+
+    resolver_error = None
+    if stub:
+        resolver_error = 501
+
+    app = App(__name__,
+              swagger_json=hide_spec is False,
```
jmcs: `not hide_spec`

connexion/cli.py
```diff
@@ -0,0 +1,112 @@
...
+              swagger_ui=hide_console_ui is False,
```
jmcs: not hide_console_ui

jmcs:
:+1:

hjacobs:
@rafaelcaricio how do you run the CLI from the workdir? Can we add a `__main__.py` file or at least add `if __name__ == '__main__'` into cli.py?

rafaelcaricio:
@hjacobs Added both.

rafaelcaricio:
Coverage down... :disappointed:

hjacobs:
We should configure INFO logging by default to at least see the HTTP port...

connexion/cli.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Runs a server compliant with a OpenAPI/Swagger 2.0 Specification file.
+
+    Arguments:
+
+    - SPEC_FILE: specification file that describes the server endpoints.
+
+    - BASE_MODULE_PATH (optional): filesystem path where the API endpoints handlers are going to be imported from.
+    """
+    logging_level = logging.ERROR
```
hjacobs: I would use `INFO` by default.

Or alternatively: ues `WARN` by default and make it more verbose with `-v` (INFO) and `-vv` (DEBUG).

hjacobs:
Please use `INFO` log level by default, having no output at all is really confusing right now... (first I thought it hangs)

```
$ python3 -m connexion run examples/helloworld/swagger/helloworld-api.yaml  --stub

File: connexion/cli.py (hunk 8 above)
Comment (rafaelcaricio) [inline connexion/cli.py]: I like the idea of changing levels by adding -v or -vv. Will do that.

**Comment by rafaelcaricio** [comment]
@hjacobs I've added support for different log levels. Please check.

File: connexion/cli.py [hunk 9]
```diff
@@ -0,0 +1,127 @@
...
+              is_flag=True, default=False)
+@click.option('--validate-responses',
+              help='Enable validation of response values from operation handlers.',
+              is_flag=True, default=False)
+@click.option('--strict-validation',
+              help='Enable strict validation of request payloads.',
+              is_flag=True, default=False)
+@click.option('--debug', '-d', help='Show debugging information.',
+              is_flag=True, default=False)
+@click.option('--verbose', '-v', help='Show logging information.',
```
hjacobs: nicer way of doing that: http://click.pocoo.org/5/options/#counting

rafaelcaricio:
@hjacobs done.

hjacobs:
:+1:

hjacobs:
Released in https://github.com/zalando/connexion/releases/tag/1.0.121

PR #285, opened 2016-09-12
Fixed response validation when a flask.response object is being returned by a service endpoint.

dpopov76:
I found that if connexion is set to validate **responses** and a service endpoint (or controller method) returns a flask.response object there is a TypeError exception being thrown from decorators/response.validate_response().

The reason is the current code tries to perform json.dumps() on a flask.response object which fails. It looks like only the case where an endpoint returns separate response body, http status and headers is handled in validation. 

My proposed fix is to check whether the response is a flask.response object and if is to perform json.dumps on its data property (via get_data() method, since flask documentation advises against using the flask.response.data directly).

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -21,6 +21,7 @@ from jsonschema import ValidationError
 from ..exceptions import (NonConformingResponseBody,
                           NonConformingResponseHeaders)
 from ..problem import problem
+from ..utils import is_flask_response
 from ..utils import produces_json
 from .decorator import BaseDecorator
 from .validation import ResponseBodyValidator
@@ -57,8 +58,11 @@ class ResponseValidator(BaseDecorator):
             try:
                 # For cases of custom encoders, we need to encode and decode to
                 # transform to the actual types that are going to be returned.
-                data = json.dumps(data)
-                data = json.loads(data)
+                if not is_flask_response(data):
+                    data = json.dumps(data)
+                    data = json.loads(data)
+                else:
+                    data = json.loads(data.get_data())
 
                 v.validate_schema(data)
             except ValidationError as e:
--- a/tests/api/test_responses.py
+++ b/tests/api/test_responses.py
@@ -152,3 +152,14 @@ def test_maybe_blob_or_json(simple_app):
     text, number = unpack('!4sh', resp.data)
     assert text == b'cool'
     assert number == 8
+
+def test_flask_response(simple_app):
+    app_client = simple_app.app.test_client()
+
+    resp = app_client.get('/v1.0/flask-response')
+    assert resp.status_code == 200
+    assert resp.content_type == 'application/json'
+
+    response = json.loads(resp.data.decode())
+    assert response['message'] == 'Hello'
+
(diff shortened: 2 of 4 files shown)


hjacobs:
Thanks for the PR --- I'm not sure which use case it tries to fix.
The current tests fail...

Could you also add a new test case for the issue you are trying to fix?

connexion/decorators/response.py
```diff
@@ -57,8 +57,11 @@ def validate_response(self, data, status_code, headers):
             try:
                 # For cases of custom encoders, we need to encode and decode to
                 # transform to the actual types that are going to be returned.
-                data = json.dumps(data)
-                data = json.loads(data)
+                if not is_flask_response(data):
```
rafaelcaricio: You have to import the function `is_flask_response` before using it.

dpopov76:
I appologize about the missed import, I am still new to Python and GitHub. I will amend my commit with a new test case that is supposed (at least to me) to pass, but fails without the fix.

dpopov76:
I fixed the missing import, added a new test case and made sure all of the 117 tests pass.

Here is more about the fix:
1. The new method `get_flask_response()` in `tests/fakeapi/hello.py` simply returns a `flask.response` object.
2. Without the fix, when `validate_response()` (in `connexion/decorators/response.py`) is reached, the following line causes an exception:
`data = json.dumps(data)`
because `data `is a `flask.response` object and you cannot apply `json.dumps()` on it.

rafaelcaricio:
Please rebase your branch.

rafaelcaricio:
I was reviewing your PR and looking the current Connexion code. I think that the problem might be that we are not getting only the data where [we should](https://github.com/zalando/connexion/blob/master/connexion/decorators/decorator.py#L39).

rafaelcaricio:
And we do not get the data where we should because we reuse the `data` position in the return list of [`get_full_response`](https://github.com/zalando/connexion/blob/master/connexion/decorators/decorator.py#L39) to return [the immutable `flask.Response` object](https://github.com/zalando/connexion/blob/master/connexion/decorators/produces.py#L119-L121) the user created.

Because the order of evaluation of the decorators respective responses in case of user returning a `flask.Response` object is:
- security decorator: returns whatever use sends back
- parameters validation: returns whatever user sends back
- validations decorators (whatever is set): returns the `flask.Response` that user sent back
- produces decorator: (calls get_full_response method) and returns `flask.Response`, status from response object, headers from response object.
- response validation decorator: (calls get_full_response method) and returns `flask.Response`, status from response object, headers from response object. (BOOM! :bomb: Bug happens here!)

rafaelcaricio:
A better way of fixing this issue is **not** reusing this first position of the returned tuple which is a workaround. Better would be creating a new object called `OperationResponse` and handling the access to the `flask.Response` in it.

dpopov76:
Yes, I also initially thought the root cause was in decorator.py but fixing it there caused new problems down the pipeline - this 'data' thing is reused way too much.

I agree my fix is not perfect, but my knowledge of connexion, flask and python in general is still very limited.

hjacobs:
I'm closing the PR because of inactivity/conflicts --- please simply open a new PR when ready to go forward.

PR #286, opened 2016-09-13
Add CLI option to isort to output diff of proposed changes.

jfinkhaeuser:
It took me a while to dig down to the fact that my global Python3 interpreter has a different version from the tox one, which leads to all sorts of weird default isort behaviour.

With this extra flag, Travis will tell you precisely which imports isort would like to see changed.


rafaelcaricio:
:+1:

PR #289, opened 2016-09-14
Respect what the APIs consume

### LINKED ISSUE #287: Error validating a body parameter
### Description

Here is the spec.

```yaml
swagger: '2.0'
info:
  title: Connexion issue
  version: '0.0.1'
host: localhost:8080

paths:
  /:
    put:
      operationId: app.put_document
      consumes:
        - text/plain
      parameters:
        - name: body
          in: body
          required: true
          schema:
            type: string
      responses:
        200:
          description: OK
```

And here is the app.

```python
import connexion
from connexion import NoContent

def put_document(body):
    return NoContent

if __name__ == '__main__':
    app = connexion.App(__name__, port=8080)
    app.add_api('swagger.yaml')
    app.run()
```

### Expected behaviour

```
$ curl -X PUT --header 'Content-Type: text/plain' -d 'test' 'http://localhost:8080/'
```
returns `200 OK`

### Actual behaviour

Instead it returns `400 BAD REQUEST` with the following body.

```json
{
  "type": "about:blank",
  "title": "Bad Request",
  "status": 400,
  "detail": "None is not of type 'string'"
}
```

### Steps to reproduce

```sh
$ python app.py
$ curl -X PUT --header 'Content-Type: text/plain' -d 'test' 'http://localhost:8080/'
```

### Additional info:

Python 3.5.2
Version: 1.0.112

31z4:
Fixes #287.

 - If `application/json` is not listed in the `consumes` then don't parse and validate the request body.


connexion/api.py
```diff
@@ -125,6 +125,10 @@ def __init__(self, swagger_yaml_path, base_url=None, arguments=None,
         # API calls.
         self.produces = self.specification.get('produces', list())  # type: List[str]
 
+        # A list of MIME types the APIs can consume. This is global to all APIs but can be overridden on specific
+        # API calls.
```
rafaelcaricio: Please lets not duplicate comments.

rafaelcaricio:
@31z4  Thanks for you PR. Could you please write at least one test covering this issue?

connexion/api.py
31z4: Sorry, what exactly do you mean? The comment relates to the `consumes` param yet it's almost similar to the `produces`. Should I rewrite it somehow or just remove?

31z4:
@rafaelcaricio of course. Will do this a bit later as well as fix the emerged failures.

connexion/api.py
rafaelcaricio: Its ok.

connexion/decorators/validation.py
```diff
@@ -114,6 +116,9 @@ def __call__(self, function):
 
         @functools.wraps(function)
         def wrapper(*args, **kwargs):
+            if 'application/json' not in self.consumes:
+                return function(*args, **kwargs)
```
rafaelcaricio: Is this necessary? We should have a test that covers this scenario.

31z4: Yes it is. Without it validation fails on the plain text request. Added a test.

connexion/decorators/parameter.py
```diff
@@ -87,10 +89,13 @@ def parameter_to_arg(parameters, function):
     def wrapper(*args, **kwargs):
         logger.debug('Function Arguments: %s', arguments)
 
-        try:
-            request_body = flask.request.json
-        except exceptions.BadRequest:
-            request_body = None
+        if 'application/json' in consumes:
```
jmcs: This is not the right check. There are lots of "+json" mimetypes that have to be parsed as json

connexion/decorators/validation.py
```diff
@@ -114,6 +116,9 @@ def __call__(self, function):
 
         @functools.wraps(function)
         def wrapper(*args, **kwargs):
+            if 'application/json' not in self.consumes:
```
jmcs: Again the wrong check. And I would put the validation logic inside the if to avoid repeating `function(*args, **kwargs)`

connexion/decorators/parameter.py
jmcs: You should use [`is_json_mimetype`](https://github.com/zalando/connexion/blob/master/connexion/utils.py#L137) for this check.

connexion/decorators/validation.py
31z4: Done.

31z4:
@jmcs Thanks for your feedback!

rafaelcaricio:
👍

PR #298, opened 2016-09-23
Ability to define the host interface to bind to.

ibigpapa:
Fixes # 0.

 - Ability to define the host interface to bind to.

### DIFF
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -272,7 +272,7 @@ class App(object):
             wsgi_container = tornado.wsgi.WSGIContainer(self.app)
             http_server = tornado.httpserver.HTTPServer(wsgi_container, **options)
             http_server.listen(self.port, address=host)
-            logger.info('Listening on port %s.. on interface %s..', self.port, host)
+            logger.info('Listening on %s:%s..', host, self.port)
             tornado.ioloop.IOLoop.instance().start()
         elif self.server == 'gevent':
             try:
@@ -280,7 +280,7 @@ class App(object):
             except:
                 raise Exception('gevent library not installed')
             http_server = gevent.wsgi.WSGIServer((self.host, self.port), self.app, **options)
-            logger.info('Listening on port %s..', self.port)
+            logger.info('Listening on %s:%s..', host, self.port)
             http_server.serve_forever()
         else:
             raise Exception('Server %s not recognized', self.server)
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -13,10 +13,10 @@ logger = logging.getLogger('connexion.app')
 
 
 class App(object):
-    def __init__(self, import_name, host=None, port=None, specification_dir='',
+    def __init__(self, import_name, port=None, specification_dir='',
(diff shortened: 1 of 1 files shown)


hjacobs:
Thanks for the PR. This is a backwards-incompatible signature change (for users calling with non-keyword args), probably we can add the `host` parameter at the end instead?

connexion/app.py
```diff
@@ -246,6 +250,11 @@ def run(self, port=None, server=None, debug=None, **options):  # pragma: no cove
             self.port = port
         elif self.port is None:
             self.port = 5000
+        
+        if host is not None:
```
hjacobs: we could make it shorter: `self.host = host or self.host or '0.0.0.0'`

ibigpapa: yeah good call on backwards compat.

connexion/app.py
```diff
@@ -265,15 +271,15 @@ def run(self, port=None, server=None, debug=None, **options):  # pragma: no cove
                 raise Exception('tornado library not installed')
             wsgi_container = tornado.wsgi.WSGIContainer(self.app)
             http_server = tornado.httpserver.HTTPServer(wsgi_container, **options)
-            http_server.listen(self.port)
-            logger.info('Listening on port %s..', self.port)
+            http_server.listen(self.port, address=host)
+            logger.info('Listening on port %s.. on interface %s..', self.port, host)
```
rafaelcaricio: Maybe better be `logger.info('Listening on %s:%s ..', self.host, self.port)`

overquota:
maybe you can change default host to '::' instead of '0.0.0.0'
Reasons: 
> By default listen host='::' interface instead of host='0.0.0.0'. App would think its in ipv6-only world, but you could access it thru ipv4 cause OS would map your ipv4 addr to ipv6 (see http://stackoverflow.com/a/21689979)

And then i would be able to close my pr as duplicate

connexion/app.py
ibigpapa: Agreed

ibigpapa:
Over quota is that behavior is specific? I've not taken the time to review in depth flask run implementation and the under pinning libraries... The next question would be does this work in gevent and tornado?

overquota:
Windows, meh.. you are right, thank you. This is Flask/werkzeug/python http server issue, there are no IPV6_V6ONLY=0 option for socket, wich may be =1 by default in some OSes (Windows for example: https://msdn.microsoft.com/ru-ru/library/windows/desktop/ms738574(v=vs.85).aspx). Some old versions (e.g. XP) not support this option at all :(
okay...jpg

rafaelcaricio:
After this discovery about Windows not supporting the host `::` by default I think we better keep the `0.0.0.0` as default. Let's leave Connexion to _work by default_.

Now with the option of customising the host, after this PR gets merged, users with advanced knowledge can serve their apps over IPv4 and IPv6 sockets setting the host to `::`.

rafaelcaricio:
👍

PR #302, opened 2016-09-29
#292 add --mock option to return example data

### LINKED ISSUE #292: Support for mocking API based on spec
### Description

Recently we added support for applying a stub to unimplemented operations (https://github.com/zalando/connexion/pull/278). Following that line, I think would be interesting to add support for automatic mocking of a API based on the specification.

This mocking feature would work as the stub, only activated to not implemented operations. This is powerful because allows a broader use case for Connexion.

- Developers would get a better insight on how the API should look like after complete implementation;
- Frontend engineers could mock easily all APIs necessary to develop the client UI;
- Mobile and backend peers would be able to mock dependent services for testing;

### Expected bahaviour

```python
app = connexion.Api(__name__)
app.add_api("my_api.yaml", mock_operations=True)
```

and 

```bash
$ connexion run my_api.yaml --mock
```

hjacobs:
Fixes #292.

* add `--mock` option to CLI
* return response from `examples` property

NOTE: this PR will not return proper example structures if no `examples` property was defined (theoretically we could assemble response structures from given schema --- but this is something for the future).


hjacobs:
@rafaelcaricio @jmcs this PR is just adding minimal mock support, but it already provides value --- so I want to merge it :smile:

jmcs:
:+1:

PR #303, opened 2016-09-29
upgraded swagger-ui from upstream

kashif:


rafaelcaricio:
👎  I think we have to revisit how we handle this swagger ui code. We cannot track which branch/tag/githash it came from in the form it is done now. Right now there is no way to know if this is the original code unless we review all lines.

rafaelcaricio:
👎

hjacobs:
@rafaelcaricio I agree, I had the same doubt here ("How do I know which Swagger UI version is used?").

PR #309, opened 2016-10-04
update Swagger UI from upstream (using ./update.sh 2.2.5)

hjacobs:
Update Swagger UI to latest upstream release (2.2.5).

* add `update.sh` to have a reproducible process (this PR is the result of `./update.sh 2.2.5`)
* add `index.html.patch` to fix URL and disable online validator

This PR is addressing the issues we had with #303.


hjacobs:
:+1:

rafaelcaricio:
Why do we need to have the whole code of swagger ui in this repo?

rafaelcaricio:
Can't we just check it out during the process of building a new version in the CI?

hjacobs:
@rafaelcaricio let's discuss the general approach of how to include Swagger UI later on (maybe package for PyPI?).

PR #311, opened 2016-10-09
Receiving --host or -H option via CLI and writing test cases.

### LINKED ISSUE #304: Add --host command line option
We introduced the "host" parameter in #298, but there is no CLI option for it yet.

victor-torres:
Fixes #304.

 - receive `--host` or `-H` option which defaults to None
 - make use of host parameters introduced in  #298 
 - test cases where users specify a host and cases when they don’t

We’re using `-H` instead of `-h` in order not to be confused with help arguments.
Fabric also uses `-H` as host parameter.

### DIFF
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -42,6 +42,7 @@ def main():
 @click.argument('spec_file')
 @click.argument('base_module_path', required=False)
 @click.option('--port', '-p', default=5000, type=int, help='Port to listen.')
+@click.option('--host', '-H', default=None, type=str, help='Host to listen.')
 @click.option('--wsgi-server', '-w', default='flask',
               type=click.Choice(['flask', 'gevent', 'tornado']),
               callback=validate_wsgi_server_requirements,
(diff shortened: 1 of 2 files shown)


rafaelcaricio:
👍

connexion/cli.py
```diff
@@ -42,6 +42,7 @@ def main():
 @click.argument('spec_file')
 @click.argument('base_module_path', required=False)
 @click.option('--port', '-p', default=5000, type=int, help='Port to listen.')
+@click.option('--host', '-H', default=None, type=str, help='Host to listen.')
```
rafaelcaricio: You do not need to define `default=False` again here, by default is is already `None`. http://click.pocoo.org/5/api/#click.Parameter

For the help I would add "Host interface to bind on." instead.

rafaelcaricio:
@victor-torres Obrigado por trabalhar nesse PR. 😉

rafaelcaricio:
👍

PR #312, opened 2016-10-10
added --host option

### LINKED ISSUE #304: Add --host command line option

iamrajhans:
Fixes #304 .


hjacobs:
Thanks for the PR, but #311 was already implemented and merged.

PR #313, opened 2016-10-11
No json pretty printing

sebastianmika:

 - Do not pretty print returned json payload

Personally I think it is not best practice to return pretty printed json. This only unnecessarily enlarges the payload to be returned and this increase can be significant. Pretty printing should be
handled by the client instead.

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -125,7 +125,7 @@ class Jsonifier(BaseSerializer):
                 logger.debug('Endpoint returned an empty response (204)', extra={'url': url, 'mimetype': self.mimetype})
                 return '', 204, headers
 
-            data = [json.dumps(data, indent=2), '\n']
+            data = json.dumps(data)
             response = flask.current_app.response_class(data, mimetype=self.mimetype)  # type: flask.Response
             response = self.process_headers(response, headers)
 
--- a/connexion/problem.py
+++ b/connexion/problem.py
@@ -41,13 +41,13 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
     :return: Json serialized error response
     :rtype: flask.Response
     """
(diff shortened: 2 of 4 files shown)


sebastianmika:
As a side note: I can of course return a flask Response object with my own non-pretty printed json - but then response validation in connexion does not work, and this is one of the feature I like (amongst others) very much.

hjacobs:
@sebastianmika I personally agree, the response JSON should be machine-readable, humans can use their favorite client (e.g. [HTTPie](https://pypi.python.org/pypi/httpie)) to get pretty output.

hjacobs:
@sebastianmika basically this PR reverts #193, see the discussion there (BTW I still agree with you about not having pretty printed by default).

sebastianmika:
True enough.

Regarding what is best practice or not, I found little relevant stuff. However, at one place it was correctly stated that enabling compression is much more important than whether one pretty prints the json or not. 

Using `flask-compress` and adding to my `create_app`

```python
compress = Compress()
compress.init_app(app.app)
```

that is quickly done and indeed had a much greater impact than not pretty printing. 

The same source also argued as gdb from #193 - API results should be nice to look at for developers without any hassle.

So, I would rather close my own PR and go with compression, unless you make a strong point against that.

As an alternative, as also discussed in #193, pretty or not could be an option to `connexion.App` and/or `add_api`, but I don't know when I would have the time to add that.

hjacobs:
@sebastianmika OK, if you are fine with closing your PR :smile:

josvos:
Hi, is there any newer issue opened about this problem after this one?  As a new (and pretty happy!) Connexion user found it pretty astonishing that I saw pretty printed JSON responses and that this doesn't seem to be configurable.  The compress workaround has drawbacks (when clients do not support it for example) and I don't want to depend on that.  In my case, stress-testing the REST API appeared to be network-bound and the data I get is about 4 times the data a compact JSON REST API sends back.

PR #322, opened 2016-10-20
Ensure that 204 and NoContent responses have the correct Content-Type

### LINKED ISSUE #321: 204 responses have wrong Content-Type
### Description
HTTP 204 responses have the wrong `Content-Type`.

### Expected bahaviour
`Content-Type` is `application/json`

### Actual behaviour

`Content-Type` is `text/html; charset=utf-8`

### Steps to reproduce
* Create a DELETE endpoint
* Send a request to it and check the response headers

### Additional info:

Output of the commands:

- `python --version`

```
Version: 1.0.129
$ python --version
Python 2.7.9
```

jakul:
Fixes #321.

 - Uses Flask's `make_response` to construct the response when returning 204 status codes or "no content" responses, which pulls in the correct `Content-Type` header

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -106,13 +106,15 @@ class Jsonifier(BaseSerializer):
             if is_flask_response(data):
                 logger.debug('Endpoint returned a Flask Response', extra={'url': url, 'mimetype': data.mimetype})
                 return data
-            elif data is NoContent:
-                return '', status_code, headers
+
+            if data is NoContent:
+                data = ''
             elif status_code == 204:
                 logger.debug('Endpoint returned an empty response (204)', extra={'url': url, 'mimetype': self.mimetype})
-                return '', 204, headers
+                data = ''
+            else:
+                data = [json.dumps(data, indent=2), '\n']
 
-            data = [json.dumps(data, indent=2), '\n']
             response = flask.current_app.response_class(data, mimetype=self.mimetype)  # type: flask.Response
             response = self.process_headers(response, headers)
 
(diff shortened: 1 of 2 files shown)


hjacobs:
Hmm, "NoContent" (204) should really have no response content, so not even "JSON". What's the motivation for the change? I would suggest to have no "Content-Type" at all as we have no message body (https://tools.ietf.org/html/rfc2616#section-10.2.5).

hjacobs:
@jakul I did not want to chase you away :smirk: I simply think "text/html" is as wrong as "application/json" for "no content" responses...

jakul:
Don't worry @hjacobs, my feelings aren't hurt. I wrote a reply in my email client but it didn't arrive here for some reason.


=====

My motivation was that my automated test runner Dredd (https://github.com/apiaryio/dredd) is failing the DELETE tests in my API because it verfies the Content-Type headers on all responses, and my api is setup as returning "application/json".

I could work around my problem in at least 2 other ways:

1) explicitly return the Content-Type header from each of my views
2) raise a bug report / PR with Dredd to tell them not to validate that header when the response is 204

jakul:
My motivation was that my automated test runner Dredd (
https://github.com/apiaryio/dredd) is failing the DELETE tests in my API
because it verfies the Content-Type headers on all responses, and my api is
setup as returning "application/json".

I could work around my problem in at least 2 other ways:

1) explicitly return the Content-Type header from each of my views
2) raise a bug report / PR with Dredd to tell them not to validate that
header when the response is 204


====

You just prompted me to investigate Dredd a bit more and it seems like they
have an ongoing discussion around 204 parsing at
https://github.com/apiaryio/dredd/issues/556. It also seems like there's no
widely-accepted agreement on Content-Header for 204 responses.

I'll close this PR as I think it's better not to make changes if there's no
justification for them, and will workaround Dredd's issue some other way.


Em qui, 20 de out de 2016 às 17:55, Henning Jacobs <notifications@github.com>
escreveu:

PR #314, opened 2016-10-13
Support app.add_api() with API-dictionary as an alternative to file path

DDEFISHER:
Fixes #296 

Changes proposed in this pull request:
- Add support for an API-dictionary argument to app.add_api() as an alternative to a file path.
- test case creating app and using add_api() with an API-dictionary
- Currently the param is still called swagger_file, this and other params should probably change I am just not sure what to change it to.

PR #325, opened 2016-10-25
Allow jinja includes in swagger.yaml

jakul:
This PR implements the Jinja template loading mechanism specified in #254, which will let us split up our swagger.yaml into smaller files and include them at runtime.

- Allow swagger.yaml to specify Jinja `{% include %}` blocks

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -72,8 +72,8 @@ https://gist.github.com/rafaelcaricio/6e67286a522f747405a7299e6843cd93
 Splitting up your swagger.yaml
 ------------------------------
 
-Connexion does not yet support the `$ref` syntax from Swagger which allows you
-to split up one big zswagger.yaml` into multiple smaller ones. There is a
+Connexion [does not support](https://github.com/zalando/connexion/issues/254) the `$ref` syntax from Swagger which allows you
+to split up one big swagger.yaml` into multiple smaller ones. There is a
 workaround which can be done using Jinja templating though.
 
 
@@ -94,10 +94,10 @@ workaround which can be done using Jinja templating though.
         type: string
         format: money
 
+
 This appraoch won't work if you are including block which should be indented.
 In those cases you have to include the file using a macro, which will let you
 indent it in the main Swagger file.
-
 
 
 .. code-block:: yaml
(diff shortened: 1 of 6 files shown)


hjacobs:
I'm not sure whether we want to support full-blown Jinja macros, one issue might be interoperability (using the same Swagger YAML in IDE/editors).

jakul:
Using the Jinja includes will definitely break editors etc. My preferred solution would be to use the `$ref`s as defined by the Swagger spec, but Connexion doesn't support that yet (and I'm pretty certain it's non-trivial to add).

This PR is intended a stop-gap solution until Connexion has full support for the spec.

hjacobs:
@jakul thanks for the PR first of all :smirk:

rafaelcaricio:
@jakul thanks for the PR.

But I am 👎  for merging this PR. My concerns:

- If we add the support for full-blown Jinja macros and people start relying on it then will be difficult to remove this support later;
- All editors will break sounds not good to me. We are going to add support to a non-standard way of writing OpenAPI specifications;
- The support for full-blown Junja will not need to be done inside Connexion code after #314 gets merged. You will be able to render the specification and then just pass it to `add_api` in Connexion.

docs/cookbook.rst
```diff
@@ -67,3 +67,56 @@ Connexion application. Keep in mind that the format checkers should be
...
+    # properties.yaml
+    properties:
+      title:
+        type: string
+      price_label:
+        type: string
+        format: money
+
+
+This appraoch won't work if you are including block which should be indented.
```
Linkid: Here is a typo: "appraoch" → "approach" ;).

docs/cookbook.rst
```diff
@@ -67,3 +67,56 @@ Connexion application. Keep in mind that the format checkers should be
 defined and registered before you run your application server. A full
 example can be found at
 https://gist.github.com/rafaelcaricio/6e67286a522f747405a7299e6843cd93
+
+
+Splitting up your swagger.yaml
+------------------------------
+
+Connexion [does not support](https://github.com/zalando/connexion/issues/254) the `$ref` syntax from Swagger which allows you
+to split up one big swagger.yaml` into multiple smaller ones. There is a
```
Linkid: Missing a "`" ?

jakul:
Thanks for the feedback. I'm happy to use the approach in #314.

PR #327, opened 2016-10-26
Support for datetime.timedelta responses

### LINKED ISSUE #326: datetime.timeinterval support
### Description
Requested datetime.timeinterval objects returned by the DB are not supported and throw an error.

### Expected bahaviour
Return JSON from timedelta objects

### Actual behaviour
Server will throw a 500 with "TypeError: datetime.timedelta() is not JSON serializable"

### Steps to reproduce
Commit a timedelta object to a db, request it.

### Additional info:

Python 3.5.2
Connexion 1.0.129

jimmellor:
Fixes #326  .

Fixes an issue where timedelta objects returned by the DB throw an error "TypeError: datetime.timedelta() is not JSON serializable" by converting to a time in isoformat

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -29,6 +29,9 @@ class JSONEncoder(json.JSONEncoder):
 
         if isinstance(o, datetime.date):
             return o.isoformat()
+        
+        if isinstance(o, datetime.timedelta):
+            return (datetime.datetime.min + o).time().isoformat()
 
         return json.JSONEncoder.default(self, o)
 


hjacobs:
This only works for deltas < 24h:

```
>>> o = datetime.timedelta(days=500)
>>> (datetime.datetime.min + o).time().isoformat()
'00:00:00'
```

Not sure what the canonical JSON representation would be.

rafaelcaricio:
@jimmellor  Thanks for the PR. Would be nice some test for this. Including the test case proposed by @hjacobs.

connexion/decorators/produces.py
```diff
@@ -29,6 +29,9 @@ def default(self, o):
 
         if isinstance(o, datetime.date):
             return o.isoformat()
+
```
rafaelcaricio: White space here. Please remove it. That's why `flake8` is complaining in the TravisCI job.

jimmellor:
@hjacobs good catch! My immediate use case didn't need this, but much better to support >24h. A logical choice would be an implementation of ISO 8601 duration format, e.g. 'P1Y2M3DT4H5M6S'.

rafaelcaricio:
@jimmellor Hi! Just checking if there is any plans of moving this forward?

rafaelcaricio:
Is there interest in moving this issue forward?

PR #333, opened 2016-10-31
add --proxy option to serve data through a local proxy

dtkav:
Hi - I'm not sure if this is useful to you, but I've added a proxy option to my fork.
It has been useful bypassing CORS, as well as for building an API proxy/gateway.

 - add ProxyResolver which is initialized with a remote server URL
 - add --proxy command line argument to pass in the base server url from the command line

Let me know what you think!

Best.

Daniel

### DIFF
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -3,9 +3,9 @@ import sys
 from os import path
 
 import click
-import connexion
 from clickclick import AliasedGroup, fatal_error
 
+import connexion
 from connexion.mock import MockResolver
 from connexion.proxy import ProxyResolver
 
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -3,9 +3,9 @@ import sys
 from os import path
 
 import click
+import connexion
 from clickclick import AliasedGroup, fatal_error
 
-import connexion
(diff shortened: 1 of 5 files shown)


hjacobs:
@dtkav thanks for using Connexion :smile:. To your PR: as far as I can see it always does a `GET` on the upstream server, so it's not really a full proxy right now.

Another question is whether this broadly useful to other people.

dtkav:
Thanks for your comments @hjacobs. Connexion is a great tool and has been really fun to work with.
I wasn't sure of the best place to start a discussion. I'm happy to just maintain the feature for my own needs if you don't think upstreaming it adds value.

It is a good point that "proxy" is a bit misleading. I'll add the other methods and fix up/add unit tests.

Thanks!

rafaelcaricio:
I think this feature might be useful for other people. But it does not need to be included in Connexion code. It can be an additional package that adds this functionality to Connexion.

If we are willing to move this PR forward we need to take care of many details that it is still missing. Missing parts:
- the hardcoded `application/json`
- missing tests

etc

rafaelcaricio:
Since no interest have been shown, I am closing this issue.

PR #334, opened 2016-10-31
Parameter names are not sanitized

themiwi:
### Description

Parameters that are not valid Python names are not handled gracefully at the moment. E.g. OData query parameters (`$skip`, `$top`, `$filter`) are only passed to the controller if it uses `**kwargs`.

### Expected bahaviour

Parameter names should be sanitized by stripping invalid characters.

### Actual behaviour

If the controller does not have `**kwargs` there is no way for the parameters to be passed as arguments.

### Steps to reproduce

Create a `swagger.yaml` that declares a parameter with a name that is not a valid argument name in Python, such as `$top`.

### Additional info:

Output of the commands:

- `python --version`

    Python 2.7.12

    Version: 2016.0.dev1

- `git describe --always`

    1.0.88-234-g10a0c15

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -2,6 +2,7 @@ import copy
 import functools
 import inspect
 import logging
+import re
 
 import flask
 import six
@@ -58,6 +59,10 @@ def get_val_from_param(value, query_param):
         return make_type(value, query_param["type"])
 
 
+def sanitize_param(name):
+    return name and re.sub('^[^a-zA-Z_]+', '', re.sub('[^0-9a-zA-Z_]', '', name))
+
+
 def parameter_to_arg(parameters, consumes, function):
     """
     Pass query and body parameters as keyword arguments to handler function.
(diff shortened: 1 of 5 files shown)


hjacobs:
Thanks for the PR. Can you add tests?

themiwi:
@hjacobs Done, updated the PR.

themiwi:
Scratch that, I only test for query parameters, will update. Give me 5 mins.

themiwi:
Sanitization now handles query and form parameters and the body name. Path parameters are validated by werkzeug (invalid names are rejected instead of sanitized). Headers are also already mangled, so also no sanitization there.

themiwi:
And finally it also passes the tests on python 3...

rafaelcaricio:
@themiwi Thank you for the PR, seems very useful. I just think that would be nice to mention this in our documentation. This is something that might lead to people spending some time with code trying to get what is going on. Even thought it sounds obvious sanitation, it might not be so obvious to everyone.

connexion/decorators/parameter.py
```diff
@@ -58,6 +59,10 @@ def get_val_from_param(value, query_param):
         return make_type(value, query_param["type"])
 
 
+def sanitize_param(p):
```
rafaelcaricio: I would like to avoid using `p` as variable name. Could you please rename this?

tests/fakeapi/hello.py
```diff
@@ -354,3 +354,12 @@ def more_than_one_scope_defined():
 
 def test_args_kwargs(*args, **kwargs):
     return kwargs
+
+def test_param_sanitization(query = None, form = None):
```
rafaelcaricio: Could you please remove the spaces around the parameter default values? Let's try follow one code style consistently.

tests/fakeapi/hello.py
```diff
@@ -354,3 +354,12 @@ def more_than_one_scope_defined():
...
+    result = {}
+    if query: result['query'] = query
```
rafaelcaricio: Please break lines here. Again, let's try to keep one coding style.

rafaelcaricio:
👍

PR #343, opened 2016-11-07
Correctly pass host parameter from App construction in run() method

### LINKED ISSUE #342: `host` parameter from `App` construction not used correctly in `run()`
I'm just reporting this here so anyone running into this problem can find this issue. Pull request fixing this will follow shortly.

### Description

At the moment passing the `host` parameter to the `App()` constructor and then calling `App.run()` without the `host` parameter leads to `host=None` being passed to the server. For example when calling `app = App(host='0.0.0.0')` and then `app.run()` the default flask host (`127.0.0.1`) is used instead. 

### Expected behaviour

The `host` parameter from `App()` should be used in the `App.run()` method if no new `host` parameter is passed.

### Actual behaviour

The `App.host` parameter is ignored in `App.run()`. 

### Steps to reproduce

~~~
import connexion
app = connexion.App('app.yaml', host='0.0.0.0')
app.run()
~~~

### Additional info:

Output of the commands:

- `python --version` Python 3.4.4 :: Continuum Analytics, Inc.
- `pip show connexion | grep "^Version\:"` Version: 1.0.129

f11r:
Fixes #342 .

### DIFF
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -268,7 +268,7 @@ class App(object):
 
         logger.debug('Starting %s HTTP server..', self.server, extra=vars(self))
         if self.server == 'flask':
-            self.app.run(host, port=self.port, debug=self.debug, **options)
+            self.app.run(self.host, port=self.port, debug=self.debug, **options)
         elif self.server == 'tornado':
             try:
                 import tornado.wsgi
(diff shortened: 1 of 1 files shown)


f11r:
If this needs a test it would be appreciated if someone more familiar with this library than me could point me in the right direction (or just implement it themselves ;)).

rafaelcaricio:
👍

PR #349, opened 2016-11-17
Handle unicode strings in body properly on Python2

### LINKED ISSUE #348: Connexion should use unicode strings on python 2
### Description
On Python 2 Connexion passes parameter data which may be UTF-8 encoded through the `str()` constructor, which is very unlikely to be the desired behavior from the user's perspective.

### Expected bahaviour
Connexion should use python2 unicode strings when appropriate.

### Actual behaviour
Connexion casts all incoming strings to python2 native strings.

### Steps to reproduce
Define an endpoint that uses a parameter of type `string` in your swaggerfile and send a request with unicode characters (e.g. "café") to connexion. The controller function will receive a python2 `str` type:

```python
def controller(some_param="café"):
    print some_param
    # Output: 'caf\xc3\xa9'
    print type(some_param)
    # Output: str
```

### Additional info:

Output of the commands:

- `python --version`:     
Python 2.7.12
 2016.0.dev1

nikvdp:
Fixes #348 .

 - Use `make_type` helper function for body raw body parameters
 - Add a `str_helper` helper function to use unicode strings in python2

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -7,15 +7,30 @@ import re
 import flask
 import six
 import werkzeug.exceptions as exceptions
+import codecs
+import platform
 
 from ..utils import all_json, boolean, is_null, is_nullable
 
 logger = logging.getLogger(__name__)
 
+def str_helper(in_str):
+    """
+    Work around python 2's (lack of) unicode handling.
+    Decode string from utf-8 into python 2 native string on python 2, return 
+    string unchanged on python 3.
+
+    :type function: Callable
+    :rtype: str
+    """
+    if platform.python_version_tuple()[0] == '2':
+        return codecs.decode(in_str, 'utf-8')
+    return in_str
+
 # https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
 TYPE_MAP = {'integer': int,
             'number': float,
-            'string': str,
+            'string': str_helper, 
             'boolean': boolean,
             'array': list,
             'object': dict}  # map of swagger types to python types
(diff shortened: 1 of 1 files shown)


hjacobs:
People still do Python 2? :smirk:

rafaelcaricio:
Build is not passing. Could you please take a look on what is going on @nikvdp ?

connexion/decorators/parameter.py
```diff
@@ -7,15 +7,39 @@
...
+def str_helper(in_str):
+    """
+    Work around python 2's (lack of) unicode handling.
+    Decode string from utf-8 into python 2 native string on python 2, return 
+    string unchanged on python 3.
+
+    :param in_str: Input string to try and decode
+    :rtype: str
+    """
+    if platform.python_version_tuple()[0] == '2':
```
rafaelcaricio: Please use `tox` for this.

connexion/decorators/parameter.py
```diff
@@ -7,15 +7,39 @@
...
+        out = None
+        try:
+            out = codecs.decode(in_str, 'utf-8')
+        except:
```
rafaelcaricio: Please use `ValueError` or some exception that you know can happen here.

connexion/decorators/parameter.py
```diff
@@ -7,15 +7,39 @@
...
+            logger.warn("Unable to decode input string, returning as is...")
```
rafaelcaricio: I think we don't need this log warn here.

connexion/decorators/parameter.py
```diff
@@ -7,15 +7,39 @@
...
+    import codecs
+
+
 logger = logging.getLogger(__name__)
 
+
+def str_helper(in_str):
+    """
+    Work around python 2's (lack of) unicode handling.
+    Decode string from utf-8 into python 2 native string on python 2, return
```
rafaelcaricio: Please remove this trailing whitespace.

connexion/decorators/parameter.py
```diff
@@ -7,15 +7,39 @@
...
+            logger.warn("Unable to decode input string, returning as is...")
+            out = in_str
+        return out
+    return in_str
+
 # https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
 TYPE_MAP = {'integer': int,
             'number': float,
-            'string': str,
+            'string': str_helper,
```
rafaelcaricio: Please remove this trailing whitespace.

PR #363, opened 2016-12-09
Allow custom rendering of oAuth issues

wilsonge:

 - Raises an exception in the oAuth security provider instead of returning a flask response

This allows easier overriding of the oAuth security failures because you add your own error handler to customise the response (if you choose not to use the 'application/json+problem' return type):

```
def render_unauthorized(exception):
    return Response(response=json.dumps({'error_description': 'Test String', 'error': 'test_error'}), status=401, mimetype="application/json")

app = connexion.App(__name__, specification_dir='./../swagger/', debug=debug, swagger_ui=False)
app = app.add_error_handler(401, render_unauthorized)
```

The current messages will continue to be returned in a b/c way because the current default exception handlers find the exceptions now being thrown and format them in the same way as of present.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -6,8 +6,7 @@ import textwrap
 
 import requests
 from flask import request
-
-from ..problem import problem
+from werkzeug.exceptions import Forbidden, Unauthorized
 
 logger = logging.getLogger('connexion.api.security')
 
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for the PR --- we'll have to check it...

connexion/decorators/security.py
```diff
@@ -57,27 +56,27 @@ def wrapper(*args, **kwargs):
...
+                raise Unauthorized('Provided oauth token is not valid')
             token_info = token_request.json()  # type: dict
             user_scopes = set(token_info['scope'])
             logger.debug("... Scopes required: %s", allowed_scopes)
             logger.debug("... User scopes: %s", user_scopes)
             if not allowed_scopes <= user_scopes:
                 logger.info(textwrap.dedent("""
                             ... User scopes (%s) do not match the scopes necessary to call endpoint (%s).
-                             Aborting with 401.""").replace('\n', ''),
+                             Aborting with 403.""").replace('\n', ''),
```
hjacobs: Nice catch! Thanks :smile:

wilsonge:
Awesome! Let me know if you have any questions/issues :)

rafaelcaricio:
@wilsonge Thank you for the PR! Sounds good, but the way it is now will break all existing applications using Connexion. We need to have a way to support the current behaviour (returning 401/403 problem) if I do not configure my own error handler.

wilsonge:
You do keep full b/c on the responses because you guys have error handlers via https://github.com/zalando/connexion/blob/master/connexion/app.py#L66

So as far as I know this doesn't break all apps

rafaelcaricio:
@wilsonge You are right, sorry, sleepy morning. Indeed it shouldn't break anything.

rafaelcaricio:
👍

rafaelcaricio:
@hjacobs BTW, on a slightly related note I strongly believe we should have something like `connexion.Problem` as an exception.

rafaelcaricio:
Opened a new issue for what I just mentined #366

wilsonge:
Thank you 🙇

PR #365, opened 2016-12-10
Refactor code to fix confusion with returned values

### LINKED ISSUE #356: Incorrect set http response data when returning flask.Response objects
### Description

Incorrect set http response data when returning flask.Response objects

### Expected bahaviour

From the [documentation](https://connexion.readthedocs.io/en/latest/response.html#response-serialization):

> "If the endpoint returns a Response object this response will be used as is."

### Actual behaviour

The response is parsed in connexion.decorators.decorator.BaseDecorator.get_full_response() and the [response object itself is mistakenly set as the response's data instead of response.data](https://github.com/zalando/connexion/blob/master/connexion/decorators/decorator.py#L27):

This later throws a JSON serialize error when trying to [JSON serialize the data variable](https://github.com/zalando/connexion/blob/master/connexion/decorators/response.py#L47) (mistaken Response object)

### Additional info:

Output of the commands:

- `python --version`
Python 2.7.12
Version: 1.0.129

rafaelcaricio:
Fixes #356,  related to #285

I finally found time to make this change of how Connexion code internally works. I was very bothered by the `get_full_response` method returning multiple values and we were getting lost a lot to sort out those returned values, reusing positions, etc. This caused the #356 and generated #285. It is an old bug that is fixed by this PR. Also this PR makes it easier to maintain the Connexion codebase as we move forward.

 - We rely on `ResponseContainer` as a wrapper to whatever the operation handler function returns.
 - The `ResponseContainer` object is responsible to manage the data that the operation handler function provided.
 - The `ResponseContainer` object is what all internal Connexion decorators interact with and not with the pure values returned by the operation handler function.
- In the end of the Connexion's internal lifecycle of the request, we return a `flask.Response` instance with the data after all the decorators process it with their respective functionalities (validating response, validating parameters, serialization, etc).

/ping @dpopov76

### DIFF
--- a/connexion/problem.py
+++ b/connexion/problem.py
@@ -1,5 +1,5 @@
 from .decorators.decorator import ResponseContainer
-from .decorators.produces import Jsonfier
+from .decorators.produces import Jsonifier
 
 
 def problem(status, title, detail, type='about:blank', instance=None, headers=None, ext=None):
@@ -41,7 +41,7 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
     # return a problem payload in JSON format.
     return ResponseContainer(
         mimetype='application/problem+json',
-        data=Jsonfier.dumps(problem_response),
+        data=Jsonifier.dumps(problem_response),
         status_code=status,
         headers=headers
     )
--- a/connexion/problem.py
+++ b/connexion/problem.py
@@ -36,7 +36,7 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
 
     # We serialize here because even if the default content type is
     # not set to JSON, which means that the
-    # `decorators.produces.Jsonfier` will not be added to the request
+    # `decorators.produces.Jsonifier` will not be added to the request
     # life-cycle (so we cannot rely on that serialization), we will
     # return a problem payload in JSON format.
     return ResponseContainer(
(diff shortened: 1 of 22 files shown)


rafaelcaricio:
Fixed a typo, coverage decreased... :suspect:

rafaelcaricio:
@hjacobs @jmcs Could you please review this?

rafaelcaricio:
@jmcs @hjacobs Could you please review this?

connexion/problem.py
```diff
@@ -35,10 +34,14 @@ def problem(status, title, detail, type='about:blank', instance=None, headers=No
...
-    body = [json.dumps(problem_response, indent=2), '\n']
-    response = flask.current_app.response_class(body, mimetype='application/problem+json',
-                                                status=status)  # type: flask.Response
-    if headers:
-        response.headers.extend(headers)
-
-    return response
+    # We serialize here because even if the default content type is
+    # not set to JSON, which means that the
+    # `decorators.produces.Jsonfier` will not be added to the request
```
hjacobs: nitpick: Jsonifier

rafaelcaricio: Fixed!

hjacobs:
:+1:

PR #367, opened 2016-12-11
[WIP] Re-organize configs

### LINKED ISSUE #283: Better handling of configurations
### Description

TL;TR: Move Connexion's configurations to the config attribute of the Flask object.

Currently we have over 10 parameters for the constructor of `connexion.app.App`, `connexion.api.API` classes and `connexion.api.App.add_api`. Most of those parameters are configurations that can be set for a specific API definition or globally in the Connexion server instance. This set of configurations are passed along one-by-one and it was OK for a small set of possible configurations, but now we are constantly increasing the possibilities of how to configure Connexion and adding new options to configure. To permit a more flexible and clean way to move forward whether when adding more features or maintaining the current code base we need to change how we deal with configuration options. ~~Flask already has [a built-in](http://flask.pocoo.org/docs/0.11/config/) configuration handling module and we could make use of that for Connexion.~~

rafaelcaricio:
Fixes #283 

**PLEASE DO NOT MERGE, THIS IS A WORKING IN PROGRESS.**

Proposal to remove dependency on parameters to `connexion.App` and `connexion.App#add_api` methods.

 - Possibility to use configuration files.
 - Read configurations automatically from environment variables (for that the environment variable has to be in the format `CONNEXION_{{config_variable_name}}`).
 - All the load and validations are deferred to when the method `connexion.App#run` is called. So we can do properly configuration resolution.

### DIFF
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -6,6 +6,7 @@ import click
 from clickclick import AliasedGroup, fatal_error
 
 import connexion
+from connexion.config import Config
 from connexion.mock import MockResolver
 
 logger = logging.getLogger('connexion.cli')
@@ -71,7 +72,7 @@ def main():
               help='Enable validation of response values from operation handlers.',
               is_flag=True, default=False)
 @click.option('--strict-validation',
-              help='Enable strict validation of request payloads.',
+              help='Enable strict validation of endpoint parameters.',
               is_flag=True, default=False)
 @click.option('--debug', '-d', help='Show debugging information.',
               is_flag=True, default=False)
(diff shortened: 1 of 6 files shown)


connexion/cli.py
```diff
@@ -71,7 +73,7 @@ def main():
               help='Enable validation of response values from operation handlers.',
               is_flag=True, default=False)
 @click.option('--strict-validation',
-              help='Enable strict validation of request payloads.',
+              help='Enable strict validation of endpoint parameters.',
```
rafaelcaricio: This configuration name is misleading. I think we should think about renaming it. I named in the config variable as `STRICT_PARAMETERS_VALIDATION` to be more specific. But I would mark this to be removed in a future 2.x version of Connexion.

rafaelcaricio:
Closing for now. I will try a different approach for this problem.

PR #368, opened 2016-12-12
Problem exceptions

### LINKED ISSUE #366: Support connexion.Problem as an exception
### Description

Sometimes you are creating API code that needs to just stop and return an error to the client (also rollback db changes in case it happens). A standard way of achieving that is using exceptions. Connexion provides support for standard problem responses for APIs (https://tools.ietf.org/html/rfc7807) which is great. We currently support that by providing the `connexion.problem` function that can be used by our operation handlers to return problems to clients.

Currently if we want to use exceptions to break execution you need to create a internal exception for your API and then using a decorator turn the exception in a problem response to be returned as the operation handler response.

Would be nice if Connexion would offer a `connexion.Problem` exception that would be automatically handled by Connexion itself to return problem responses.

### Expected behaviour

```python
from connexion import Problem

def my_operation_handler(param_1):
    some_internal_work()
    return 'Cool stuff!'

def some_internal_work():
    if not you_are_really_allowed():
        raise Problem(403, 'Oh noes', 'I cannot allow you to perform this action')
    ...
```

### Current behaviour

You can mimic this by doing:

```python
import connexion
from decorator import decorator

class Problem(Exception):
    def __init__(self, status_code, title, description):
        self.status_code = status_code
        self.title = title
        self.description = description

@decorator
def handle_problem_exceptions(func: callable, *args, **kwargs):
    try:
        return func(*args, **kwargs)
    except Problem as prb:
        return connexion.problem(prb.status_code, prb.title, prb.description)

@handle_problem_exceptions
def my_operation_handler(param_1):
    some_internal_work()
    return 'Cool stuff!'

def some_internal_work():
    if not you_are_really_allowed():
        raise Problem(403, 'Oh noes', 'I cannot allow you to perform this action')

### PR DESCRIPTION
**Pull request by wilsonge**
Fixes #366 .

 - Adds in a base problem exception class which allows better customisation of responses
 - Demonstration of use for oAuth 2 Decorator

### DIFF
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -5,7 +5,6 @@ import flask
 import werkzeug.exceptions
 
 from connexion.decorators.produces import JSONEncoder as ConnexionJSONEncoder
-from connexion.exceptions import ProblemException
 from connexion.resolver import Resolver
 
 from .api import Api
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -5,6 +5,7 @@ import flask
 import werkzeug.exceptions
 
 from connexion.decorators.produces import JSONEncoder as ConnexionJSONEncoder
+from connexion.exceptions import ProblemException
 from connexion.resolver import Resolver
 
 from .api import Api
@@ -85,8 +86,13 @@ class App(object):
         """
         :type exception: Exception
         """
+        # If we have an instance of problem exception we can use the title to render a more complete title.
+        if isinstance(exception, ProblemException) and isinstance(exception, werkzeug.exceptions.HTTPException):
+            return problem(title=exception.title, detail=exception.description, status=exception.code)
+
         if not isinstance(exception, werkzeug.exceptions.HTTPException):
             exception = werkzeug.exceptions.InternalServerError()
+
         return problem(title=exception.name, detail=exception.description, status=exception.code)
 
     def add_api(self, specification, base_path=None, arguments=None, auth_all_paths=None, swagger_json=None,
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -87,7 +87,7 @@ class App(object):
         :type exception: Exception
         """
         # If we have an instance of problem exception we can use the title to render a more complete title.
-        if isinstance(exception, ProblemException) and isinstance(exception, werkzeug.exceptions.HTTPException):
+        if isinstance(exception, ProblemException):
             return problem(title=exception.title, detail=exception.description, status=exception.code)
 
         if not isinstance(exception, werkzeug.exceptions.HTTPException):
--- a/connexion/app.py
+++ b/connexion/app.py
@@ -86,14 +86,11 @@ class App(object):
         """
         :type exception: Exception
         """
-        # If we have an instance of problem exception we can use the title to render a more complete title.
-        if isinstance(exception, ProblemException):
-            return problem(title=exception.title, detail=exception.description, status=exception.code)
-
         if not isinstance(exception, werkzeug.exceptions.HTTPException):
             exception = werkzeug.exceptions.InternalServerError()
 
-        return problem(title=exception.name, detail=exception.description, status=exception.code)
+        problem_title = getattr(exception, 'title', exception.name)
+        return problem(title=problem_title, detail=exception.description, status=exception.code)
 
     def add_api(self, specification, base_path=None, arguments=None, auth_all_paths=None, swagger_json=None,
                 swagger_ui=None, swagger_path=None, swagger_url=None, validate_responses=False,
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -1,5 +1,6 @@
 from werkzeug.exceptions import Forbidden, Unauthorized
 
+
 class ConnexionException(Exception):
     pass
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: connexion/app.py [hunk 1]
```diff
@@ -85,8 +86,13 @@ def common_error_handler(exception):
         """
         :type exception: Exception
         """
+        # If we have an instance of problem exception we can use the title to render a more complete title.
+        if isinstance(exception, ProblemException) and isinstance(exception, werkzeug.exceptions.HTTPException):
```
rafaelcaricio: Isn't it enough to check if it is an instance of `ProblemException` here?

connexion/exceptions.py
```diff
@@ -60,3 +75,19 @@ def __init__(self, message, reason="Response body does not conform to specificat
 class NonConformingResponseHeaders(NonConformingResponse):
     def __init__(self, message, reason="Response headers do not conform to specification"):
         super(NonConformingResponseHeaders, self).__init__(reason=reason, message=message)
+
+
+class OAuthProblem(ProblemException, Unauthorized):
+    def __init__(self, title='Unauthorized', **kwargs):
```
rafaelcaricio: Why do we duplicate here the title? It could come from [`self.name`](https://github.com/pallets/werkzeug/blob/master/werkzeug/exceptions.py#L107).

rafaelcaricio:
@wilsonge Thank you for the PR. It is a good start!

I think it would be a good idea to use subclasses of `ProblemException` throughout Connexion code with custom templates following ideas from the RFC https://tools.ietf.org/html/rfc7807 as I mentioned at #368. But it might be food for a new PR. 😉

connexion/exceptions.py
```diff
@@ -1,7 +1,22 @@
+from werkzeug.exceptions import HTTPException, Forbidden, Unauthorized
+
+
 class ConnexionException(Exception):
     pass
 
 
+class ProblemException(ConnexionException, HTTPException):
```
rafaelcaricio: Let's expose this in the `connexion.__init__.py` to make it more convenient to import `from connexion import ProblemException` or `import connexion; connexion.ProblemException`.

connexion/app.py
wilsonge: Kinda. `ProblemException` actually needs to be combined with a direct subclass of `werkzeug.exceptions.HTTPException` (but your right my code doesn't achieve that). As several functions in the `werkzeug.exceptions.HTTPException` will fail without a status code. Hence why my 3 OAuth execption classes combine a `werkzeug.exceptions.HTTPException_SUBCLASS` with a `ProblemException`

connexion/exceptions.py
wilsonge: I wanted to make this something custom (such as OAuth Credentials Error) but I wasn't sure if this would be considered a b/c break?

connexion/exceptions.py
wilsonge: :+1: Done

wilsonge:
I would convert decorator by decorator. It also makes things easier to revert if there's a problem with one of the implementations :) I think if this is accepted I'll move onto the ParamValidation as that's probably the next most interesting one to tackle

wilsonge:
OK I think this is ready for review again :)

connexion/exceptions.py
```diff
@@ -1,7 +1,25 @@
...
+    def __init__(self, title=None, description=None, response=None):
+        """
+        :param title: Title of the problem.
+        :type title: str
+        """
+        ConnexionException.__init__(self)
+        HTTPException.__init__(self, description, response)
+
+        self.title = title
```
rafaelcaricio: You can simply do:
```python
self.title = title or self.name
```

connexion/app.py
```diff
@@ -85,8 +86,13 @@ def common_error_handler(exception):
         """
         :type exception: Exception
         """
+        # If we have an instance of problem exception we can use the title to render a more complete title.
```
rafaelcaricio: Please remove this comment.

connexion/app.py
```diff
@@ -85,8 +86,13 @@ def common_error_handler(exception):
...
+        if isinstance(exception, ProblemException):
+            return problem(title=exception.title, detail=exception.description, status=exception.code)
+
         if not isinstance(exception, werkzeug.exceptions.HTTPException):
             exception = werkzeug.exceptions.InternalServerError()
+
         return problem(title=exception.name, detail=exception.description, status=exception.code)
```
rafaelcaricio: Instead of the previous `if` from line 90. Please do this here:
```python
problem_title = getattr(exception, 'title', exception.name)
return problem(title=problem_title, detail=exception.description, status=exception.code)
```

rafaelcaricio:
@wilsonge Could you please remove this import:
```
connexion/app.py:8:1: F401 'connexion.exceptions.ProblemException' imported but unused
```

wilsonge:
Done :)

rafaelcaricio:
👍

rafaelcaricio:
@wilsonge Thank you for working on this! 😃

wilsonge:
Thank you for all the feedback  🙇

PR #369, opened 2016-12-14
Add exception for when extra parameters are supplied

wilsonge:
This adds custom exception type for when too many parameters are supplied. 

This bit is relatively straightforward so I'm separating it out from the Parameter validation which is probably going to be much more complex

### DIFF
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -1,4 +1,5 @@
-from werkzeug.exceptions import BadRequest, Forbidden, HTTPException, Unauthorized
+from werkzeug.exceptions import BadRequest, Forbidden, HTTPException,
+                                Unauthorized
 
 
 class ConnexionException(Exception):
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -1,4 +1,4 @@
-from werkzeug.exceptions import BadRequest, Forbidden, HTTPException,
+from werkzeug.exceptions import BadRequest, Forbidden, HTTPException,\
                                 Unauthorized
 
 
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -1,5 +1,5 @@
-from werkzeug.exceptions import BadRequest, Forbidden, HTTPException,\
-                                Unauthorized
+from werkzeug.exceptions import (BadRequest, Forbidden, HTTPException,
+                                 Unauthorized)
 
 
 class ConnexionException(Exception):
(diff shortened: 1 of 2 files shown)


wilsonge:
OK What's the code sniffer after here? Not seeing it

connexion/exceptions.py
```diff
@@ -95,3 +96,20 @@ def __init__(self, token_scopes, required_scopes, title=None, **kwargs):
         self.missing_scopes = required_scopes - token_scopes
 
         super(OAuthScopeProblem, self).__init__(title=title, **kwargs)
+
+
+class ExtraParameterProblem(ProblemException, BadRequest):
+    def __init__(self, formdata_parameters, query_parameters, title=None, description=None, **kwargs):
```
rafaelcaricio: We should separate this in two different Exceptions (one for formdata and another for query). The way it is now this exception means "there is a extra parameter in formdata or query or both, I am not sure". This is not good.

rafaelcaricio: So with this idea we would have a handle of this exception that could render something like:

```json
{
   "type": "https://example.net/validation-error",
   "title": "Your request has extra parameters.",
   "extra-query-params": ["name", "age"],
   "extra-formdata-params": ["reason", "color"]
}
```

wilsonge: Correct. Or you could have something like:

```
{
    "type": "https://example.net/validation-error",
    "title": "Your request has extra parameters in the query.",
    "extra_params": ["name", "age"],
}
```
Or like you can keep b/c with the existing code (like i have with the description)

The point is having all this information in the exception allows you the choice :) You're giving the user as much information as possible to allow them to render the exception into the format they decide is most optimal for their use case.

rafaelcaricio: Ok. Sounds good.

rafaelcaricio:
@wilsonge Could you please start some docs about the exceptions in Connexion? If not I will open an issue for that.

rafaelcaricio:
Also please fix the imports ordering:
```patch
--- /home/travis/build/zalando/connexion/connexion/exceptions.py:before	2016-12-14 14:34:03.012488
+++ /home/travis/build/zalando/connexion/connexion/exceptions.py:after	2016-12-14 14:36:53.691374
@@ -1,5 +1,5 @@
-from werkzeug.exceptions import BadRequest, Forbidden, HTTPException,\
-                                Unauthorized
+from werkzeug.exceptions import (BadRequest, Forbidden, HTTPException,
+                                 Unauthorized)
```

rafaelcaricio:
👍

wilsonge:
https://github.com/zalando/connexion/pull/371 Here's a start on the docs

hjacobs:
@wilsonge thanks!

PR #370, opened 2016-12-14
Add tests for the parameter validation method

wilsonge:
Some extra tests for the `connexion.decorators.validation.ParameterValidator.validate_parameter` method - before I start to try working on things


hjacobs:
:+1:

PR #379, opened 2017-01-07
Support the flask url variable converter "path"

patrickw276:
The Flask variable converter "path" allows url variables to include forward slashes in their values. This PR adds support for that feature if the Swagger specification indicates that a route parameter's format is "path". 

Example:

    parameters:
      - name: name
        in: path
        required: true
        format: path
        type: string

will add a route parameter `<path:name>`.

Excerpt from the Swagger specification:

> the format property is an open string-valued property, and can have any value to support documentation needs. Formats such as "email", "uuid", etc., can be used even though they are not defined by this specification.

### DIFF
--- a/docs/routing.rst
+++ b/docs/routing.rst
@@ -90,6 +90,30 @@ under-score is encountered. As an example:
 Without this sanitation it would e.g. be impossible to implement an
 [OData](http://www.odata.org) API.
 
+Parameter Variable Converters
+-----------------------------
+
+Connexion supports Flask's ``int``, ``float``, and ``path`` route parameter
+`variable converters
+<http://flask.pocoo.org/docs/0.12/quickstart/#variable-rules>`_.
+Specify a route parameter's type as ``integer`` or ``number`` or its type as
+``string`` and its format as ``path`` to use these converters. For example:
+
+.. code-block:: yaml
+
+  paths:
+    /greeting/{name}:
+      # ...
+      parameters:
+        - name: name
+          in: path
+          required: true
+          type: string
+          format: path
+
+will create an equivalent Flask route ``/greeting/<path:name>``, allowing
+requests to include forward slashes in the ``name`` url variable.
+
 API Versioning and basePath
 ---------------------------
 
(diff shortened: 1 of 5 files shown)


hjacobs:
@patrickw276 thanks for the PR! :smile: 

I had to check what the OpenAPI spec says about forward slashes. Apparently it says "nothing" (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathsObject), so I think your approach is good and does not validate any specification.

hjacobs:
:+1:

hjacobs:
@patrickw276 LGTM, I'll just wait for @rafaelcaricio and @jmcs to have a look at it.

hjacobs:
@patrickw276 can you add a small section in the ["Routing" docs](https://connexion.readthedocs.io/en/latest/routing.html) about this new feature (https://github.com/zalando/connexion/blob/master/docs/routing.rst)? Thanks!

patrickw276:
Thanks for the quick PR review @hjacobs. :smile:

I updated docs/routing.rst with a short section on parameter variable converters.

docs/routing.rst
```diff
@@ -90,6 +90,28 @@ under-score is encountered. As an example:
 Without this sanitation it would e.g. be impossible to implement an
 [OData](http://www.odata.org) API.
 
+Parameter Variable Converters
+-----------------------------
+
+Connexion supports Flask's ``int``, ``float``, and ``path`` route parameter
```
hjacobs: Can you link to the relevant Flask docs here?

docs/routing.rst
```diff
@@ -90,6 +90,28 @@ under-score is encountered. As an example:
...
+    /greeting/{name}:
+      # ...
+      parameters:
+        - name: name
+          in: path
+          required: true
+          type: string
+          format: path
+
+will create an equivalent Flask route ``/greeting/<path:name>``.
```
hjacobs: I would add one sentence on the implications, i.e. that the greeting name (in this example) can now contain forward slashes.

patrickw276:
Updated to address documentation comments. :smile:

hjacobs:
:+1:

rafaelcaricio:
👍

PR #382, opened 2017-01-10
fix content-type header bug

### LINKED ISSUE #381: Incorrect Mimetype/Content-Type Handling
I think there may be a bug in Content-Type/Mimetype handling.

If an operation has a single `produces` type defined in swagger and the implementation returns a flask Response object, both the swagger `produces` value and the Response object's `content-type` are returned. Example:

```yaml
operationId: app.get_greeting
produces:
  - text/plain
```

with the implementation:

```python
def post_greeting(name):
    response = flask.make_response('Hello {}\n'.format(name))
    response.headers['content-type'] = 'text/plain'
    response.mimetype = 'text/plain'
    return response
```

`curl -v` request to the route returns:

```
< Content-Type: text/plain; charset=utf-8
< Content-Length: 11
< Content-Type: text/plain; charset=utf-8
< Content-Length: 11
```

If you don't specifically set the Response object's content-type or mimetype, then the following is returned. 

```
< Content-Type: text/plain; charset=utf-8
< Content-Length: 11
< Content-Type: text/html; charset=utf-8
< Content-Length: 11
```

This is a bigger issue when an operation has multiple `produces` values. If you use the same code as above with an added `application/json` to the `produces` values, `application/json` is always returned in addition to `text/plain`. 

```
< Content-Type: application/json
< Content-Length: 11
< Content-Type: text/plain; charset=utf-8
< Content-Length: 11
```

I suspect that the code causing this issue is in `connexion/decorator.py::BeginOfRequestLifecycleDecorator.get_response_container`.

patrickw276:
Fixes #381.

### DIFF
--- a/connexion/decorators/decorator.py
+++ b/connexion/decorators/decorator.py
@@ -174,8 +174,8 @@ class ResponseContainer(object):
         :rtype: flask.Response
         """
         self._response = flask.current_app.response_class(
-            self.data, mimetype=self.mimetype)  # type: flask.Response
+            self.data, mimetype=self.mimetype, content_type=self.headers.get('content-type'),
+            headers=self.headers)  # type: flask.Response
         self._response.status_code = self.status_code
-        self._response.headers.extend(self.headers)
 
         return self._response
--- a/tests/decorators/test_decorators.py
+++ b/tests/decorators/test_decorators.py
@@ -1,0 +1,15 @@
+import flask
+from connexion.decorators.decorator import ResponseContainer
+
+
+def test_response_container_content_type():
+    app = flask.Flask(__name__)
+    response = flask.Response(response='test response', content_type='text/plain')
+    container = ResponseContainer(mimetype='application/json', response=response)
+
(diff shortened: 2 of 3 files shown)


rafaelcaricio:
👍

hjacobs:
Can someone clarify the use of `mimetype` vs. `content_type`?

rafaelcaricio:
@hjacobs 

> _The mimetype (content type without charset etc.)_
ref: http://flask.pocoo.org/docs/0.12/api/#flask.Response.mimetype

hjacobs:
@rafaelcaricio yes, but why is it different in the test (`text/plain` vs `application/json`) and what's the "supposed" meaning at that point (container has "default" mimetype and response overwrites it?)?

rafaelcaricio:
@hjacobs in this case it's because we supposed to return a the `flask.Response` without modifications. So whatever is set in the `flask.Response` should be in the final object.

hjacobs:
:+1:

PR #383, opened 2017-01-11
Problem as exceptions should convert properly to problem responses

rafaelcaricio:
The `ProblemException` is an internal behaviour of Connexion and shouldn't mix with the Flask responses. Furthermore, the previous implementation of the `ProblemException` introduced in #368 could not be converted directly to problem responses. This PR fixed this.


jmcs:
:+1:

PR #384, opened 2017-01-12
ProblemExceptions should be caught by our common handler

rafaelcaricio:
- `ProblemException` should be caught by the common handler.


jmcs:
:+1:

Issue #380, opened 2017-01-09
Framework-agnostic Connexion


garbas:
Hi,

First thank you for a great work you did, at Mozilla Release Engineering team we started using it few months ago and it proved to be the step into the right direction.

Currently Connexion is very flask specific which worked ok for initial work we are doing and since majority of our team is familiar with flask. But talking to others at Mozilla they like the idea but would like to use their existing framework. You think it would be worthwhile having Connexion being a cross framework utility, with framework specific helpers/layers.

I was wondering if you would be interesting into such contribution? Of course I would consider doing (and later helping maintain it) the majority of the work, since I'm proposing it.

I guess first step could be making Connexion framework agnostic. Since this would mainly involve re-organization of the code, we might also meet in person (I'm based in Berlin).

rafaelcaricio:
Hi @garbas ,

Thank you for the feedback we really appreciate it.

About making Connexion more framework agnostic, Yes. I was discussing about it some time ago with @jmcs and this is something that we are interested in working on. That is something that I personally would like to be involved with, so I would be willing to join you in making this idea move forward. Connexion cold evolve to be like a library/layer over other frameworks (Tornado/Django/etc).

We can setup some time to sit in person as well, that would be cool! 😃

garbas:
@rafaelcaricio awesome! I will ping you on twitter to meet in person.

I expect what we will come up with we will post here (or in additional PR) for further discussion to not exclude others.

tarekziade:
Hi all, thanks @garbas for starting this.

What I would be interested in is too see if there's a layer between Connexion and Swagger that can be reused as a lib without making any assumptions on how its actually used in a web framework (spec=>code or code=>spec, how it's linked to views etc) (a little bit like bravado-core but specific to the server-side)
 
I started to work on this on my side but I'd love to see if we can share some stuff

hjacobs:
@garbas @rafaelcaricio I would also be happy to join any Berlin meeting :smile:

jonathansp:
It would be awesome! I'm currently using connexion in a personal project and after writing some code I realized that it's very flask-specific while frameworks such as tornado and django have a few libraries for open API initiative. Thanks for starting this thread @garbas. Please let me know if I can help.

hjacobs:
Thoughts:

* `pip install connexion` should only install core features (i.e. not pulling in Flask or any other web framework)
* `connexion` console script will tell user to install at least one supported web framework of their choice (like Flask or Pyramid)
[...]

PR #386, opened 2017-01-12
Frameworks

hjacobs:
First goal to decouple Connexion from Flask: extract calls to Flask into the new `Framework` interface.

Addresses #380 

** WORK IN PROGRESS - DO NOT MERGE **

PR opened for discussion cc @garbas

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -3,17 +3,12 @@ import logging
 import pathlib
 import sys
 
-import flask
 import jinja2
 import six
-import werkzeug.exceptions
-
 import yaml
 from swagger_spec_validator.validator20 import validate_spec
 
-from . import utils
 from .exceptions import ResolverError
-from .handlers import AuthErrorHandler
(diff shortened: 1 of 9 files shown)


dutradda:
@hjacobs are you working on it now?
I am thinking to code to decouple flask from the validators classes.
I am creating a AioHttpFramework class and I stuck on validators. Can I decouple this things and the all things that I will need to create the AioHttpFramework?
I am thinking to create two simple namedtuple for Request/Response to wrap from on frameworks. This is necessary because all the frameworks have differents interfaces, and the connexion needs to have it own request/response interface.
So the Framework class should has two more abstract methods: _request_decorator and _response_decorator. This methods will do the conversion from frameworks request/response types to connexion types.

rafaelcaricio:
@dutradda You can move forward with that.

rafaelcaricio:
For now we are moving forward with #394. Closing this.

Issue #389, opened 2017-01-18
Add support for x-authentication-scheme and Bearer auth scheme in apiKey security definition


advance512:
### Description

As seen in issue #124, many users require support for JWT, specifically using the Bearer authentication scheme. 

Using JWT seems like a very common scenario, as are various non-Basic authentication schemes that use the [`Bearer` authentication scheme](https://tools.ietf.org/html/rfc6750#section-2.1) as basis. 
A hackish solution was offered in that thread, but the resultant experience isn't very comfortable and requires you to edit the Connexion source files (hence it is a problem to be based on the latest version of connexion).

I suggest a vendor-specific setting, `x-authentication-scheme`, in the security definition might suffice:

```
securityDefinitions:
  jwt:
    type: apiKey
    name: Authorization
    in: header
    x-authentication-scheme: Bearer
```

This will cause the API key set in the Authorize dialog to be treated as a JWT token, hence when it is part of a Authorization header the prefix "Bearer " will be appended.

rafaelcaricio:
Related issue #406

PR #390, opened 2017-01-18
Issue #389 - Add support for x-authentication-scheme and Bearer authentication scheme in apiKey security definition

advance512:
PR for issue #389.

    Added support for x-authentication-scheme in ApiKeyAuthorization object
    Added x-authentication-scheme field in Authorize dialog
    Automatically adds 'Bearer ' prefix to Authorization header in case x-authentication-scheme is 'Bearer'
    Beautified swagger-ui.js
    Updated swagger-ui.min.js
    index.html now uses minified version of swagger-ui.js for better performance (86.57% compression, saving 2804.37 kb)


rafaelcaricio:
Hi @advance512! Thank you for working on it. But this PR is impossible to review. Could you please make the necessary changes in the existing files?

advance512:
Good point.

How about you beautify the existing JavaScript in a new commit in the branch master (or develop), and then I'll rebase on that version so you can see the diff?

I used: http://jsbeautifier.org/

It is impossible to do any proper work in the existing JS file.

rafaelcaricio:
That one of the problems of this vendor of the Swagger UI within Connexion. We must find a way to handle this better. We shouldn't be developing on the minified version of the JS. 

We should have a submodule to the original Swagger UI repo under a directory like `/vendor`, plus adding our patch and static files processing (minification, compilation, appending, etc) in the `release.sh` script just before publishing a new Connexion package.

I know it is a lot more work, and I think it is not fair to ask you for that since you want to change something completely different and small. I will try to find time as soon as possible to do some work in this direction and then we can move forward with further modification in those static files. How does it sound?

advance512:
Well, sure. I will try and create another patch, temporarily, that affects just the non-minified field, with the existing JS formatting.

rafaelcaricio:
I have opened an issue for the Swagger UI handling https://github.com/zalando/connexion/issues/406. Lets continue in the new PR #391 to proceed with this change.

Issue #395, opened 2017-02-06
Unicode Strings in Response body


LeenaBhegade:
### Description
I am using Python 2.7 , connexion == 1.0.129

I am using connexion for writing my api endpoints. In the code I retrieve unicode strings ( for example: £ ) from database, create a dictionary and return it to client. Initially I assumed connexion will support utf-8 encoding and I don't have to make any additional changes to the configuration. 
When I did a  GET/ request  to my endpoint (which is further retrieving data from database and sending back response) , below is what I get on client (browser ) side

 {
        "user": 123, 
        "format": "\u00a3", 
        "id": 10, 
  }

The unicode strings are escaped and £ is not getting printed as expected. 

After reading more about encoding I came through Flask configuration variable JSON_AS_ASCII  to not encode strings to ASCII and output strings as-is and return unicode strings.
 I set the JSON_AS_ASCII in the app.app.config['JSON_AS_ASCII']=False.After running the code again I triggered a get request and got another

UnicodeEncodeError: 'ascii' codec can't encode character u'\xa3' in position 991: ordinal not in range(128)

After looking at code I noticed that the failure is caused during serialisation of the json response in decorators.produces class. 
The method Jsonifier.dumps() in produces class fails while returning the json response ---->
 "return {}\n".format(json.dumps(data, indent=2))

I checked the implementation on json.dumps in  flask.json.dumps() and noticed that the json message is not encoded before returning. the method is using keyword argument "encoding" for encoding the json message and currently it is "none" instead of "utf-8" ( No keyword argument is passed to Flask.json.dumps() from connexion currently )

 
### Expected behaviour
 Response on browser
 {
        "user": 123, 
        "format": "£", 
        "id": 10, 
  }

### Actual behaviour
Response on browser
{
        "user": 123, 
        "format": "\u00a3", 
        "id": 10, 
  }

### Additional info:
The connexion project was created from swagger.io

### Questions 
Is there any way to set a set the keyword argument "encoding:utf-8"  to flask.json.dumps() in  connexion or do I need to use python 3  ?

Is there a configuration parameter in connexion which  sets the  keyword argument encoding = "utf-8"  ?

I am very new to this framework so please correct me if I have missed important information while implementation.

rafaelcaricio:
[...]

PR #397, opened 2017-02-07
pass encoding=utf-8 keyword argument to flask.json.dumps

LeenaBhegade:
Issue #395 
Passing keyword argument "encoding=utf-8" to flask.json.dumps to make sure responses are encoded using utf-8.

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -2,9 +2,9 @@
 import datetime
 import functools
 import logging
-import six
 
 import flask
+import six
 from flask import json
 
 from .decorator import BaseDecorator
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -76,7 +76,7 @@ class Jsonifier(BaseSerializer):
         """ Central point where JSON serialization happens inside
         Connexion.
         """
-        return "{}\n".format(json.dumps(data, indent=2))
+        return "{}\n".format(json.dumps(data, indent=2, encoding="utf-8"))
 
     def __call__(self, function):
         """
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -77,7 +77,7 @@ class Jsonifier(BaseSerializer):
         """ Central point where JSON serialization happens inside
         Connexion.
         """
-        if six.PY2 == True:
+        if six.PY2:
             json_content = json.dumps(data, indent=2, encoding="utf-8")
         else:
             json_content = json.dumps(data, indent=2)
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -78,11 +78,11 @@ class Jsonifier(BaseSerializer):
         Connexion.
         """
         if six.PY2 == True:
-            encoding = "utf-8"
+            json_content = json.dumps(data, indent=2, encoding="utf-8")
         else:
-            encoding = None
+            json_content = json.dumps(data, indent=2)
 
-        return "{}\n".format(json.dumps(data, indent=2, encoding=encoding))
+        return "{}\n".format(json_content)
 
     def __call__(self, function):
         """
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -2,6 +2,7 @@
 import datetime
 import functools
 import logging
+import six
 
 import flask
 from flask import json
@@ -76,7 +77,12 @@ class Jsonifier(BaseSerializer):
         """ Central point where JSON serialization happens inside
         Connexion.
         """
-        return "{}\n".format(json.dumps(data, indent=2, encoding="utf-8"))
+        if six.PY2 == True:
+            encoding = "utf-8"
+        else:
+            encoding = None
+
+        return "{}\n".format(json.dumps(data, indent=2, encoding=encoding))
 
     def __call__(self, function):
         """
--- a/tests/api/test_bootstrap.py
+++ b/tests/api/test_bootstrap.py
@@ -133,13 +133,13 @@ def test_single_route(simple_app):
 def test_resolve_method(simple_app):
     app_client = simple_app.app.test_client()
     resp = app_client.get('/v1.0/resolver-test/method')  # type: flask.Response
-    assert resp.data.decode() == '"DummyClass"\n'
+    assert resp.data.decode('utf-8') == '"DummyClass"\n'
 
 
 def test_resolve_classmethod(simple_app):
     app_client = simple_app.app.test_client()
     resp = app_client.get('/v1.0/resolver-test/classmethod')  # type: flask.Response
-    assert resp.data.decode() == '"DummyClass"\n'
+    assert resp.data.decode('utf-8') == '"DummyClass"\n'
 
 
 def test_add_api_with_function_resolver_function_is_wrapped(simple_api_spec_dir):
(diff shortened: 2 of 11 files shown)


hjacobs:
Thanks for the PR! Currently all the Travis CI tests fail.. (haven't checked the root cause)

rafaelcaricio:
For this PR there is an actual problem. Since we are encoding now with `utf-8` the returned valus are different and tests are failing. Example:

```
    def test_path_parameter_somefloat(simple_app):
        resp = app_client.get('/v1.0/test-float-path/123.45')  # type: flask.Response
>       assert resp.data.decode() == '"float"\n'
E       assert 'b\'"float"\'\n' == '"float"\n'
E         - b'"float"'
E         ? --       -
E         + "float"
```

The endpoints are returning `b'"float"'` instead of `"float"` and it is a valid JSON. Then all `json.loads` in the Connexion response are failing.

LeenaBhegade:
Hi @rafaelcaricio 

Thanks for your comment.
In my case for client side testing swagger has generated code for me  - resp.data.decode('utf-8'). 
Python 2.7 will by default use ascii for resp.data.decode()

From web browsers when I trigger the request the responses are decoded properly in utf-8 without any issues.

If we pass 'utf-8' parameter to the decoding method it should work I think. 

If you don't want that change then I can alternatively pass the encoding=utf-8 parameter conditionally when Flask config parameter app.app.config['JSON_AS_ASCII']=False is true.

Please let me know what you think is the best option ?

rafaelcaricio:
@LeenaBhegade I think the best option is to update the tests then to decode properly with 'utf-8'. Thank you for working on this.

rafaelcaricio:
BTW, please rebase with the latest changes from master.

LeenaBhegade:
Hi @rafaelcaricio 

I am happy to make changes to the tests. 

Thanks,
Leena

LeenaBhegade:
hi @rafaelcaricio,

I have added the "utf-8" parameter in the decoding function in all the tests. This has fixed all the test cases in python 2.x tests.

 But python 3.x  test cases are failing due to two errors
1) Assertion error assert 'b\'"DummyClass"\'\n' == '"DummyClass"\n'
2) json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

The decode function in Python3.x somehow is not stripping the prefix 'b' from the response. This is failing the 1) assertion and 2) json.loads (as it expects a bracket and not a character at the beginning)

A prefix of 'b' or 'B' is ignored in Python 2. Python 3 clearly represent byte string and they are still valid strings.

We could either not perform any encoding in python 3.x as utf-8 encoding is well handled in Python3 internally or strip the prefix b's off before assertion and json.loads.

I will continue to check further in details if it is some other reason for the problem.

rafaelcaricio:
@LeenaBhegade The way to go is to do a conditional for Python2.7 and Python3.x. Splitting the "b" from the string representation is not an acceptable solution. You can use six package for that.

LeenaBhegade:
hi @rafaelcaricio 

Thanks for your comment. I will implement the changes soon.

Thanks,
Leena

rafaelcaricio:
@LeenaBhegade Thank you for taking care of this!

LeenaBhegade:
@rafaelcaricio  No worrries. I am glad I could make some contribution towards it.

connexion/decorators/produces.py
```diff
@@ -76,7 +77,12 @@ def dumps(data):
         """ Central point where JSON serialization happens inside
         Connexion.
         """
-        return "{}\n".format(json.dumps(data, indent=2))
+        if six.PY2 == True:
```
rafaelcaricio: `if six.PY2:` should do.

connexion/decorators/produces.py
```diff
@@ -76,7 +77,12 @@ def dumps(data):
...
+            encoding = "utf-8"
+        else:
+            encoding = None
+
+        return "{}\n".format(json.dumps(data, indent=2, encoding=encoding))
```
rafaelcaricio: Maybe better to do:
```python
    json_content = json.dumps(data, indent=2, encoding='utf-8')
else:
    json_content = json.dumps(data, indent=2)

return "".format(json_content)
```

LeenaBhegade: Thanks for your comments. Will do.

LeenaBhegade: @rafaelcaricio,

you have suggested return "".format(json_content) 
but why are we not doing  return "{}\n".format(json_content) as it was being returned originally ?

hjacobs: @LeenaBhegade I assume it was a typo and @rafaelcaricio meant "{}\n".format(...)

rafaelcaricio:
My mistake. I would prefer to leave it exactly as it was before.

rafaelcaricio:
@LeenaBhegade You can add as many test cases as you think is necessary.

LeenaBhegade:
Sorry I have been pushing the code before testing the test cases. Somehow can't get the test cases running locally. Continuously getting module import errors.

rafaelcaricio:
@LeenaBhegade It's ok, you can squash the commits later.

LeenaBhegade:
@rafaelcaricio 

All the tests are passing except for Python 3.5 version  --- >ERROR: InvocationError: '/home/travis/build/zalando/connexion/.tox/flake8/bin/python setup.py flake8'

If the above error is not related to this change and you are happy with the test and changes, please could you approve the changes.

connexion/decorators/produces.py
rafaelcaricio: the `== True` is not necessary here.

rafaelcaricio:
@LeenaBhegade Thank you for working on this. I would like to kindly ask you to squash your commits. Please find out more about it here: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

LeenaBhegade:
@rafaelcaricio Done. Please approve whenever you get a chance.

rafaelcaricio:
@LeenaBhegade I would like to point out one of my code review requests was ignored: https://github.com/zalando/connexion/pull/397/files#r100395476

Please change that so we can get this merged.

LeenaBhegade:
@rafaelcaricio  
Sorry about missing that out. I have made the changes.
Thanks
Leena

rafaelcaricio:
Please apply this last change:
```patch
--- /home/travis/build/zalando/connexion/connexion/decorators/produces.py:before	2017-02-12 12:47:19.333433
+++ /home/travis/build/zalando/connexion/connexion/decorators/produces.py:after	2017-02-12 12:50:03.352099
@@ -2,9 +2,9 @@
 import datetime
 import functools
 import logging
-import six
 
 import flask
+import six
 from flask import json
 
 from .decorator import BaseDecorator
```

Basically it is just to move the line `import six` line to after `import logging`.

I am sorry to ask so many changes not directly related to your primary objective. But we need to keep the code quality of patches. I know it may be frustrating, but I hope you understand.

All those checks you could have executed in your local machine by running the command `tox` before pushing to the remote branch. Of course you would need to install all python versions we need to run the tests and checks.

LeenaBhegade:
Np.I apologise for missing out these little things. I should have taken care.

rafaelcaricio:
👍

rafaelcaricio:
@LeenaBhegade Thank you for working on this. PR merged! :1st_place_medal:

LeenaBhegade:
Thanks for letting me to do the work!

LeenaBhegade:
Hi @rafaelcaricio 

Just wanted to know when I would be able to install the new version of  connexion via pip ?

Thanks,
Leena

rafaelcaricio:
@LeenaBhegade I will be releasing a new version today. I will let you know.

rafaelcaricio:
@LeenaBhegade Sorry the delay. A new version of Connexion with your contribution has been just released https://pypi.python.org/pypi/connexion version `1.1.5`. Thank you again! 👍

PR #400, opened 2017-02-09
Update swagger ui to v2.2.8

antyale:

 - Update swagger ui to v2.2.8


rafaelcaricio:
Thank you for the PR. There are some patches that need to be applied on top of the Swagger UI code. Those patches are customisations added by Connexion, without it we cannot accept this upgrade. Also unfortunately it is impossible to review this code (guarantee that it is the real code from the Swagger UI repo), so I cannot accept this PR.

The upgrade/use of the Swagger UI should be handled differently.

Related issue #406

PR #403, opened 2017-02-16
Improve the release script

rafaelcaricio:

 - Set script flags explicitly for better maintainability;
 - Make script portable. This is to be used in MacOS systems (`sed` command interface is slightly different);
 - Use `twine` (https://packaging.python.org/distributing/#upload-your-distributions) to securely upload new packages versions to PyPi;
 - Signs new release tags.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -22,7 +22,7 @@ fi
 tox --skip-missing-interpreters
 
 python3 setup.py sdist bdist_wheel
-#twine upload dist/*
+twine upload dist/*
 
 # revert version
 git checkout -- */__init__.py
--- a/release.sh
+++ b/release.sh
@@ -5,21 +5,27 @@ if [ $# -ne 1 ]; then
     exit 1
 fi
 
-set -xe
+set -o errexit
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -5,21 +5,27 @@ if [ $# -ne 1 ]; then
...
+set -o xtrace
 
 python3 --version
 git --version
 
 version=$1
 
-sed -i "s/__version__ = .*/__version__ = '${version}'/" */__init__.py
+if [[ "$OSTYPE" == "darwin"* ]]; then
+	sed -i "" "s/__version__ = .*/__version__ = '${version}'/" */__init__.py
```
hjacobs: I don't get what this does and why it would be needed on Mac...

rafaelcaricio: On MacOS you need to define the backup file extension when replacing contents with `sed`. Ref: http://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

hjacobs:
:+1:

PR #404, opened 2017-02-16
Remove unnecessary dependency on strict-rfc3339

Griffon26:
This pull request removes the unnecessary dependency on strict-rfc3339 specified in setup.py and requirements.txt. Calls to functions of this module were removed in commit 559af21f90e833044be4acdf872ef97d42cb0181 and the actual import in 6554de8dc38a0d04ee4a73b6e4e1a43c2822ab66, both of which were made some time in 2015.

Since strict-rfc3339 is licensed under the GPLv3 and connexion under the Apache 2.0 license, this change also makes it more clear that there is no licensing issue any more.

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -1,7 +1,6 @@
 -e git+https://github.com/Julian/jsonschema.git#egg=jsonschema
 -e git+https://github.com/pallets/flask.git#egg=flask
 -e git+https://github.com/kennethreitz/requests#egg=requests
--e git+https://github.com/danielrichman/strict-rfc3339.git#egg=strict-rfc3339
 -e git+https://github.com/Yelp/swagger_spec_validator.git#egg=swagger-spec-validator
 -e git+https://github.com/zalando/python-clickclick.git#egg=clickclick
 -e hg+https://bitbucket.org/pitrou/pathlib#egg=pathlib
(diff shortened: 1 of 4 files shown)


hjacobs:
Hmm, tests are failing for all Python versions?

Griffon26:
Shame on me. I should have known better than to submit a pull request in a hurry.

I've looked into the test failure and it turns out that some of the tests depend on validation of date-time, for which jsonschema requires its "format" extras to be installed, which includes strict-rfc3339 (and rfc3987, which is also under GPLv3).

If I understand correctly, that means an application using connexion to validate date-time can only be licensed under the GPLv3. Do you agree?

I'm unsure what this means for the license of jsonschema, and by extension the licenses of swagger-spec-validator and connexion.

In any case, how would you suggest to fix this? Add `extras_require = { 'format' : [ 'jsonschema[format]' ] }` and `tests_require=['jsonschema[format]']`?

rafaelcaricio:
@Griffon26 I think we should indeed remove this dependency. Please change this test https://github.com/zalando/connexion/blob/master/tests/fixtures/different_schemas/swagger.yaml#L247 to a different format (maybe `uri`) and it will pass. Then we have to check how to add back the `date-time` validation.

rafaelcaricio:
👍

PR #411, opened 2017-03-06
Fixes #377 | resolve_function_from_operation_id hides cause of import…

### LINKED ISSUE #377: resolve_function_from_operation_id hides cause of import error
### Description
This has happened a couple of times to me now. Somehow I introduce an error into my code that my IDE does not detect and stops the import process from working. When I start the application I get error messages for failed opreationIds that do not help me to find the root cause of the problem.

Is this a configuration problem my end or is this a valid problem that we could improve on?

### Expected behaviour
The resolver correctly finds function resources.report.pbx_resource.delete but on import gets an error from an imported class that does not meet an interface requirement, for example:

"cannot import name MySupportingClass" 

If the error message returned included the actual import error, it would be easy to find the class that is causing the problem and fix it.

### Actual behaviour
The error message returned just tells me which path failed to resolve, but does not report back the underlying issue with the import:

```
ERROR:connexion.api:Failed to add operation for DELETE /v1/report/pbx/domainId/{domainId}/id/{id}
  File "/var/www/public-api/env2/src/connexion/connexion/api.py", line 233, in add_paths
    self.add_operation(method, path, endpoint, path_parameters)
  File "/var/www/public-api/env2/src/connexion/connexion/api.py", line 185, in add_operation
    resolver=self.resolver)
  File "/var/www/public-api/env2/src/connexion/connexion/operation.py", line 166, in __init__
    resolution = resolver.resolve(self)
  File "/var/www/public-api/env2/src/connexion/connexion/resolver.py", line 39, in resolve
    return Resolution(self.resolve_function_from_operation_id(operation_id), operation_id)
  File "/var/www/public-api/env2/src/connexion/connexion/resolver.py", line 66, in resolve_function_from_operation_id
    raise ResolverError(msg, sys.exc_info())
ResolverError: <ResolverError: Cannot resolve operationId "resources.report.pbx_resource.delete"!>
```

This makes it very difficult to debug and find out what is causing the operation resolution to fail.

### Additional info:
We could do something like this on line 63 of resolver.py:
```
except ImportError as e:
    import_error = e.message
```
This import error string could be used a part of the error message for ResolverError. Currently we discard the import error and just report the problem operationId.

```
 54     def resolve_function_from_operation_id(self, operation_id):
 55         """
 56         Invokes the function_resolver
 57 
 58         :type operation_id: str
 59         """
 60         msg = 'Cannot resolve operationId "{}"!'.format(operation_id)
 61         try:
 62             return self.function_resolver(operation_id)
 63         except ImportError:
 64             import sys
 65             raise ResolverError(msg, sys.exc_info())
 66         except AttributeError:
 67             raise ResolverError(msg)
```

Output of the commands:

- `python --version`
Python 2.7.12
Version: 2016.0.dev1

danballance:
Fixes #377 by extending the log message if there's an ImportError.

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -60,7 +60,9 @@ class Resolver(object):
         msg = 'Cannot resolve operationId "{}"!'.format(operation_id)
         try:
             return self.function_resolver(operation_id)
-        except ImportError:
+        except ImportError as e:
+            if str(e):
+                msg += " Import error was '{0}'".format(str(e))
             import sys
             raise ResolverError(msg, sys.exc_info())
         except AttributeError:
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -62,7 +62,7 @@ class Resolver(object):
             return self.function_resolver(operation_id)
         except ImportError as e:
             if str(e):
-                msg += " Import error was '{0}'".format(str(e))
+                msg = "{} Import error was '{}'".format(msg, str(e))
             import sys
             raise ResolverError(msg, sys.exc_info())
         except AttributeError:
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -61,8 +61,7 @@ class Resolver(object):
         try:
             return self.function_resolver(operation_id)
         except ImportError as e:
-            if str(e):
-                msg = "{} Import error was '{}'".format(msg, str(e))
+            msg = "{} Import error was '{}'".format(msg, str(e))
             import sys
             raise ResolverError(msg, sys.exc_info())
         except AttributeError:


connexion/resolver.py
```diff
@@ -60,7 +60,9 @@ def resolve_function_from_operation_id(self, operation_id):
         msg = 'Cannot resolve operationId "{}"!'.format(operation_id)
         try:
             return self.function_resolver(operation_id)
-        except ImportError:
+        except ImportError as e:
+            if str(e):
```
hjacobs: why do we have `if str(e)`, can `str(e)` be empty/False?

connexion/resolver.py
```diff
@@ -60,7 +60,9 @@ def resolve_function_from_operation_id(self, operation_id):
...
+                msg += " Import error was '{0}'".format(str(e))
```
rafaelcaricio: Please use this format:
```patch
@@ -65 +65 @@
-msg += " Import error was '{0}'".format(str(e))
+msg = '{} - Import error was: {}'.format(msg, str(e))
```

hjacobs:
:+1:

PR #413, opened 2017-03-08
Support Decimal values in JSONEncoder

### LINKED ISSUE #412: JSON encoding of Decimal fails
### Description

Decimal values could be returned from DB queries.
This issue could be fixed by patching ``app.json_encoder``, but I thought it could be useful to be part of the base connexion encoder.

```python
TypeError: Decimal('0') is not JSON serializable
```

### Expected behaviour

API responds with valid JSON including float values (e.g. ``0.1`` or ``1.1e-05`` etc ...)

### Actual behaviour

JSON serialization fail

### Steps to reproduce

```python
import json
from decimal import Decimal
from connexion.decorators.produces import JSONEncoder

json.dumps(Decimal(0.1), cls=JSONEncoder)
```

### Additional info:
http://www.json.org/
http://stackoverflow.com/questions/19554972/json-standard-floating-point-numbers

Output of the commands:

- `python --version`
``Python 3.5.2``

``Version: 1.1.4``

mohabusama:
Fixes #412 .

 - Update JSONEncoder to handle ``Decimal`` values.

### DIFF
--- a/connexion/decorators/produces.py
+++ b/connexion/decorators/produces.py
@@ -3,6 +3,8 @@ import datetime
 import functools
 import logging
 
+from decimal import Decimal
+
 import flask
 import six
 from flask import json
@@ -29,6 +31,9 @@ class JSONEncoder(json.JSONEncoder):
 
         if isinstance(o, datetime.date):
             return o.isoformat()
+
+        if isinstance(o, Decimal):
+            return float(o)
 
         return json.JSONEncoder.default(self, o)
 
--- a/tests/test_produces.py
+++ b/tests/test_produces.py
@@ -1,5 +1,8 @@
 import datetime
 import json
+import math
+
+from decimal import Decimal
 
 from connexion.decorators.produces import JSONEncoder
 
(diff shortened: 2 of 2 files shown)


connexion/decorators/produces.py
```diff
@@ -30,6 +32,9 @@ def default(self, o):
         if isinstance(o, datetime.date):
             return o.isoformat()
 
+        if isinstance(o, Decimal):
+            return float(o)
```
hjacobs: but this would lose precision..

hjacobs:
Can you check the Travis build?

connexion/decorators/produces.py
mohabusama: We can fallback to strings for precise representation, but I am not sure if this is the desired ``default`` behavior. I felt using floats (👿 ) is somehow expected/acceptable, unless precision is required which in this case dev would use their own implementation.

mohabusama:
Build looks fine after a restart.

hjacobs:
Thanks for the PR!

hjacobs:
:+1:

hjacobs:
:+1:

PR #414, opened 2017-03-09
Fixes #340 | Convert camelcased parameter names to Python style

### LINKED ISSUE #340: Convert camelcased parameter names to Python style convention
### Description
It's common to define Swagger parameter names in camelcase ("postalCode", "modifiedDate", "statusCode", etc.), but this goes against Python PEP-8 convention.

### Expected behaviour

From Swagger:

    get:
      parameters:
        - name: postalCode
          in: query
          type: string

I would like to be able to write

    def search(postal_code, ...):
        ...

### Actual behaviour

The parameters are not found:

    TypeError: search() missing 1 required positional argument: 'postal_code'

### Additional info:

Output of the commands:

$ python --version
Python 3.5.2
Version: 1.0.129

Here's a regex for the conversion: http://stackoverflow.com/a/1176023/594211

danballance:
Fixes #340 | Convert camelcased parameter names to Python style

 - added config variable pythonic_params
 - parameter changes made in parameter_to_arg
 - sanitize_param made a local function within parameter_to_arg so that it can access pythonic_params as a closure and keep the code a little cleaner
- unit tests added with fakeapi and swagger spec

### DIFF
--- a/connexion/api.py
+++ b/connexion/api.py
@@ -62,7 +62,8 @@ class Api(object):
     def __init__(self, specification, base_url=None, arguments=None,
                  swagger_json=None, swagger_ui=None, swagger_path=None, swagger_url=None,
                  validate_responses=False, strict_validation=False, resolver=None,
-                 auth_all_paths=False, debug=False, resolver_error_handler=None, validator_map=None):
+                 auth_all_paths=False, debug=False, resolver_error_handler=None,
+                 validator_map=None, pythonic_params=False):
         """
         :type specification: pathlib.Path | dict
         :type base_url: str | None
@@ -81,6 +82,9 @@ class Api(object):
         :param resolver_error_handler: If given, a callable that generates an
             Operation used for handling ResolveErrors
         :type resolver_error_handler: callable | None
+        :param pythonic_params: When True CamelCase parameters are converted to snake_case and an underscore is appended
+        to any shadowed built-ins
+        :type pythonic_params: bool
         """
         self.debug = debug
         self.validator_map = validator_map
@@ -142,6 +146,9 @@ class Api(object):
         logger.debug('Strict Request Validation: %s', str(validate_responses))
         self.strict_validation = strict_validation
 
+        logger.debug('Pythonic params: %s', str(pythonic_params))
+        self.pythonic_params = pythonic_params
+
         # Create blueprint and endpoints
         self.blueprint = self.create_blueprint()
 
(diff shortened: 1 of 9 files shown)


rafaelcaricio:
@danballance Thank you for the PR! Before I go through the code, could you please check what is causing the tests to fail under the Python interpreter versions >=3.4 ?

Also please fix the ordering of the imports applying this patch:
```patch
--- /home/travis/build/zalando/connexion/connexion/decorators/parameter.py:before	2017-03-09 16:58:03.933834
+++ /home/travis/build/zalando/connexion/connexion/decorators/parameter.py:after	2017-03-09 17:00:57.317468
@@ -3,17 +3,20 @@
 import inspect
 import logging
 import re
+
+import flask
 import inflection
+import six
+import werkzeug.exceptions as exceptions
+
+from ..utils import all_json, boolean, is_null, is_nullable
+
 try:
     import builtins
 except ImportError:
     import __builtin__ as builtins
 
-import flask
-import six
-import werkzeug.exceptions as exceptions
 
-from ..utils import all_json, boolean, is_null, is_nullable
 
 logger = logging.getLogger(__name__)
```

danballance:
@rafaelcaricio certainly, apologies for this, I will look at this today.

danballance:
Blimey, apologies about all of that noise. I _think_ I now have my local development environment set up in such a way that I will catch this stuff locally in future. Do you guys have a doc anywhere with a recommended approach to testing? I had a quick look and couldn't see anything. I'm still not 100% I have things configured optimally.

In terms of the actual PR I guess the main topic to consider is how I am checking the list of built-ins and expecting the called function to have an builtin_ format parameter name for any python built-ins used in the swagger specification. (So  for example path /resource/{id} expects the entry point function to be using id_ for its parameter name.) As I just found out in my unit tests, of course the list of built-ins changes between python versions and would mean that some parameter names would not work. For example reduce is a built-in in Python 2.7 but not in 3.5 so in Python 2.7 we'd require parameter name reduce_ but in Python 3.5 we'd look for the unmodified "reduce".

An alternative approach could be to maintain a common list of variable names such as id and list that we we will handle and not try to handle every built in. Or we could of course drop the automatic handling of built-in shadowed variables.

What do you think is the best way to proceed here? I can make some changes if you think there is a better approach :)

rafaelcaricio:
@danballance Thank you for your PR again and for taking care that all the tests pass.

Your approach looks good to me. I wouldn't worry by the builtins being different between Python versions.

rafaelcaricio:
👍

PR #415, opened 2017-03-14
 Security decorator should log client IP address

### LINKED ISSUE #410: Security decorator should log client IP address
### Description
We use a project called Fail2Ban on most of our web services in order to block brute force attacks. Fail2Ban can be configured to scan custom application logs and look for authentication failures and block the offending IPs. Obviously in order to do this it needs the IP address to be recorded in the log.

### Expected behaviour
Logs lines look like this:

- Token info (200) for client IP 123.123.123.123
- ... User scopes (set([u'reporting:write', u'reporting:read'])) do not match the scopes necessary to call endpoint (set(['pbx.provisioning:read'])). Aborting with 403 for client IP 123.123.123.123.

### Actual behaviour
Logs lines look like this:

- Token info (200)
- ... User scopes (set([u'reporting:write', u'reporting:read'])) do not match the scopes necessary to call endpoint (set(['pbx.provisioning:read'])). Aborting with 403.

### Steps to reproduce
Set up authentication and try to authenticate with an invalid access token and then insufficient scope.

Output of the commands:
- `python --version`

Python 2.7.12
Version: 2016.0.dev1

Would you accept a pull request for me to add the IPs into the log lines with the above format with request.remote_addr? The decorator is already accessing the Flask request object, so it should be a simple change to make.

danballance:
Fixes #410 

 - trusted_ips list optionally passed in at application start. If passed then we'll only trust X-Forwarded-For if remote_addr is in the list of trusted IPs (i.e. known proxies under our control)
- "for client IP 'x.x.x.x" format string is added to log lines
- testfixtures added to setup.py in order to make use of LogCapture. I wouldn't normally recommend testing log lines (it does feel fragile) but in this case we need to be sure the correct IP is logged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,8 @@ tests_require = [
     'decorator',
     'mock',
     'pytest',
-    'pytest-cov'
+    'pytest-cov',
+    'testfixtures'
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ install_requires = [
     'six>=1.9',
     'strict-rfc3339>=0.6',
     'swagger-spec-validator>=2.0.2',
+    'inflection>=0.3.1'
 ]
 
 if py_major_minor_version < (3, 4):
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -1,7 +1,9 @@
 import pytest
 from connexion.decorators.security import get_tokeninfo_url, verify_oauth
-from connexion.exceptions import OAuthProblem
-from mock import MagicMock
+from connexion.exceptions import (OAuthProblem, OAuthResponseProblem,
+                                  OAuthScopeProblem)
+from mock import MagicMock, PropertyMock
+from testfixtures import LogCapture
 
 
 def test_get_tokeninfo_url(monkeypatch):
@@ -25,7 +27,7 @@ def test_verify_oauth_invalid_auth_header(monkeypatch):
     def func():
         pass
 
-    wrapped_func = verify_oauth('https://example.org/tokeninfo', set(['admin']), func)
+    wrapped_func = verify_oauth('https://example.org/tokeninfo', set(['admin']), [], func)
 
     request = MagicMock()
     app = MagicMock()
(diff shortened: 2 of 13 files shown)


setup.py
```diff
@@ -42,7 +43,8 @@ def read_version(package):
     'decorator',
     'mock',
     'pytest',
-    'pytest-cov'
+    'pytest-cov',
+    'testfixtures'
```
rafaelcaricio: Why this ?

rafaelcaricio:
Could you please rebase this PR? I am seeing a lot of changes related to the PR #414.

tests/decorators/test_security.py
```diff
@@ -34,3 +36,137 @@ def func():
...
+    def func():
+        pass
+
+    app = MagicMock()
+    monkeypatch.setattr('flask.current_app', app)
+
+    # 1) remote_addr, no x-forwarded-for, invalid token (401)
+    with LogCapture() as l:
+        wrapped_func = verify_oauth('https://example.org/tokeninfo', set(['admin']), [], func)
+        request, headers = _setup_request_mock(remote_addr="123.123.123.123")
```
rafaelcaricio: Better than doing all this mocking is to use `test_request_context`. It is safer because we relay on how Flask actually works and not how the mocks are set up.

```
api_client.application.test_request_context('/').push()
```
Reference:
- http://flask.pocoo.org/docs/0.12/reqcontext/
- http://flask.pocoo.org/docs/0.12/api/#flask.Flask.test_request_context

rafaelcaricio:
@danballance Thank you for the PR. Would be nice if we could have some documentation on what it does and how this works. What do you think? Do not need to be long. Basically a summary of your comments from #410.

tests/decorators/test_security.py
danballance: Okay, no problem, I will review this code. A few things to clean up. Thanks for the feedback.

rafaelcaricio:
@danballance Please review the changes made in this PR. It needs rebasing.

danballance:
Hi,

I've just been looking for a gap here where I can look again at this. Tomorrow may be possible - certainly this week.

danballance:
I am looking at this, but 1.1.7 appears to be a major re-factor and lots of things have broken. I still hope to resolve everything this week, but it's not going to be tomorrow now.

rafaelcaricio:
@danballance Yes, sorry about that. It was an expected internal refactor to move the project forward (see #380).

danballance:
No problem - it makes complete sense to decouple Flask. It's a big change though! I can see you guys are still changing quite a lot of code so I think I'll wait until the dust settles before I return to this :)

rafaelcaricio:
@danballance Sure, no problem. Yeah, now we are fixing some things since the decoupling started.

PR #424, opened 2017-04-04
Flask required only if necessary

rafaelcaricio:
Related to #380

Only requires Flask package to be installed when using specifically related functionality.


README.rst
```diff
@@ -63,7 +64,7 @@ Why Connexion
 
 With Connexion, you write the spec first. Connexion then calls your Python
 code, handling the mapping from the specification to the code. This
-incentivizes you to write the specification so that all of your
+incentives you to write the specification so that all of your
```
hjacobs: I think "incentivizes" was correct!

rafaelcaricio: Changing it back...

hjacobs:
@rafaelcaricio can we first release Connexion without this breaking change? This breaking change (existing users need to update their requirements to pull in Flask) should also get a major version number bump I guess. Let's first roll out the Framework changes in a compatible way (no changes necessary by users) and see if it works for everybody.

rafaelcaricio:
@hjacobs I was thinking about that. But still this changes here are important. I will just change the `setup.py` for now to include Flask by default and change back the README.

rafaelcaricio:
@hjacobs Done in https://github.com/zalando/connexion/pull/424/commits/86891ab74df0bdbf68051327432e7fe05a026cd2.

rafaelcaricio:
Changes in this PR will make easier to release Connexion 2.x in near future. I still want to make some more changes before going for 2.x, mostly related to https://github.com/zalando/connexion/issues/283. So we release together those breaking changes.

hjacobs:
@rafaelcaricio we could release the current master as version 1.*. What do you think?

vimalloc:
(totally selfish here), but if at all possible, I would like to get #425 in the next released version. Currently, we are doing a pip install from the forked version, but it isn't ideal.

hjacobs:
@rafaelcaricio after the latest merge -> conflicts to be resolved :smirk:

vimalloc:
Sorry about that :( Looks like it is just imports, shouldn't be too hard to fix. I *think* after the resolution it should look like

```python
￼from flask import jsonify, redirect, request
￼  
￼from connexion import NoContent, ProblemException, problem
```

rafaelcaricio:
@vimalloc Don't worry. I am fixing that. 😉 

@hjacobs Yes, after merging this PR we can release a new version of Connexion still under 1.x. If people are not playing with Connexion's internals in their services everything should continue work as usual.

hjacobs:
@rafaelcaricio I felt adventurous and released https://github.com/zalando/connexion/releases/tag/1.1.6

rafaelcaricio:
@hjacobs it is ok, but flask was removed from the `setup.py` in the current master. This PR puts it back. So after merging this, we should release again with this "bug fix".

hjacobs:
@rafaelcaricio uh, I did not see this!

hjacobs:
:+1:

hjacobs:
@rafaelcaricio go for it, my GnuPG setup is also broken atm, so you can do `tag -s`.

PR #425, opened 2017-04-04
Fix returning flask Response objects in a tuple

vimalloc:
Allows these flask paradigms to work in connexion endpoints:

```python
return jsonify({'foo': 'bar'}), 200
```

```python
return jsonify({'foo': 'bar'), {'new_header': 'header_value'}
```

```python
return jsonify({'foo': 'bar'}), 201, {'new_header': 'header_value'}
```

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -172,6 +172,23 @@ class FlaskApi(AbstractAPI):
 
         return flask_response
 
+    @staticmethod
+    def _update_flask_response(response, status_code=None, headers=None):
+        """
+        Takes an object that is already a flask Response, and updates it
+        with the given status code and headers.
+
+        This allows one to use the common flask paradigm
+        ```return jsonify(foo='bar'), 201```
+        or
+        ```return jsonify(foo='bar'), 201, {'new_header': 'header_value'}```
+        """
+        if headers:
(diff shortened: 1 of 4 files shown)


rafaelcaricio:
@vimalloc Could you please write a test for this new code?

rafaelcaricio:
👍

hjacobs:
Stupid question: why would anybody return a tuple where only the first element is actually used (`response[0]`)? Or did I get it wrong?

connexion/apis/flask_api.py
```diff
@@ -208,6 +208,9 @@ def _get_flask_response(cls, response, mimetype):
         if flask_utils.is_flask_response(response):
             return response
 
+        elif isinstance(response, tuple) and flask_utils.is_flask_response(response[0]):
```
hjacobs: isn't this throwing away all the other tuple elements?

vimalloc: Here we are only checking if the first element in the tuple is already a flask response, along the lines of (<Response>, status_code, headers). If so, we send the full tuple to the underlying flask code which turns a returned tuple into a flask response: 

https://github.com/pallets/flask/blob/master/flask/app.py#L1698

hjacobs: Right, my fault

PR #447, opened 2017-05-08
Added python version check for Jsonifier dumps

mvrogowski:
Fix encoding bug in Jsonifier.dumps() for Python 2.7.

```
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/lib/python2.7/site-packages/connexion/decorators/decorator.py", line 66, in wrapper
    response = function(request)
  File "/usr/lib/python2.7/site-packages/connexion/decorators/validation.py", line 293, in wrapper
    return function(request)
  File "/usr/lib/python2.7/site-packages/connexion/decorators/decorator.py", line 43, in wrapper
    return self.api.get_response(response, self.mimetype, request)
  File "/usr/lib/python2.7/site-packages/connexion/apis/flask_api.py", line 149, in get_response
    flask_response = cls._get_flask_response(response, mimetype)
  File "/usr/lib/python2.7/site-packages/connexion/apis/flask_api.py", line 224, in _get_flask_response
    return cls._build_flask_response(mimetype=mimetype, data=response)
  File "/usr/lib/python2.7/site-packages/connexion/apis/flask_api.py", line 189, in _build_flask_response
    data = cls._jsonify_data(data, mimetype)
  File "/usr/lib/python2.7/site-packages/connexion/apis/flask_api.py", line 201, in _jsonify_data
    return cls.jsonifier.dumps(data)
  File "/usr/lib/python2.7/site-packages/connexion/apis/flask_api.py", line 23, in dumps
    return "{}\n".format(flask.json.dumps(data, indent=2))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 141: ordinal not in range(128)
```

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -254,7 +254,12 @@ class Jsonifier(object):
         """ Central point where JSON serialization happens inside
         Connexion.
         """
-        return "{}\n".format(flask.json.dumps(data, indent=2))
+        if six.PY2:
+            json_content = flask.json.dumps(data, indent=2, encoding="utf-8")
+        else:
+            json_content = flask.json.dumps(data, indent=2)
+
+        return "{}\n".format(json_content)
 
     @staticmethod
     def loads(data):


hjacobs:
@mvrogowski thanks for the PR! Could you add a test for this (we are running Python 2.7 and 3.* on Travis)?

mvrogowski:
Sure I can, @hjacobs! I'll do it ASAP.

mvrogowski:
Hey, @hjacobs!

Actually, I guess the statements are already evaluated by current tests - tests/api/test_responses.py::test_get_unicode_response() to be more specific.

What happens is that when running tests for Python 2.7, the line 260 is not evaluated; for Python 3, the line 258 is not evaluated.

If these tests are not enough, let me know. :)

jmcs:
For some reason I can't replicate this exact error, but wouldn't [PEP 263 -- Defining Python Source Code Encodings](https://www.python.org/dev/peps/pep-0263/) or use unicode strings fix this? I don't like the idea of hardcoding utf-8 encoding here because other people may be using other encodings, and this could break their code.

mvrogowski:
Hi, @jmcs! How is it going?

To replicate the error, try to return a response with a non-ASCII char, like "é" (this is the exactly char that was raising the exception) - actually, it's a very weird problem, because sometimes it returns ok, sometimes not. But, setting the encoding, it works flawless.

About hardcoding the UTF-8, as there are another places on code that works like this, I realized it's ok.

Nevertheless, I'll do a double check wth your suggestion.

hjacobs:
> To replicate the error, try to return a response with a non-ASCII char, like "é" (this is the exactly char that was raising the exception) - actually, it's a very weird problem, because sometimes it returns ok, sometimes not. But, setting the encoding, it works flawless.

@mvrogowski could you add a test which fails without your change, i.e. test an actual non-ASCII char like you described?

mvrogowski:
Sure, @hjacobs! I will try to do it today.

Issue #449, opened 2017-05-10
path formatted parameters


kcacciatore:
### Description
Hi - I'm trying to build an api that allows extensions to be added to the API created. For example:

`/v1/service/{serviceid}/{extension}`

Where the method invoked would pass {extension} as the remainder of the URL as a parameter to the invoked code, allowing it to process that part of the URL. 

I understand and have tested in Flask something like this:
`@app.route('/myapp/<serviceId>/<path:extension>', methods = ['POST', 'GET'])`

When I post something like:
`/service/myservice/my/custom/extension`
and as expected get serviceId='myservice' as one parameter and the remainder of the url (e.g. 'my/custom/extension' ) as the value for extension. 

I built what I believe to be equivalent using OpenAPI:
```
paths:
  /service/{serviceid}/{extension}:
    post:
      tags:
      - "serviceextension"
      summary: "Extension endpoint for services""
      operationId: "my_extension_service_by_id"
      produces:
      - "application/json"
      parameters:
      - name: 'serviceid'
        in: path
        type: string
        required: true
      - name: 'extension'
        in: path
        type: string
        required: true
        format: path

```
### Expected behaviour

I expect when posting an URL:

`/service/myservice/my/custom/extension`

I would have the my_extension_service_by_id invoked with  serviceid='myservice' and extension='my/custom/extension'

### Actual behaviour

I get a 404 error

### Steps to reproduce

Create an API as above - I used http://editor.swagger.io/#/ and generated the server stubs using python-flask

### Additional info:
python --version
Python 3.6.0b2

pip show connexion | grep "^Version\:"
Version: 1.0.129
Output of the commands:

- `python --version`
- `pip show connexion | grep "^Version\:"`

hjacobs:
I guess this makes sense. I could not find anything regarding constraints for path parameters in the OpenAPI spec, so it should not be valid in Swagger.

We are open for PRs :smile:

kcacciatore:
Sure - I just started using swagger etc.. a couple of weeks ago, but I will see what I can do.

kcacciatore:
Added pull request for a test I wrote before changing any core code - it showed me that the most recent codebase works just fine, and it's the version of connexion that is delivered with swagger stub generation that needs updating.

hjacobs:
@kcacciatore great to hear :smile:

PR #450, opened 2017-05-10
449 path format

kcacciatore:
Addresses #449 (adds test) .

None. I wrote a test based on what I expected the behavior to be, prior to making any code changes, and the test passes.  turns out I was using a version of connexion that came with the generated swagger.io stubs, and it is older (connexion-1.0.129), and the path formatting seems to not work as of that version. I upgraded connexion, fixed a small JSON encoding issue with the stub code, and it now runs as expected. But here's a test that covers the scenario I described.

None, just another coverage point.


hjacobs:
:+1:

rafaelcaricio:
Contract signed. 😄

PR #454, opened 2017-05-23
Improve message if operationId has no model prefix

MikeRalphson:
Fixes # n/a

 - Improves error message on importing an OpenAPI definition which does not include module prefixes on its operationIds

An alternative solution would be to provide a default module name as a command-line parameter.

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -1,6 +1,8 @@
 import functools
 import importlib
 
+from connexion.exceptions import ResolverError
+
 
 def deep_getattr(obj, attr):
     """
@@ -17,6 +19,8 @@ def get_function_from_name(function_name):
 
     :type function_name: str
     """
+    if (not function_name) or ('.' not in function_name):
+        raise ResolverError('operationIds must include a {module}. prefix')
     module_name, attr_path = function_name.rsplit('.', 1)
     module = None
     last_import_error = None
--- a/tests/fixtures/unprefixed_op_id/swagger.yaml
+++ b/tests/fixtures/unprefixed_op_id/swagger.yaml
@@ -1,0 +1,17 @@
+swagger: "2.0"
+
+info:
+  title: "{{title}}"
+  version: "1.0"
+
+basePath: /v1.0
+
+paths:
+  /welcome:
+    put:
+      operationId: putWelcome # has no module. prefix
+      responses:
+        200:
+          description: greeting response
+          schema:
+            type: object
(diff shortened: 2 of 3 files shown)


MikeRalphson:
Ok - need to update test-suite.

MikeRalphson:
Build now passing again.

hjacobs:
Could you add a unit test for this behavior? Thanks!

MikeRalphson:
@hjacobs test added that an unprefixed operationId does stop setup in non-debug mode and does not in debug-mode. Believe the other cases (missing, 'bad' and 'good' operationIds) are handled by existing tests.

I noticed that `test_api.py` appears to include the following lines twice:

```python
    with pytest.raises(ResolverError):
        FlaskApi(TEST_FOLDER / "fixtures/missing_op_id/swagger.yaml",
                 base_path="/api/v1.0", arguments={'title': 'OK'})
```

and

```python
    api = FlaskApi(TEST_FOLDER / "fixtures/missing_op_id/swagger.yaml",
                   base_path="/api/v1.0", arguments={'title': 'OK'}, debug=True)
    assert api.specification['info']['title'] == 'OK'
```

Is this intentional, or is it a cut-and-paste error? Should they be testing a different fixture?

rafaelcaricio:
@MikeRalphson This is not a C&P error. It is like that because in the debug mode you should be allowed to have invalid/non-existent `operationId` to make it easier for incremental development.

MikeRalphson:
> This is not a C&P error. It is like that because in the debug mode you should be allowed to have invalid/non-existent operationId to make it easier for incremental development.

@rafaelcaricio I meant that lines [74-76](https://github.com/zalando/connexion/blob/master/tests/test_api.py#L74-L76) appear to be duplicated at lines [86-88](https://github.com/zalando/connexion/blob/master/tests/test_api.py#L86-L88). 

Surely the same test code should have reproducably the same effect, so including it twice looks like an error.

rafaelcaricio:
@MikeRalphson I see now. That is indeed a duplication and should be removed.

MikeRalphson:
@rafaelcaricio should I submit that as a separate PR?

rafaelcaricio:
@MikeRalphson Yes, please.

jmcs:
@MikeRalphson are you still interested in merging this PR? If so please resolve the merge conflicts. Thanks for the PR.

hjacobs:
I'm closing this PR for now.

PR #464, opened 2017-06-08
Create a ConnexionHeaders type

ttung:

 -headers should be a list of name, value tuples, because multiple headers can be sent down with the same name.
 -The flask response construction expects either a dictionary (less flexible) or an iterable yielding (name, value) tuples.  The `__iter__` method in `ConnexionHeaders` dispatches to the inner list for that.

### DIFF
--- a/connexion/lifecycle.py
+++ b/connexion/lifecycle.py
@@ -34,4 +34,31 @@ class ConnexionResponse(object):
         self.mimetype = mimetype
         self.content_type = content_type
         self.body = body
-        self.headers = headers or {}
+        self.headers = ConnexionHeaders(headers)
+
+
+class ConnexionHeaders(object):
+    def __init__(self, headers):
+        self.headers = headers or list()
+
+    def getlist(self, name):
+        """
+        Given a header name, return the list of values that was sent with that
+        name.
+        """
+        return [v
+                for n, v in self.headers
+                if n == name]
+
+    def getfirst(self, name, defaultvalue=None):
+        """
+        Given a header name, return the first value that was sent with that
+        name.  If the header name is not found, return `defaultvalue`.
+        """
+        for n, v in self.headers:
+            if n == name:
+                return v
+        return defaultvalue
+
+    def __iter__(self):
+        return self.headers.__iter__()
--- a/connexion/problem.py
+++ b/connexion/problem.py
@@ -21,7 +21,7 @@ def problem(status, title, detail, type=None, instance=None, headers=None, ext=N
                      further information if dereferenced.
     :type instance: str
     :param headers: HTTP headers to include in the response
-    :type headers: dict | None
+    :type headers: list | None
     :param ext: Extension members to include in the body
     :type ext: dict | None
     :return: error response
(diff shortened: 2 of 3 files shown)


hjacobs:
Could you elaborate on why you think this change is needed?

ttung:
@rafaelcaricio asked for it in https://github.com/zalando/connexion/pull/464

connexion/lifecycle.py
```diff
@@ -34,4 +34,21 @@ def __init__(self,
         self.mimetype = mimetype
         self.content_type = content_type
         self.body = body
-        self.headers = headers or {}
+        self.headers = ConnexionHeaders(headers)
+
+
+class ConnexionHeaders(object):
```
jmcs: For backward compatibility it should be possible to use the headers object as a map/dict.

ttung: How long has this API existed?  Is it worth making it usable as a map/dict for eternity?  Was this API documented or was it internal?

I guess I'm having a hard time understanding why a shim API that's in the process of being developed needs backwards compatibility.  If I insta-crash on anyone accessing `ConnexionHeaders`, it only breaks a handful of unit tests, and those only because they call `problem()`!

FWIW, I'd have an easier time understanding that someone could pass in a map/dict as headers, and that the constructor of ConnexionHeaders should migrate that into a better representation.

ttung: IMO, `ConnexionHeaders` should implement the correct representation, which is definitely not a map of names to values.  list of tuples OR map from names to lists of values both seem acceptable.  Given the relatively short number of headers typically, lists of tuples makes the most sense.

It is the responsibility of the layer that translates to/from the underlying framework to transform the correct representation to whatever the underlying framework expects.

jmcs: We use it for [authorization](https://github.com/zalando/connexion/blob/53908d30e566ccb81d73d3e27b461f407a8a8d31/connexion/decorators/security.py#L56), for example. It's a legitimate use case if it only makes sense to have one value for a certain header, and that's why frameworks, including flask support it.

rafaelcaricio: I would vote for migrating the representation to the list of tuples in case someone pass a dict to the constructor of `ConnexionHeaders`. This is an internal API for Connexion. But I would implement here a method `get(name, default=None)`.

rafaelcaricio:
@hjacobs This will explicitly show that someone using Connexion plugged with a new framework underlining it will need to take care of header access (as well as request object access).

ttung:
Updated with a `getfirst(name, defaultvalue=None)` method.

ttung:
ping

Jyhess:
I think your class should be case insensitive to be standard compliant.
Maybe [CIMultiDict](https://aiohttp.readthedocs.io/en/v0.14.4/_modules/aiohttp/multidict.html) can be a good inspiration for header container.
My 2 cents.

RobbeSneyders:
Closing this since it's based on outdated code. Happy to review an updated PR on this.

PR #465, opened 2017-06-09
suppress DeprecationWarning -- using request.get_json() instead of request.json

podhmo:
Fixes # .

When using  connexion, deprecation warning messages are showed, such as bellow.

```
/home/me/venv/lib/python3.5/site-packages/werkzeug/local.py:343: DeprecationWarning: json is deprecated.  Use get_json() instead.
```

This is noisy, little bit.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -121,7 +121,7 @@ def parameter_to_arg(parameters, consumes, function, pythonic_params=False):
         kwargs = {}
 
         if all_json(consumes):
-            request_body = request.json
+            request_body = request.get_json()
         else:
             request_body = request.body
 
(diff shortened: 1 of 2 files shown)


jmcs:
This change could break other Frameworks besides Flask, because [`ConnexionRequest`](https://github.com/zalando/connexion/blob/63abcad51203321fbc81d2e9d93c40b6a673bcca/connexion/lifecycle.py#L2) includes a json property. Maybe it's time to wrap the Flask request

podhmo:
oh, I thought this is tiny change, but need to decision for architecture.
If so, closing this PR and creating issue is better?

rafaelcaricio:
This PR can be closed, we actually use `get_json`, look here: https://github.com/zalando/connexion/blob/63abcad51203321fbc81d2e9d93c40b6a673bcca/connexion/apis/flask_api.py#L247

Maybe there is another place where it is called again?

PR #474, opened 2017-06-28
Convert JSON/Swagger date or dateTime value to Python datetime Object

### LINKED ISSUE #451: [Question] Support for date and date-time formats
Help wanted

### Expected behaviour

```
  Scenario: Invalid date
    When I send POST request to "/myUrl":
"""
{
  "effectiveDateTime": "this is an invalid date time string, but in my spec date-time format is specified"
}
"""
    Then response code should be 400
```
### Actual behaviour

Connexion passes object to the controller.

Is there reasons why it is not implemented? 
Should I use custom type format as described here?http://connexion.readthedocs.io/en/latest/cookbook.html? 
How can I force casting into datetime object?

juliensiebert:
Fixes #451(at least trying)

 - Convert JSON/Swagger date value to Python datetime Object
 - Convert JSON/Swagger dateTime value to Python datetime Object
 -

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -1,5 +1,6 @@
 import functools
 import importlib
+from datetime import datetime
 
 
 def deep_getattr(obj, attr):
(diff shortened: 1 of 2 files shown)


connexion/utils.py
```diff
@@ -98,6 +99,28 @@ def boolean(s):
...
+
+def date_time(d):
+    '''
+    Convert JSON/Swagger dateTime value to Python datetime Object, raise ValueError otherwise
+    
+    >>> date_time("2017-01-01T10:20:30")
+    datetime.datetime(2017, 1, 1, 10, 20, 30)
+
+    '''
+    return datetime.strptime(d, "%Y-%m-%dT%H:%M:%S")
```
hjacobs: What about time zones?

hjacobs:
Please also add tests :smile:

juliensiebert:
@hjacobs Thanks for the quick answer.

I was reading again the Open API specs, and I think I am mistaking type and format for the dates. The code I wrote (without test, I do apologize) made the assumptions that the date type were either "date" or "dateTime" but after reading the specifications again, it turns out that the type is "string" and the format is "date" or "dateTime". Therefore this piece of code is obsolete (until the open api specifications change).

I'll close this PR.

Best,

julien

palmkevin:
I would really appreciate a new option, that activates automatic parsing of fields of type "string" and format "dateTime" or "date". (By default, this option should be inactive of course)

PR #477, opened 2017-06-29
Split tokeninfo scope if it is a string

### LINKED ISSUE #475: OAuth token validator is incompatible with Google's tokeninfo endpoint
Google's tokeninfo endpoint at https://www.googleapis.com/oauth2/v3/tokeninfo returns the scope as a space-separated string, not a JSON array:
```
>http https://www.googleapis.com/oauth2/v3/tokeninfo?access_token=TOKEN
HTTP/1.1 200 OK
Alt-Svc: quic=":443"; ma=2592000; v="38,37,36,35"
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Content-Encoding: gzip
Content-Type: application/json; charset=UTF-8
Date: Wed, 28 Jun 2017 22:46:05 GMT
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Server: GSE
Transfer-Encoding: chunked
Vary: Origin
Vary: X-Origin
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block

{
    "access_type": "offline",
    "aud": "ACCT.apps.googleusercontent.com",
    "azp": "ACCT.apps.googleusercontent.com",
    "email": "email",
    "email_verified": "true",
    "exp": "1498693546",
    "expires_in": "3581",
    "scope": "https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email",
    "sub": "SUB"
}

```
Giving this token to connexion:
```
DEBUG:connexion.api.security:http://localhost:5000/v1/search?query={"foo"%3A+"bar"} Oauth verification...
DEBUG:connexion.api.security:... Getting token from https://www.googleapis.com/oauth2/v3/tokeninfo
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): www.googleapis.com
DEBUG:urllib3.connectionpool:https://www.googleapis.com:443 "GET /oauth2/v3/tokeninfo?access_token=TOKEN HTTP/1.1" 200 None
DEBUG:connexion.api.security:... Token info (200): {
 "azp": "ACCT.apps.googleusercontent.com",
 "aud": "ACCT.apps.googleusercontent.com",
 "scope": "https://www.googleapis.com/auth/genomics",
 "exp": "1498691985",
 "expires_in": "2513",
 "access_type": "offline"
}

DEBUG:connexion.api.security:... Scopes required: {'foo'}
DEBUG:connexion.api.security:... User scopes: {'g', '.', 't', 'e', 'u', '/', ':', 'l', 'h', 'p', 'a', 'i', 'c', 'n', 's', 'w', 'o', 'm'}
INFO:connexion.api.security:... User scopes ({'g', '.', 't', 'e', 'u', '/', ':', 'l', 'h', 'p', 'a', 'i', 'c', 'n', 's', 'w', 'o', 'm'}) do not match the scopes necessary to call endpoint ({'foo'}). Aborting with 403.
```
Judging by the code, it expects `scope` to be a JSON array, but I don't think that's specified anywhere (and I couldn't find a compatible implementation). The closest I found to a spec is https://tools.ietf.org/html/rfc7662#section-2.2, which says `scope` is "A JSON string containing a space-separated list of scopes associated with this token".

kislyuk:
Fixes #475. 

I couldn't find a relevant test.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -71,7 +71,10 @@ def verify_oauth(token_info_url, allowed_scopes, function):
                     token_response=token_request
                 )
             token_info = token_request.json()  # type: dict
-            user_scopes = set(token_info['scope'])
+            if isinstance(token_info['scope'], list):
+                user_scopes = set(token_info['scope'])
+            else:
+                user_scopes = set(token_info['scope'].split())
             logger.debug("... Scopes required: %s", allowed_scopes)
             logger.debug("... User scopes: %s", user_scopes)
             if not allowed_scopes <= user_scopes:
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -1,10 +1,11 @@
 import json
 
+import requests
+
 import pytest
 from connexion.decorators.security import get_tokeninfo_url, verify_oauth
 from connexion.exceptions import OAuthProblem, OAuthScopeProblem
 from mock import MagicMock
-import requests
 
 
 def test_get_tokeninfo_url(monkeypatch):
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -1,7 +1,10 @@
+import json
+
 import pytest
 from connexion.decorators.security import get_tokeninfo_url, verify_oauth
-from connexion.exceptions import OAuthProblem
+from connexion.exceptions import OAuthProblem, OAuthScopeProblem
 from mock import MagicMock
+import requests
 
 
 def test_get_tokeninfo_url(monkeypatch):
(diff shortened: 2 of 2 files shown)


connexion/decorators/security.py
```diff
@@ -71,7 +71,10 @@ def wrapper(request):
                     token_response=token_request
                 )
             token_info = token_request.json()  # type: dict
-            user_scopes = set(token_info['scope'])
+            if isinstance(token_info['scope'], list):
+                user_scopes = set(token_info['scope'])
+            else:
+                user_scopes = set(token_info['scope'].split())
```
rafaelcaricio: What are you trying to do here? Maybe you want to split by `,` comma?

kislyuk: No, I'm splitting by whitespace. As I mentioned in #475, the spec says the scope items are space-separated.

kislyuk:
The current split does split by whitespace. Could you show me where/how to add a test? I couldn't find a test for the security decorator's behavior at runtime. So instead, I tested this against Google's OAuth2 API.

rafaelcaricio:
@kislyuk You can add the tests to this file [tests/decorators/test_security.py](https://github.com/zalando/connexion/blob/master/tests/decorators/test_security.py) and use [responses](https://github.com/getsentry/responses) lib to mock a call to an external server. To get the scopes. You can even add responses exactly like you are getting from Google's OAuth2 API.

kislyuk:
Would you mind sketching up a test that mocks a tokeninfo response? I'm not familiar with your test instrumentation, and the only test there now just calls an invalid tokeninfo endpoint and tests for failure.

kislyuk:
@rafaelcaricio I added a test for both string and array scopes, positive and negative, please review.

I would suggest building out the OAuth functionality to be more modular and extensible. There is very little standardization of the tokeninfo route, so a tokeninfo provider class could be more appropriate, with easily overloaded methods for getting a uid, scopes, etc. More generally, I'm interested in building out OAuth support beyond implicit flow. I can work on that in a future PR, but for now I would like to get this initial fix merged and in PyPI to unblock my team.

hjacobs:
:+1:

kislyuk:
Thank you for merging. Would it be possible to cut a PyPI release? We'd like to use this feature in our app.

hjacobs:
@kislyuk sure, I released it directly after merging (6 days ago): https://github.com/zalando/connexion/releases/tag/1.1.11 :tada:

kislyuk:
Oh, thanks! I was looking at the git history expecting to see it there.
Will try deploying with it now.

PR #478, opened 2017-06-30
Fix bug on mocking APIs partially

rafaelcaricio:
Connexion is mocking all functions even when mocking method is **not** "mock all". This PR fixes that and also remove exposition of Connexion's `ResolverError` internal exception letting go the original exception which will be much more useful for users to debug their applications.

 - Fixes mocking partially functions as expected;
 - Expose to users native errors when functions are not found. Make much easier to debug problems;
 - Log messages when functions are mocked or not. Improve development experience;
- Test under Python 3.6.

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -1,6 +1,6 @@
 import logging
-import sys
 import re
+import sys
 
 import connexion.utils as utils
 from connexion.exceptions import ResolverError
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 import re
 
 import connexion.utils as utils
@@ -45,7 +46,7 @@ class Resolver(object):
         :type operation: connexion.operation.Operation
         """
         spec = operation.operation
-        operation_id = spec.get('operationId')
+        operation_id = spec.get('operationId', '')
         x_router_controller = spec.get('x-swagger-router-controller')
         if x_router_controller is None:
             return operation_id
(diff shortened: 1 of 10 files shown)


connexion/resolver.py
```diff
@@ -45,7 +46,7 @@ def resolve_operation_id(self, operation):
         :type operation: connexion.operation.Operation
         """
         spec = operation.operation
-        operation_id = spec.get('operationId')
+        operation_id = spec.get('operationId', '')
```
rafaelcaricio: This will show error that empty module not found. Which in my opinion is better for debugging.

Removing Connexion from de way when `operationId` is not defined in the specification file. This gives a hint to developers that a module must exist in the  `operationId` field of the specification.

tests/test_api.py
```diff
@@ -53,7 +53,7 @@ def test_invalid_operation_does_stop_application_to_setup():
         FlaskApi(TEST_FOLDER / "fixtures/op_error_api/swagger.yaml",
                  base_path="/api/v1.0", arguments={'title': 'OK'})
 
-    with pytest.raises(ResolverError):
+    with pytest.raises(ValueError):
```
rafaelcaricio: Connexion internal exceptions are not exposed to users. They will see the actual errors, which is much better.

tests/test_mock.py
```diff
@@ -91,3 +95,23 @@ def test_mock_resolver_notimplemented():
                           parameter_definitions={},
                           resolver=resolver)
     assert operation.operation_id == 'fakeapi.hello.get'
+
+    # mock only the nonexistent ones
```
rafaelcaricio: This was the missing test for this feature of mocking APIs partially.

jmcs:
:+1:

rafaelcaricio:
👍

PR #479, opened 2017-07-02
Remove typo in routing.rst

holdenweb:

 Small fix to docs.


hjacobs:
I think the docs are correct and it's not a typo: the operation will be `search` for collection endpoints when using `RestyResolver`. See https://github.com/zalando/connexion/blob/d8019b346e1646be8216ffe4fc2172d28cce5185/connexion/resolver.py#L70 .

rafaelcaricio:
This should be `search` indeed.

PR #486, opened 2017-07-19
Default to all scopes checked for oauth2

WoLpH:

 - Automatically check all oauth2 scopes by default

If you're using the UI to debug it's very annoying to have to check all of the scopes every time you refresh the page. This fixes that issue and is a much more convenient default.


hjacobs:
The problem with this "monkey patching" of Swagger UI is that the change would be overwritten on next Swagger UI update.

WoLpH:
I am aware of that. I've looked at that update and it would make
significant changes anyhow without real benefit.

Personally I prefer the current release, I did get it working but the new
interface has gone down hill in my opinion.

It's only a small patch which speeds up development greatly for me.

jmcs:
I'm closing this PR since it conflicts with both #614 and #606.

PR #487, opened 2017-07-19
Validate responses against the default response field if the code is not found

ttung:
Currently, if a spec does not define a schema for a response code but defines a default schema, it is not validated.  This change pulls the default schema if the response code lookup fails.


hjacobs:
Makes sense, thank you for the PR! :smile:

hjacobs:
For reference: https://swagger.io/specification/#responses-object-54

hjacobs:
:+1:

Issue #409, opened 2017-02-23
Mock Server should build response from field's example in schema


samperd:
### Description
on connexion release 1.0.129
Running on windows
Executing from GitBash

As a Connexion user or API developer
I want to run a mock server using data listed in swagger examples
So that I can quickly validate and troubleshoot the bahaviours of the API spec.

### Expected behaviour
I would expect to get the sample data defined in the
` responses: $ref: '#/definitions/Department'`
defined as:
```
Department:
    type: object
    required:
      - legal_title
      - department_type
    properties:
      id:
        type: string
        description: System identifier
        example: "1"
        readOnly: true
      legal_title:
        type: string
        description: Department legal title
        example: "Department of the Environment"
        minLength: 1
        maxLength: 100
.
.
.
```

### Actual behaviour
when I point my browser to: 
`http://localhost:5000/departments`
I get:
`No example response was defined.`

### Steps to reproduce
When I run a mock server as:
`$ connexion.exe run goc-od-api/swagger/swagger.yaml --mock=all -v`
I get:
`INFO:werkzeug: * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)`
and when I point my browser to: 
`http://localhost:5000/departments`
I get:
`No example response was defined.`

### Additional info:

Output of the commands:

- `python --version`
```
$ python --version
Python 3.5.2
```
- `pip show connexion | grep "^Version\:"`
```
$ pip show connexion | grep "^Version\:"
Version: 1.0.129
```

rafaelcaricio:
@samperd Thank you for reporting this. But it is a known limitation of the current implementation of mock. As described in https://github.com/zalando/connexion/pull/302. Unfortunately, we never intended to have it working the way you expected.

Maybe you want to send a PR adding this support?

samperd:
OK, 

I read "https://github.com/zalando/connexion/pull/302"

Maybe I am not understanding the issue or I described the issue incorrectly.

I will try to reframe the question.

While in MOCK mode and Given a request of: 
http://localhost:5000/departments

Am I supposed to get a response of:
```
{
"id": "1",
"legal_title:"Department of the Environment"
.
.
.
}
```

I am not quite sure where I am supposed to add examples to the response body if not in:

> properties

:
```
      id:
        type: string
        description: System identifier
        example: "1"
        readOnly: true
```

I am sure I am missing something. I am just not sure what.

rafaelcaricio:
We return the "example" property. Please try like that:
[...]

PR #491, opened 2017-07-21
Fix for Issue #308: MockResolver: support "example" in definitions

### LINKED ISSUE #308: MockResolver: support "example" in definitions
Follow up to #292: we should also support returning example responses which are only defined indirectly via `example` fields in type definitions. Our current implementation only honors the `examples` property on the Response object in the Swagger spec.

gb-dsigmeth:
Fixes #308 .

Adding support for "example" in definitions to be returned from the mock server instead of "No example response was defined." for more robust mocking.

 - If response example is not provided, use "example" defined in response type schema
 - If no response example or schema example is provided return error message as before

### DIFF
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -67,8 +67,8 @@ class MockResolver(Resolver):
             # No response example, check for schema example
             response_schema = response_definition.get('schema', {})
             definitions = response_schema.get('definitions')
-            ref = response_schema.get('$ref')
-            ref = ref[ref.rfind('/')+1:]
+            ref = response_schema.get('$ref', '')
+            ref = ref[ref.rfind('/')+1:] or ''
             ref_schema = definitions.get(ref, {})
             schema_example = ref_schema.get('example')
             if schema_example:
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -66,7 +66,7 @@ class MockResolver(Resolver):
         else:
             # No response example, check for schema example
             response_schema = response_definition.get('schema', {})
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for the PR! :smile: 

It does not completely fix #308, as examples might be defined only for individual object properties, but not complete schema. Would you be willing to work towards full support for `example` properties in nested objects? (it's not trivial)

connexion/mock.py
```diff
@@ -64,4 +64,14 @@ def mock_operation(self, operation, *args, **kwargs):
         if examples:
             return list(examples.values())[0], status_code
         else:
-            return 'No example response was defined.', status_code
+            # No response example, check for schema example
+            response_schema = response_definition.get('schema', {})
+            definitions = response_schema.get('definitions', {})
+            ref = response_schema.get('$ref', '')
```
hjacobs: this would not work if the schema is defined inline

gb-dsigmeth:
#409 more specifically addresses the concept of composing an example response from individual property examples, in contrast to #308 which refers only to full-schema examples. I am in impending need of the full-schema example functionality and found it's absence quite unexpected.

Conversely #409 is more of a nice-to-have feature that I never expected, and seems more limited in it's applicability. Whether I, or someone else, adds that functionality, it can just drop in after that final else, it can come in as it's own PR which directly references the #409 issue it solves.

Long story short: I consider fixing #308 valuable enough on it's own without necessarily solving #409 also.

hjacobs:
:+1:

PR #492, opened 2017-07-21
Feature/universal wheel - fix for #488

### LINKED ISSUE #488: Unable to install connexion in environments without access to global pypi on python2.X.  Please, consider building py2 wheels
### Description
Due to current pip/setuptools behavior packages listed in setup_requires are always downloaded from global pypi, regardless of index argument passed to pip install (please, see pypa/setuptools#209) for more details. This makes it impossible to install source distribution in environments without access to global pypi on python2.7.

### Expected behaviour

Should be able to install connexion in environments that can't access global pypi on python2.7.X

### Actual behaviour

fails to install in python2.7.X because it can't reach global pypi to get flake8

### Steps to reproduce

* Download all dependencies locally (in /tmp/connexion/depts, for example)
* Disconnect from internet.  If you have a network without access to global pypi, that'll work too
* create py2.7 venv and `pip install connexion --find-packages=file:///tpm/connexion/depts -v`

The command will fail unable to find flake8
### Additional info:

Output of the commands:

- `python --version`
2.7.12
any version

### Possible solutions
* add setup.cfg to the project with the following content:
```
[bdist_wheel]
universal=1
```
this will allow release.sh to build py2py3 wheels which don't suffer from setuptools bug mentioned above

* figure which setup.py commands need ['flake8'] and do something along the lines of
```
import sys
...
need = {'flake8'}.intersection(sys.argv)
setup_requires = ['flake8'] if need else []
```

lorochka85:
Fixes #488 

 - added setup.cfg to allow building py2 wheels

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal=1


hjacobs:
Thanks for the PR! :smile:

hjacobs:
:+1:

lorochka85:
Thank you!

Issue #493, opened 2017-07-21
Missing pathlib dependency


ChamaraG:
### Description
This error occurs when building applications with connexion 1.1.12.  The `pathlib` dependency appears to be missing. 
```  
File "/work/code/sfl2/lib.cxsupport/build/tox/test/lib/python2.7/site-packages/connexion/apps/abstract.py", line 3, in <module>
    import pathlib
ImportError: No module named pathlib
```

### Additional info:

Output of the commands:

- `python --version`
    ```Python 2.7.13```

- `pip show connexion | grep "^Version\:"`
```Version: 1.1.12```

- `pip freeze | grep pathlib`
   No output

wayne-li2:
Same issue here. Use connexion 1.1.11 to workaround.

hjacobs:
Hmm, I guess the "unversal wheel" introduced with #492 broke it.

Use Python 3 as a workaround... :speak_no_evil:

ChamaraG:
Longish term, Python 3 is a solution, but that wouldn't work for us immediately ... 

Perhaps a breaking changes like this should be reserved for a major version change?

vimalloc:
Something else to keep in mind is pathlib was only introduced in python3.4.

hjacobs:
@ChamaraG actually the breaking change was introduced with the unversal wheel which was requested by a Python 2 user. See #492 

I'm not sure how to proceed here, simple workaround would be to add `pathlib` dependency in your own requirements. Or we simply revert the "universal wheel".

Any suggestions?

underyx:
@lorochka85 do you have any idea, as the author of the universal wheel PR?

underyx:
@hjacobs instead of using code in `setup.py`, couldn't you just use environment markers to conditionally install the compat packages? Like this: `'pathlib2; python_version < "3.4"'`

See more info about these here: https://www.python.org/dev/peps/pep-0508/#environment-markers

hjacobs:
@underyx ah, I did not know about this. Would you be willing to do a PR?

lorochka85:
Seems like environment-markers is the way to go - nice find @underyx!   I'm curious though, was it a problem with *.tar installations on py2.7 or was only introduced with universal wheel?

underyx:
>Would you be willing to do a PR?

Sure thing @hjacobs: https://github.com/zalando/connexion/issues/495

ChamaraG:
[...]

PR #495, opened 2017-07-27
setup.py: Use env markers for Python version-specific dependencies

underyx:
Changing package properties dynamically in setup.py means that universal
wheels will actually be built only for the builder's environment, and
might not work properly on different platforms/PYthon versions.

Closes https://github.com/zalando/connexion/issues/493

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.5"
   - "3.6"
 install:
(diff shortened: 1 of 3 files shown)


underyx:
…aaand it's green 😮 

finally, can't even believe it :D

hjacobs:
:+1:

PR #497, opened 2017-08-02
Fix KeyError if no Content-Type was provided in request

f11r:
At the moment if no `Content-Type` is provided in the request the validation wrapper raises a `KeyError`.

Can we expect a fix to make it into a release soon or should we look for a temporary workaround until that happens?

### DIFF
--- a/tests/api/test_responses.py
+++ b/tests/api/test_responses.py
@@ -1,6 +1,8 @@
 import json
 from struct import unpack
 
+from werkzeug.test import Client, EnvironBuilder
+
 from connexion.apps.flask_app import FlaskJSONEncoder
 
 
@@ -219,7 +221,6 @@ def test_post_wrong_content_type(simple_app):
     # (https://github.com/pallets/werkzeug/issues/1159)
     # so that content-type is added to every request, we remove it here manually for our test
     # this test can be removed once the werkzeug issue is addressed
-    from werkzeug.test import Client, EnvironBuilder
     builder = EnvironBuilder(path='/v1.0/post_wrong_content_type', method='POST',
                              data=json.dumps({"some": "data"}))
     try:
--- a/tests/api/test_responses.py
+++ b/tests/api/test_responses.py
@@ -219,7 +219,7 @@ def test_post_wrong_content_type(simple_app):
     # (https://github.com/pallets/werkzeug/issues/1159)
     # so that content-type is added to every request, we remove it here manually for our test
     # this test can be removed once the werkzeug issue is addressed
-    from werkzeug.test import EnvironBuilder
+    from werkzeug.test import Client, EnvironBuilder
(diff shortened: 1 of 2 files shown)


hjacobs:
@f11r looks good, thanks for the PR! Could you add a test for this case please? Thanks!

hjacobs:
@f11r answering your question: yes, we can quickly release it, but please add a test :smile:

f11r:
@hjacobs `tests/decorators/test_validation.py` seems to only have tests for `ParameterValidator. Is this still the correct file to add a test or could you point me to the correct place to add it? I'm not really familiar with the tests here.

hjacobs:
@f11r I think you can add it here: https://github.com/zalando/connexion/blob/6668835420231c15095a71e264b2e2b0e1a68bf3/tests/api/test_responses.py#L203

It currently only tests the status code (415) on "wrong" content type (i.e. not the case where `Content-Type` is missing).

f11r:
It actually looks like you already did have the correct test here https://github.com/zalando/connexion/blob/6668835420231c15095a71e264b2e2b0e1a68bf3/tests/api/test_responses.py#L213, but werkzeug always sets `Content-Type` to an empty string if not provided in the `post`. I opened an issue (https://github.com/pallets/werkzeug/issues/1159) over there.

hjacobs:
@f11r maybe you can do an "artificial" manual HTTP request in the test and not use the test client?

f11r:
Added a test, not particularly pretty. Not sure if something like `requests.post()` could actually work since if I understand it correctly the test server is just a "fake"-server.

hjacobs:
Hmm, test is failing now: https://travis-ci.org/zalando/connexion/jobs/260553758

f11r:
Figured out what is going on: `app_client` is a `flask.testing.FlaskClient` which does not support passing an environment as the only argument to `open()` in `flask<=0.12.2`. This was only implemented recently in https://github.com/pallets/flask/pull/2412. That's why it succeeded for the development builds but not for the "real" builds. I implemented a workaround but that doesn't make the test any prettier.

tests/api/test_responses.py
```diff
@@ -213,6 +213,26 @@ def test_post_wrong_content_type(simple_app):
                            )
     assert resp.status_code == 415
 
+    # this test checks exactly what the test directly above is supposed to check,
+    # i.e. no content-type is provided in the header
+    # unfortunately there is an issue with the werkzeug test environment
+    # (https://github.com/pallets/werkzeug/issues/1159)
+    # so that content-type is added to every request, we remove it here manually for our test
+    # this test can be removed once the werkzeug issue is addressed
+    from werkzeug.test import Client, EnvironBuilder
```
jmcs: Can you move this import to the top level (beginning of the file)?

f11r: done

jmcs:
:+1:

PR #499, opened 2017-08-02
Updated parameter.py to accept list parameters

### LINKED ISSUE #323: Support for type array with collectionFormat multi
### Description

According to the OpenAPI specification for the collectionFormat, there is an additional type **multi** to allow multiple parameter instances instead of multiple values for a single instance.
### Expected bahaviour

Connexion to support the collectionFormat of type **multi**.
### Actual behaviour

Currently, Connexion does only support the types of **csv**, **tsv** and **pipes**.
### Additional info:
- Python 3.5.2
- connexion Version: 1.0.109

lhannest:
Fixes #323 

 - On line 153 of parameter.py: Adding the items in `requests.query` to `query_arguments` in such a way to preserve the multiplicity of values with the same key.
 - On line 58 of parameter.py: Adding a condition to the if-then-else block to handle when `value` is a list.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -55,7 +55,12 @@ def get_val_from_param(value, query_param):
     if is_nullable(query_param) and is_null(value):
         return None
 
-    if query_param["type"] == "array":  # then logic is more complex
+    if isinstance(value, list):
+        if query_param["type"] != "array":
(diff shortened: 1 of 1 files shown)


lhannest:
Tests failed, I re-implemented this and put in another pull request here: https://github.com/zalando/connexion/pull/500

PR #500, opened 2017-08-02
Preserving the multiplicity of array type query parameters

lhannest:
Fixes issue #323

This is a better implementation of what I was doing in the #499 pull request.

 - On line 146 of parameter.py: updating the request query parameters being added to `query_arguments` so as to preserve lists that are encoded by having multiple values to a single key.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -101,8 +101,7 @@ def parameter_to_arg(parameters, consumes, function, pythonic_params=False):
 
     def make_request_query(request):
         request_query = {}
-        isMultiDict = callable(getattr(request.query, "to_dict", None))
-        if isMultiDict:
+        try:
             for k, v in request.query.to_dict(flat=False).items():
                 k = sanitize_param(k)
                 query_param = query_types.get(k, None)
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for the PR! Could you check the failing test (see https://travis-ci.org/zalando/connexion/jobs/260404325) and add a test for the array parameter?

lhannest:
This PR isn't really a fix for #323. It's really just a partial solution that will allow users to get the functionality of multi formatted arrays. But I have not done anything to support a multi collection format in the swagger.yaml specification file.

Before, only the first of multiple keys would be accepted. Now they will all be accepted. So doing `/api?items=A,B,C&items=D` will be the same as doing `items=A,B,C,D` as long as the collection format is csv. On the other hand, if the collection format is pipes, then `/api?items=A,B,C&items=D` will be the same as doing `/api?items=A,B,C|D`.

lhannest:
@hjacobs do you have any other thoughts?

lhannest:
It's been two weeks, can the "waiting for contributor feedback" label be removed at least so someone eventually gets around to looking at this PR?

hjacobs:
@lhannest sorry for the the slow response, @jmcs can you take a look?

lhannest:
@mvalkon Thank you for the review, and sorry for taking a while to respond. I pushed the changes you suggested.

mvalkon:
@lhannest lgtm, thanks for the contribution!

mvalkon:
👍

hjacobs:
:+1:

PR #504, opened 2017-08-08
Allow http.HTTPStatus enums as response status codes.

### LINKED ISSUE #503: Connexion does not support use of http.HTTPStatus enums as response codes.
### Description
Python 3.5 introduced a new enum type for representing HTTP response status codes: "http.HTTPStatus". Enums are generally int like enough to work transparently, however, the response validation attempts to convert the status_code to a string for comparison, a place where enum is insufficiently int-like.

### Expected behaviour
Enums like http.HTTPStatus.OK should be usable as response status codes from view functions.

### Actual behaviour
With the introduction of default response checking in connexion 1.1.12, HTTPStatus enums now trigger the default response validation instead of being ignored, as was true in previous versions.

### Steps to reproduce
return {}, http.HTTPStatus.OK

### Additional info:

Output of the commands:

- `python --version`
python 3.5.3
Version: 1.1.13

jkrukoff:
Python 3.5 introduced a new enumeration "http.HTTPStatus" for
representing HTTP response status codes. The default response validation
introduced in connexion 1.1.12 highlighted the fact that connexion does
not natively support this type and was previously silently ignoring
non-integer status code representations.

This modifies the response validation code to extract the value when
given an enum instead of an int. Somewhat hacky test code is added to
check for enum support on python versions that include
"http.HTTPStatus".

Fixes #503 

 - Convert enums to ints in response validation matching.

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -154,6 +154,12 @@ class FlaskApi(AbstractAPI):
         flask_response = flask.current_app.response_class(**kwargs)  # type: flask.Response
 
         if status_code is not None:
+            try:
+                # If we got an enum instead of an int, extract the value.
+                status_code = status_code.value
+            except AttributeError:
+                pass
+
             flask_response.status_code = status_code
 
         if data is not None and data is not NoContent:
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -154,11 +154,9 @@ class FlaskApi(AbstractAPI):
         flask_response = flask.current_app.response_class(**kwargs)  # type: flask.Response
 
         if status_code is not None:
-            try:
-                # If we got an enum instead of an int, extract the value.
+            # If we got an enum instead of an int, extract the value.
+            if hasattr(status_code, "value"):
                 status_code = status_code.value
-            except AttributeError:
-                pass
 
             flask_response.status_code = status_code
 
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -1,7 +1,4 @@
 #!/usr/bin/env python3
-import sys
-import unittest
-
 import flask
 from flask import jsonify, redirect
 
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+import sys
+import unittest
 
 import flask
 from flask import jsonify, redirect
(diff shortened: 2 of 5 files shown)


jmcs:
I like it, but can you had a unit test for custom enum (this would cover the case where someone uses a backport of the enum).

jkrukoff:
Do you have a suggestion on a good way to only run the test in python versions where Enum is available, or should I use the same ImportError based approach?

mvalkon:
@jkrukoff there's a way to do this yes. First you need to indicate that the test should be skipped in python versions that do not support enums. You can use the [@unittest.skipIf][1] decorator:

```python
@unittest.skipIf(sys.version_info < (3, 5), "Not supported in this version")
def test_your_test_name(...):
```

If you want to run your tests with multiple python versions, we're using [tox][2] for this. To verify that the above works you can run tox with `-e` command line parameter as follows:

```bash
$ tox -e py27,py34,py36  
```

You can install tox with `pip`:

```bash
$ pip install tox
```
tox is currently not in https://github.com/zalando/connexion/blob/master/requirements-devel.txt but I'll make a PR to add it.

[1]: https://docs.python.org/3/library/unittest.html#unittest.skipIf
[2]: https://tox.readthedocs.io/en/latest/

jkrukoff:
Added the test for arbitrary enums, which exposed an assumption inside flask that appears to only work for ints/actual HTTPStatus Enums. Moved the enum value extraction deeper into the stack to compensate.

Tried to use @unittest.skipif in the test suite, but that failed with an unsupported error from unittest for python versions <3.5. Reverted back to the previous approach to get tests passing again.

mvalkon:
@jkrukoff `unittest.skipIf` has been introduced in 3.1. Out of interest, what sort of an exception did you get?

The change looks good to me.

jkrukoff:
It was pretty close to "unittest.skipif is unsupported in this version". It it's been present since 3.1, would explain why the python 2.7 test suite was exploding with that error.

tests/fakeapi/hello.py
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+import sys
+import unittest
```
jmcs: This imports are not being used.

connexion/apis/flask_api.py
```diff
@@ -154,6 +154,12 @@ def _build_flask_response(cls, mimetype=None, content_type=None,
         flask_response = flask.current_app.response_class(**kwargs)  # type: flask.Response
 
         if status_code is not None:
+            try:
```
jmcs: Please use ``hasattr`` here since the most common use case is ``status_code`` being an integer.

[``hasattr`` is faster than ``try... except AtributeError`` for cases where the attribute doesn't exist](https://stackoverflow.com/a/903238).

tests/fakeapi/hello.py
jkrukoff: Ah, yeah, forgot to remove those from the attempt to use uniftest.skipIf.

connexion/apis/flask_api.py
jkrukoff: Yes, and you get to repeat yourself while doing the check. It'd be even faster if just didn't support enums at all!

Will update.

hjacobs:
:+1:

PR #506, opened 2017-08-16
Fix the examples

mvalkon:
Some of the [examples](https://github.com/zalando/connexion/tree/master/examples/) were raising an `Exception` when ran according to the provided `README` documentation. This occurred because the port was passed incorrectly as the `server`-argument. Fixed also the exception string which had incorrect string formatting.


rafaelcaricio:
@mvalkon Thank you for taking care of that! 😃

rafaelcaricio:
👍

hjacobs:
@mvalkon thanks a lot :smile:

PR #513, opened 2017-09-20
Generate example recursively in mock server

### LINKED ISSUE #409: Mock Server should build response from field's example in schema

jadkik:
Fixes #409.

Adds support in mock server for generating examples based on nested schemas.

Based on [the implementation of swagger-ui](https://github.com/swagger-api/swagger-ui/blob/v3.2.1/src/core/plugins/samples/fn.js) with some modifications.

 - Follow example precedence as defined in [the specification](https://swagger.io/docs/specification/2-0/adding-examples/)
 - Basic generation of example values for primitives (using `type`, `format`, `minimum`, `maximum`, etc.)
 - Basic generation of object/array examples (using `properties`, `additionalProperties`, `items`)

Things that are still missing and/or can be improved:

  * Support for polymorphism using allOf
  * More validation/format support (e.g. mutipleOf, pattern, length...)
  * Date/time support (without now/today)
  * Response headers
  * Support for nullable fields
  * Randomization (with some custom seed)
  * Support for external references

Any help with the testing is welcome.

### DIFF
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -1,8 +1,126 @@
+import datetime
 import logging
 
+from connexion import NoContent
 from connexion.resolver import Resolution, Resolver, ResolverError
 
 logger = logging.getLogger(__name__)
+
+PRIMITIVES = {
+  "string": lambda schema: "string",
+  "string_email": lambda schema: "user@example.com",
+  "string_date": lambda schema: datetime.date.today().isoformat(),
+  "string_date-time": lambda schema: datetime.datetime.now().isoformat(),
+  "number": lambda schema: numeric_primitive(schema, int),
+  "number_float": lambda schema: numeric_primitive(schema, float),
+  "integer": lambda schema: numeric_primitive(schema, int),
+  "boolean": lambda schema: schema.get('default', True),
+}
+
+
+def numeric_primitive(schema, type_):
+    minimum = schema.get('minimum')
+    exclusive_minimum = schema.get('exclusiveMinimum', False)
+    maximum = schema.get('maximum')
+    exclusive_maximum = schema.get('exclusiveMaximum', False)
+    default = schema.get('default')
+
+    if default is not None:
+        return default
+
+    boundary = 0.01 if type_ is float else 1
+    minimum = minimum + boundary if exclusive_minimum else minimum
+    maximum = maximum + boundary if exclusive_maximum else maximum
+
+    if minimum is not None and maximum is not None:
+        if type_ is int:
+            return type_(minimum + maximum) // 2
+        else:
+            return type_(minimum + maximum) / 2.0
+    elif minimum is not None:
+        return type_(minimum + 1)
+    elif maximum is not None:
+        return type_(maximum - 1)
+    else:
+        return type_()
+
+
+def primitive(schema):
+    kw = dict(type=schema['type'], format=schema.get('format', ''))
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for the PR! :smile:
Could you add some tests?

jadkik:
Does it make sense for it to return `NoContent` if a schema is not defined? Or would you rather let it return `'No example response was defined.'`?

I believe it would make more sense to return `NoContent` since this is what [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#response-object) recommends (emphasis mine):

> A definition of the response structure. It can be a primitive, an array or an object. **If this field does not exist, it means no content is returned as part of the response.** As an extension to the Schema Object, its root type value may also be "file". This SHOULD be accompanied by a relevant produces mime-type.

Also, I was considering using their petstore example as a reference, but it seems they don't abide by the spec themselves, specifically regarding this:

> It is not expected from the documentation to necessarily cover all possible HTTP response codes, since they may not be known in advance. However, **it is expected from the documentation to cover a successful operation response** and any known errors.

However, in some cases in [that file](http://petstore.swagger.io/v2/swagger.json), they only define error responses, so the mock server is returning error codes in those cases.

In light of this, I suggest:

  * returning a status code of `418` with `NoContent` when there is no "success" 2xx response defined; since this goes against the spec (unless there are legitimate use cases for API endpoints that always return error codes?)

  * in case the response is defined with no schema return the correct status code with `NoContent` since this is mentioned in the specification directly.

Does that make sense?

jadkik:
Well it seems I've based the implementation on the 3.x branch of swagger-ui. The implementation for the 2.x branch is [here](https://github.com/swagger-api/swagger-js/blob/2.x/lib/schema-markup.js) or [here](https://github.com/swagger-api/swagger-ui/blob/v2.2.5/src/main/javascript/view/partials/signature.js).

The only differences that jump out to me in the implementation are the handling of the `items` for the `array` type, the support for `writeOnly` fields, and "recursive" nesting of schemas.

JustinFeng:
Hey guys, any update about this, we need the same feature as well

RobbeSneyders:
Closing since this was added in #1148.

PR #523, opened 2017-10-05
FIX a typo causing an ERROR log

### LINKED ISSUE #522: Erroneous warnings produced since #500 (Function argument not defined in specification)
### Description
It looks like #500 introduced a bug which reports that the function arguments weren't defined when they in fact were.

### Expected behaviour
Function arguments with proper definitions should not produce warnings.

### Actual behaviour
Function arguments with proper definitions do produce warnings.

### Steps to reproduce
Consider the following swagger spec:
```yaml
swagger: "2.0"
basePath: "/v1"
schemes:
- "https"
paths:
  /photo:
    post:
      summary: "Upload a photo along with its metadata"
      description: ""
      operationId: "upload.upload_photo"
      consumes:
      - "multipart/form-data"
      produces:
      - "application/json"
      parameters:
      - name: "photo"
        required: true
        in: "formData"
        type: "file"
        description: "The photo file"
      - name: "imageId"
        required: true
        in: "formData"
        type: "string"
        description: "The UUID assigned to this photo"
      - name: "associations"
        required: true
        in: "formData"
        type: "array"
        collectionFormat: "csv"
        items:
          type: "string"
        description: "A CSV list of QR IDs associated with the photo"
      - name: "takenAt"
        required: true
        in: "formData"
        type: "string"
        description: "ISO 8601 compliant datetime stamp representing when the photo was taken"
      - name: "latitude"
        in: "formData"
        type: "number"
        description: "Latitude portion of where the photo was taken"
      - name: "longitude"
        in: "formData"
        type: "number"
        description: "Longitude portion of where the photo was taken"
      responses:
        201:
          description: "Photo was uploaded and metadata stored successfully"
        409:
          description: "Photo with that ID already exists"
```

and the following Python code:

```python
import connexion

def upload_photo(photo, imageId, associations, takenAt, latitude, longitude):
    print(locals())

app = connexion.App(__name__, specification_dir="spec")
app.add_api("v1.yml")
app.run(port=8080)
```

When running on version `1.1.14` the command line output simply lists the `locals()` output and a POST:

```
 * Running on http://0.0.0.0:8080/ (Press CTRL+C to quit)
{'latitude': 34234.32432434, 'associations': ['aaaa%2Cbbbb%2Ccccc'], 'takenAt': 'dsfdsfdsfsdffdsf', 'longitude': 324324.4444444, 'photo': <FileStorage: 'someimage.png' ('image/png')>, 'imageId': 'abcdef12345678'}
127.0.0.1 - - [04/Oct/2017 14:35:36] "POST /v1/photo HTTP/1.1" 200 -
```

However, when running on version `1.1.15`:

```
 * Running on http://0.0.0.0:8080/ (Press CTRL+C to quit)
Function argument 'imageId' not defined in specification
Function argument 'latitude' not defined in specification
Function argument 'longitude' not defined in specification
Function argument 'takenAt' not defined in specification
Function argument 'associations' not defined in specification
{'longitude': 324324.4444444, 'takenAt': 'dsfdsfdsfsdffdsf', 'associations': ['aaaa%2Cbbbb%2Ccccc'], 'imageId': 'abcdef12345678', 'latitude': 34234.32432434, 'photo': <FileStorage: 'someimage.png' ('image/png')>}
127.0.0.1 - - [04/Oct/2017 14:36:19] "POST /v1/photo HTTP/1.1" 200 -
```

### Additional info:

Output of the commands:

- `python --version` - `Python 3.5.2`
- `pip show connexion | grep "^Version\:"` - `Version 1.1.15`

As a side note - the `associations` parameter doesn't seem to get split properly. Am I doing something wrong?

mvalkon:
A small bug was introduced in #500 when sanitizing the request query
parameters. Instead of fetching the arguments from `request.query`,
the parameters were sanitized from the `request.form`. This causes an
error log to be printed, as the parameters are not expected for the
query (for example in the case of a POST request).

Fixes #522

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ tests_require = [
     'mock',
     'pytest',
     'pytest-cov',
+    'testfixtures',
     flask_require
 ]
 
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -119,7 +119,7 @@ def parameter_to_arg(parameters, consumes, function, pythonic_params=False):
                 else:
                     request_query[k] = v[0]
         except AttributeError:
-            request_query = {sanitize_param(k): v for k, v in request.form.items()}
+            request_query = {sanitize_param(k): v for k, v in request.query.items()}
         return request_query
 
     body_parameters = [parameter for parameter in parameters if parameter['in'] == 'body'] or [{}]
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -6,6 +6,7 @@
 inflection
 mock
 pytest
+testfixtures>=5.3.0
 # This repo doesn't have the latest version released on PyPI
 # -e hg+https://bitbucket.org/pitrou/pathlib#egg=pathlib
 # PyYAML is not that easy to build manually, it may fail.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -142,3 +142,8 @@ def unordered_definition_app():
 @pytest.fixture(scope="session")
 def bad_operations_app():
     return build_app_from_fixture('bad_operations', resolver_error=501)
+
+
+@pytest.fixture(scope="session")
+def query_sanitazion():
+    return build_app_from_fixture('query_sanitazion')
--- a/tests/decorators/test_parameter.py
+++ b/tests/decorators/test_parameter.py
@@ -1,6 +1,8 @@
+
 from connexion.decorators.parameter import parameter_to_arg
 # we are using "mock" module here for Py 2.7 support
 from mock import MagicMock
+from testfixtures import LogCapture
 
 
 def test_injection(monkeypatch):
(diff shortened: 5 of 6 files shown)


hjacobs:
@mvalkon can you please add a test for this? Thanks!

mvalkon:
@hjacobs yup, I was working on that already but then realised this one is hard to test with the current test functions. I might have to add a test for the `parameter.wrapper`-function which is a bit more low level than what the tests currently have.

mvalkon:
@hjacobs @flyte finally had a chance to look at the test, sorry for taking such a long time with this simple fix.

The problem I have is that the only way I could think of to actually test this is to assert that certain logging messages are either present, or not. For example, in the case of #522, we could either:
1. Assert that the correct messages get logged for `formData`-parameters
2. Assert that the messages for query parameters in case of `formData` are not logged

I can do either one (although the second one is far more ugly), but this means I'll have to introduce a backwards compatible library for testing logging messages ([testfixtures][1] for example). Do we think it's worth it for _just_ this one test?

[1]: http://testfixtures.readthedocs.io/en/latest/logging.html

mvalkon:
@hjacobs I've added the test described. Let me know if you think this should be added.

setup.py
```diff
@@ -30,6 +30,7 @@ def read_version(package):
     'inflection>=0.3.1',
     'pathlib>=1.0.1; python_version < "3.4"',
     'typing>=3.6.1; python_version < "3.6"',
+    'testfixtures>=5.3.0'
```
hjacobs: this will pull in `testfixtures` everywhere, but we only need it for tests

mvalkon: Yup, sorry about that. Added it in the `requirements-devel.txt`

mvalkon:
@hjacobs ping

hjacobs:
:+1:

PR #529, opened 2017-10-12
MINOR Readme Fix: Rename "search" to "get" in main Readme for RestyResolver example

pheanex:
If i am not mistaken, "get" should be correct here?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -191,7 +191,7 @@ the endpoints in your specification:
           # Implied operationId: api.get
      /foo:
        get:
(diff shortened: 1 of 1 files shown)


hjacobs:
No, the README is correct. For RestyResolver, the "get" function is used the individual item `/foo/{id}` and search is for getting the collection (`/foo`)

pheanex:
Ah yes. I just noticed it in the sources. Sorry for the confusion :-)

PR #530, opened 2017-10-19
Added aiohttp support

dutradda:
Hello Guys, I implemented an App and Api classes supporting aiohttp framework. It is fully tested.
I had did some modifications on the base code as well.
Please, take a look on it


hjacobs:
Oh, wow, thanks!

connexion/__init__.py
```diff
@@ -26,5 +27,21 @@ def _required_lib(exec_info, *args, **kwargs):
 App = FlaskApp
 Api = FlaskApi
 
+if sys.version_info[0:2] >= (3, 4): # pragma: no cover
+    try:
+        from .apis.aiohttp_api import AioHttpApi
```
prawn-cake: The code above does almost exactly the same, except import paths. Let's refactor it to not repeat yourself.
```
try:
    from .apis.flask_api import FlaskApi
    from .apps.flask_app import FlaskApp
    from flask import request  # NOQA
except ImportError as e:  # pragma: no cover
    import six
    import functools

    def _required_lib(exec_info, *args, **kwargs):
        six.reraise(*exec_info)

    _flask_not_installed_error = functools.partial(_required_lib, sys.exc_info())

    FlaskApi = _flask_not_installed_error
    FlaskApp = _flask_not_installed_error

...

if sys.version_info[0:2] >= (3, 4): # pragma: no cover
    try:
        from .apis.aiohttp_api import AioHttpApi
        from .apps.aiohttp_app import AioHttpApp
    except ImportError as e:
        import six
        import functools

        def _required_lib(exec_info, *args, **kwargs):
            six.reraise(*exec_info)

        _aiohttp_not_installed_error = functools.partial(_required_lib, sys.exc_info())

        AioHttpApi = _aiohttp_not_installed_error
        AioHttpApp = _aiohttp_not_installed_error
```

I see it as
```
# We can make this as a high-order function and reuse
def not_installed_error():
    import six
    import functools

    def _required_lib(exec_info, *args, **kwargs):        
        six.reraise(*exec_info)

    return functools.partial(_required_lib, sys.exc_info())

...

try:
...
except ImportError:
    WhateverApi = not_installed_error()
```

This will make it more readable

prawn-cake:
General note: the PR lacks some docs about how to plug-in aiohttp server backend, plus `cli.py` need to be updated to support it as well

connexion/apis/aiohttp_api.py
```diff
@@ -0,0 +1,266 @@
...
+from connexion.lifecycle import ConnexionRequest, ConnexionResponse
+from connexion.utils import is_json_mimetype
+
+try:
+    import ujson
+    from functools import partial
+    ujson.dumps = partial(ujson.dumps, escape_forward_slashes=True)
+
+except ImportError:  # pragma: no cover
+    import json as ujson
```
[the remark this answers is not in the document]
hjacobs: I agree!

dutradda:
@prawn-cake the CLI was already updated. see [here](https://github.com/zalando/connexion/pull/530/files#diff-6c18a0256d3c90694845677efd54e80e)

I don't known what you means with "how to plug-in aiohttp server backend".
It is how to implement aiohttp handlers? Or is how to run a aiohttp server? Because the [AioHttpApp.run](https://github.com/zalando/connexion/pull/530/files#diff-b77ad459497243b2bf19a278475d19d1R46) method run the aiohttp server. What exact I must to document better?

And about your comments I will update the PR

dutradda:
@prawn-cake 
I did the requested changes.
About the cli I did one option as deprecated, I don't if what I did is the best solution.
And about the docs what I did is sufficient?

connexion/cli.py
```diff
@@ -1,6 +1,7 @@
 import importlib
 import logging
 import sys
+from itertools import chain
```
prawn-cake: Redundant import?

connexion/cli.py
```diff
@@ -108,6 +133,23 @@ def run(spec_file,
 
     - BASE_MODULE_PATH (optional): filesystem path where the API endpoints handlers are going to be imported from.
     """
+    if wsgi_server and server:
+        raise click.BadParameter(
+            "these options are mutually excludent",
```
prawn-cake: Minor typo, I guess "mutually exclusive" is the proper one =)

connexion/cli.py
```diff
@@ -11,9 +12,25 @@
 
 logger = logging.getLogger('connexion.cli')
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
-
-
-def validate_wsgi_server_requirements(ctx, param, value):
+FLASK_APP = 'flask'
```
prawn-cake: Not sure if I understood purpose of these constants properly.
* We divide our config into **connexion apps**, **app frameworks** and **app servers** we serve our web-app with, right?
* Each app framework can be an app server at the same time, i.e serve itself
* We want to resolve dependencies between app servers and app frameworks (e.g flask can't be run with aiohttp by default although it has 3rd party [wsgi](https://github.com/etianen/aiohttp-wsgi) adapters)

Just trying to understand why we need so many configs which repeat the same names

dutradda: I don't know if I understand your points, but I will try to explain why I created these variables.

I created the AVAILABLE_SERVERS variable to ensure that the user input is right (but this validation will be done by the app when the App.run method is called). So this validation can be removed as well.

The AVAILABLE_APPS is for mapping the "--app-framework" input with the right class.

The DEFAULT_SERVERS was created because some frameworks has no default server (like the falcon framework, there's no way to run the application directly from falcon, a external wsgi server must be used), but for now the defaults are the same name of the framework, so I can remove this variable too if you want.

connexion/cli.py
dutradda: I already remove this line, this code is outdated

connexion/cli.py
dutradda: I will fix my typo :)

connexion/cli.py
```diff
@@ -10,19 +10,37 @@
...
+    AIOHTTP_APP: AIOHTTP_APP
+}
+
+
+def validate_server_requirements(ctx, param, value):
     if value == 'gevent':
         try:
             import gevent  # NOQA
-        except:
+        except Exception:
```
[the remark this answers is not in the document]
prawn-cake: This one was resolved with [this commit](https://github.com/zalando/connexion/commit/a275f7b0e20d896bedf956948ca4376ad8ef3923). Pull latest changes, please.

prawn-cake:
Also found aiohttp level issue with the `base_path` when it adds a subapp.

So basically this will fail
```python
app = connexion.AioHttpApp(__name__)
app.add_api('swagger.yaml', base_path='/')
app.run(port=8080)
```

with
```
  File "/home/maxi/myPython/tryConnexion/app.py", line 53, in <module>
    app.add_api('swagger.yaml', base_path='/')
  File "/home/maxi/myPython/connexion/connexion/apps/aiohttp_app.py", line 43, in add_api
    self.app.add_subapp(api.base_path, api.subapp)
  File "/home/maxi/myPython/tryConnexion/.env/lib/python3.5/site-packages/aiohttp/web.py", line 186, in add_subapp
    raise ValueError("Prefix cannot be empty")
ValueError: Prefix cannot be empty
```

prawn-cake:
I also created [simple project](https://github.com/prawn-cake/tryConnexion) to test your changes and try to use aiohttp-backed connexion service.
I had to struggle a bit with the handlers, took me some time to debug errors. Initially I expected that it would work transparently if I switch the app and turn my handlers into coroutines, not really. Errors it spits out are not obvious at all.

Thought of may be we could extend our hello world example with the proper aiohttp handler or at least make a note and point to the [aiohttp handlers documentation](https://docs.aiohttp.org/en/stable/web.html#handler).

connexion/cli.py
```diff
@@ -104,6 +131,29 @@ def run(spec_file,
...
+    if server is None:
+        server = DEFAULT_SERVERS[app_framework]
+
+    if app_framework not in AVAILABLE_SERVERS[server]:
+        message = "Invalid server '{}' for app-framework '{}'".format(
+            server, app_framework
+        )
+        raise click.UsageError(message)
+
+    if app_framework == AIOHTTP_APP:
```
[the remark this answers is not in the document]
dutradda: I did this check here because if a user input is "--app-framework=flask --server=aiohttp" the error that will raise is "invalid server for 'flask'".

If the check is done on "validate_server_requirements" the error that will raise is "aiohttp not found".

But I can move this check to the "validate_server_requirements" function if you prefer

dutradda:
@prawn-cake 
you suggest some modification to the base_path issue?


and about the docs issues
where are the hello world example that I can modify?

README.rst
```diff
@@ -413,6 +413,15 @@ Flask with uWSGI`_ (this is common):
...
 
+You can use the ``aiohttp`` framework as server backend as well:
+
+.. code-block:: python
+
+    import connexion
+
+    app = connexion.AioHttpApp(__name__, specification_dir='swagger/')
+    app.run(port=8080)
+
```
prawn-cake: Here we could add something like
**NOTE:** Also check aiohttp handler [examples](https://docs.aiohttp.org/en/stable/web.html#handler)

prawn-cake:
Regarding the base_path. At least we could intercept that exception and re-raise it with the more meaningul message in the `AioHttpApp.add_api` method. 
For example "_`aiohttp` doesn't allow to set empty base_path ({base_path}), use non-empty instead, e.g /api"_  

Added comment in the README

dutradda:
@prawn-cake, I think I did the requested changes. Can you take a look?

tests/fakeapi/aiohttp_hello.py
```diff
@@ -0,0 +1,69 @@
...
+@asyncio.coroutine
+def aiohttp_multiple_methods_get(name, **kwargs):
+    data = {'method': 'get'}
+    return ConnexionResponse(
+        body=data
+    )
+
+
+@asyncio.coroutine
+def aiohttp_multiple_methods_post(name, **kwargs):
```
prawn-cake: The tests with payload should also check the payload. Instead of passing a name as a path variable better to pass it as a json payload and extract it here.
I'm trying to debug it, because it doesn't work for my [demo app](https://github.com/prawn-cake/tryConnexion).
```
@asyncio.coroutine
def test_connexion(test_client):
    app = connexion.AioHttpApp(__name__)
    app.add_api('swagger.yaml', base_path='/api')
    aiohttp_app = app.app
    client = yield from test_client(aiohttp_app)

    routes = aiohttp_app.router.routes()

    resp = yield from client.get('/1')
    assert resp.status == 404

    resp = yield from client.get('/api/pets')
    json_data = yield from resp.json()
    assert json_data == []

    pet = {"id": 1, "name": "Tosi", "animal_type": "cat"}
    resp = yield from client.put('/api/pets/1', json=pet)
```

It returns some weird errors like:
```
connexion/decorators/validation.py", line 109, in wrapper
if data is None and len(request.body) > 0 and not self.is_null_value_valid:
TypeError: object of type 'FlowControlStreamReader' has no len()
```

Then I discovered that this functionality is not tested in your PR. Please, check it as well.
If I manage to debug it, I will send a PR to your branch

prawn-cake: OK. So the problem is request body is a stream and due to its' async nature you have to await for it and it should happen when you create `ConnexionRequest`. 

Currently if fails during request body validation.
The hard part is that the validation process is hapenning in the middleware which is a coroutine for aiohttp api which make validation code incompatible.

I'm still investigating into it. Just wanna give you outline of the problem

dutradda: @prawn-cake yes, I saw this problem too. I know how to fix it. In the response validation happens the same situation and I solve this with extra module import. I thing I can do something like this on body validation too.

prawn-cake: @dutradda 
I sent a [PR](https://github.com/dutradda/connexion/pull/1) to your branch. It doesn't fix the problem here, but I cleaned up some things a little bit + fixed some aiohttp deprecated methods (new `middleware` style + `request.can_read_body` calls).

Cool, if you're familiar with the problem, let's try to fix it then.

connexion/decorators/coroutines.py
```diff
@@ -0,0 +1,27 @@
+import asyncio
+import functools
+
+
+def get_request_life_cycle_wrapper(function, api, mimetype):
```
prawn-cake: Please, add a docstring here, elaborating why this is needed

connexion/decorators/coroutines.py
```diff
@@ -0,0 +1,27 @@
...
+            response = yield from response
+
+        response = yield from api.get_response(response, mimetype,
+                                               request)
+        return response
+
+    return asyncio.coroutine(wrapper)
+
+
+def get_response_validator_wrapper(function, _wrapper):
```
prawn-cake: And for this one

connexion/decorators/coroutines.py
```diff
@@ -0,0 +1,27 @@
...
+def get_request_life_cycle_wrapper(function, api, mimetype):
+    @functools.wraps(api.get_request)
+    def wrapper(*args, **kwargs):
+        request = yield from api.get_request(*args, **kwargs)
+        response = function(request)
+
+        if asyncio.iscoroutine(response):
+            response = yield from response
+
+        response = yield from api.get_response(response, mimetype,
```
prawn-cake: Does this return `ConnexionResponse` instance? Let's rename it to `connexion_resp` to distinguish native responses from connexion-wrapped ones

dutradda:
@prawn-cake, I think I did the requested changes. Can you take a look?

prawn-cake:
Can we bump aiohttp version to 2.3.10? The rest of the stuff looks good to me

dutradda:
@prawn-cake can be the aiohttp 3.0.1?
it is the latest one

prawn-cake:
3.0.1 has [too many changes](https://github.com/aio-libs/aiohttp/blob/master/CHANGES.rst) and probably backward incompatible. 

At least this one bugs me:
> Minimal supported Python versions are 3.5.3 and 3.6.0

prawn-cake:
:+1:

prawn-cake:
@hjacobs I guess we are done here.
@dutradda Thanks for the great contribution.

dutradda:
@prawn-cake 
Hello guys, do you know when this PR will be merged to the master?

haneefkassam:
This is amazing! (and I'm assuming paves the path to also using uvloop with connexion and flask, and async handler functions, right?  please correct me if I'm wrong)

prawn-cake:
@haneefkassam almost correct, this allows to switch the backend from flask to aiohttp, basically to make connexion async

hjacobs:
@dutradda let me check regarding merging, sorry for not getting back earlier.

dutradda:
@hjacobs thank you!

hjacobs:
Released in https://github.com/zalando/connexion/releases/tag/1.4

haneefkassam:
@prawn-cake Would this also mean swapping the backend from flask to sanic is much more trivial?

prawn-cake:
@haneefkassam it's definitely easier since we added api abstraction, but not sure how trivial it is to add sanic, need to check implementation details.

dutradda:
@haneefkassam
I alreadly used sanic and I can say that is very to plug it on connexion.
But the aiohttp is not good for you?
I switched from sanic to aiohttp because its maturity. And it peforms well too

haneefkassam:
@dutradda No no, it's not that aiohttp is not good; I've just seen sanic vs aiohttp go in favour of sanic (without considering replacing the purely python asyncio with uvloop in aiohttp).

I'm not one to be un-appreciative though; I'm loving the fact this PR was merged and love the fact that I now have the option of using python3 async features with connexion!

roman-telepathy-ai:
@dutradda do you have some examples for connexion + aiohttp?
* async get / post
* async get / post with blocking calls
* websockets

Would be interesting how websockets would be represented in swagger.yaml.

thx in advance!

maitham1:
Has anyone managed to get a websocket example working?

PR #536, opened 2017-11-02
Allow splitting spec file using jinja2 env and/or json references

### LINKED ISSUE #254: Breaking up large swagger.yaml files

jadkik:
Fixes #254 .

This breaks almost all tests, most probably because this change affects the way definitions are loaded and the Swagger spec file is loaded.

It's still work in progress, and any feedback or help is welcome. It's "works on my machine ™️" quality till now.

 - Use a jinja2 environment when rendering the YAML swagger spec to allow to include other jinja2 files in the template (including macros, etc.).
 - Resolve JSON references, and render those too using jinja2 when applicable (based on file extension)
 - Use `jsonschema`'s `RefResolver` to load all definitions (WIP)

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -7,14 +7,13 @@ from typing import AnyStr, List  # NOQA
 
 import jinja2
 import six
-import yaml
-from swagger_spec_validator.validator20 import validate_spec
 
 from ..exceptions import ResolverError
 from ..operation import Operation
 from ..options import ConnexionOptions
 from ..resolver import Resolver
 from ..utils import Jsonifier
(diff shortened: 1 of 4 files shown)


hjacobs:
I'm closing this PR for now (as it's pretty old).

PR #542, opened 2017-11-10
Use Authorization headers for verifying token info

### LINKED ISSUE #541: verify_oauth uses url params instead of Authorization headers for token info
### Description

``verify_oauth`` uses url params instead of Authorization headers for token info.
Using Authorization header is the recommended way for verifying tokens.

Using url params could lead to leaking access tokens to external services. We faced this issue during testing OpenTracing. Authorization headers provides a more structured way in preventing leaking access tokens (e.g. in opentracing tags).

### Expected behaviour

Using Authorization headers for verifying tokens.

### Actual behaviour

Using URL params for verifying tokens.

### Additional info:

Output of the commands:

- `python --version`
Python 3.5.2

mohabusama:
Fixes #541 

 - Change token info verification to use Authorization headers rather than url params.

Note: Not sure if url params is a requirement, left it in the test case fixture, but could be removed if Authorization headers would suffice.

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -97,7 +97,7 @@ class FlaskApp(AbstractApp):
                 import tornado.wsgi
                 import tornado.httpserver
                 import tornado.ioloop
-            except:
+            except ImportError:
                 raise Exception('tornado library not installed')
             wsgi_container = tornado.wsgi.WSGIContainer(self.app)
             http_server = tornado.httpserver.HTTPServer(wsgi_container, **options)
@@ -107,7 +107,7 @@ class FlaskApp(AbstractApp):
         elif self.server == 'gevent':
             try:
                 import gevent.wsgi
-            except:
+            except ImportError:
                 raise Exception('gevent library not installed')
             http_server = gevent.wsgi.WSGIServer((self.host, self.port), self.app, **options)
             logger.info('Listening on %s:%s..', self.host, self.port)
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -16,12 +16,12 @@ def validate_wsgi_server_requirements(ctx, param, value):
     if value == 'gevent':
         try:
             import gevent  # NOQA
-        except:
+        except ImportError:
             fatal_error('gevent library is not installed')
     elif value == 'tornado':
         try:
             import tornado  # NOQA
-        except:
+        except ImportError:
             fatal_error('tornado library is not installed')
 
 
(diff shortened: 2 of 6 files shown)


hjacobs:
@mohabusama please also update the oauth2 example in the `examples` folder: https://github.com/zalando/connexion/tree/master/examples/oauth2

mohabusama:
@hjacobs Done.

hjacobs:
:+1:

kislyuk:
This breaks compatibility with the Google tokeninfo API. It would be nice if this could be reverted, released on a hotfix basis, and then made configurable and re-committed.

hjacobs:
@kislyuk do you have some doc regarding Google tokeninfo? It's definitely wrong if they propose passing the token as a query parameter. See also https://tools.ietf.org/html/rfc7662:

```
To prevent the values of access tokens from leaking into server-side
logs via query parameters, an authorization server offering token
introspection MAY disallow the use of HTTP GET on the introspection
endpoint and instead require the HTTP POST method to be used at the
introspection endpoint.
```
(token introspection is the effort to standardize the tokeninfo idea)

komasoftware:
While Google token info maybe wrong in passing the token only as a query parameter, it is one of the most used authentication methods. Killing it off makes this library unusable for a lot of use cases.

mohabusama:
@kislyuk @komasoftware I have created this issue to enable verification using url params as a configuration.

I have checked [Google token info docs](https://developers.google.com/identity/sign-in/web/backend-auth) and it seems to me that it accepts ``id_token`` as param and not ``access_token`` as implemented in Connexion. Was wondering if this actually worked before out-of-the-box?

kislyuk:
It did work before. I don't recall the details. I had to lock my requirements to a version of connexion before this change was made.

kislyuk:
OK, I recall now. We used connexion just for the auth middleware, not for the oauth roundtrip. The token was obtained independently of connexion. However, the update broke the auth middleware, since it became unable to successfully call the tokeninfo route to verify the token. You can see the Swagger YAML for the app in question here: https://github.com/HumanCellAtlas/data-store/blob/master/dss-api.yml

komasoftware:
Rolling back to version connexion==1.1.15
This config works :
  googAuth:
    type: oauth2
    flow: implicit
    authorizationUrl: https://accounts.google.com/o/oauth2/v2/auth
    x-tokenInfoUrl: https://www.googleapis.com/oauth2/v3/tokeninfo
    scopes:
      https://www.googleapis.com/auth/userinfo.email: read user email
But not with the latest release

PR #543, opened 2017-11-14
Add for support relative paths in schema refs

### LINKED ISSUE #461: Impossible to reference local files
### Description

`connexion` doesn't appear to be able to resolve local files `$ref` references, resulting in an exception being thrown.

### Expected behaviour

`$ref` should be able to reference local files.

### Actual behaviour

When using relative references as defined in https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#relative-files-with-embedded-schema-example connexion fails with the following exception:

```python traceback
  ... SNIP... 
  File "/home/snip/.venv/lib/python3.6/site-packages/jsonschema/validators.py", line 366, in resolving
    url, resolved = self.resolve(ref)
  File "/home/snip/.venv/lib/python3.6/site-packages/jsonschema/validators.py", line 375, in resolve
    return url, self._remote_cache(url)
  File "/home/snip/.venv/lib/python3.6/site-packages/jsonschema/validators.py", line 385, in resolve_from_url
    raise RefResolutionError(exc)
jsonschema.exceptions.RefResolutionError: unknown url type: 'definitions.yaml'
```

Tracing the code, it would appear that `connexion` relies on the `jsonschema` library to resolve all `$ref` references. Unfortunately, as far as I know, Json Schema doesn't support relative path file references, and in that specific implementation, the fallback mechanism is to use `urllib`'s `urlopen`... Which doesn't manage to open the file.

### Steps to reproduce

Using relative file `$ref`

### Additional info:

We may be able to alter the implementation by altering the value of the `$ref` values we provide to `jsonschema`.

For instance, if we had this OpenAPI reference: 
```yaml
$ref: "definitions.yaml#/Pet"
```

We should be able to overcome `jsonschema`'s limitation by altering the `$ref` value in such a fashion that it's able to open it in `urlopen`:

```yaml
$ref: "file:///absolute/path/to/definitions.yaml#/Pet"`
```

s-knibbs:
Refs #461
Refs #254 

Fix to allow the json schema validation to handle relative refs by passing in the spec file path. This will only work when passing a file path to `add_api` and not a `dict`.

**Note:** This will only work with response validation disabled. There are more changes required to support
local refs elsewhere in the code.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -78,10 +78,14 @@ class AbstractAPI(object):
                             'swagger_url': self.options.openapi_console_ui_path,
                             'auth_all_paths': auth_all_paths})
 
+        # File url for resolving local refs
+        specification_file_url = ''
+
         if isinstance(specification, dict):
             self.specification = specification
         else:
             specification_path = pathlib.Path(specification)
+            specification_file_url = str(specification_path.resolve().as_uri())
             self.specification = self.load_spec_from_file(arguments, specification_path)
 
         self.specification = compatibility_layer(self.specification)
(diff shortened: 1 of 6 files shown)


prawn-cake:
Any tests for this usecase? =)

dtkav:
Closing in favor of the newer #798

PR #545, opened 2017-11-14
refactor catch-all exceptions

mvalkon:
The flake8 checks were failing for py3.5 due to a couple of catch-all
exceptions when importing tornado and gevent libraries. Refactored to
catch the `ImportError`.

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -97,7 +97,7 @@ class FlaskApp(AbstractApp):
                 import tornado.wsgi
                 import tornado.httpserver
                 import tornado.ioloop
-            except:
+            except ImportError:
                 raise Exception('tornado library not installed')
             wsgi_container = tornado.wsgi.WSGIContainer(self.app)
             http_server = tornado.httpserver.HTTPServer(wsgi_container, **options)
@@ -107,7 +107,7 @@ class FlaskApp(AbstractApp):
         elif self.server == 'gevent':
             try:
                 import gevent.wsgi
-            except:
+            except ImportError:
                 raise Exception('gevent library not installed')
             http_server = gevent.wsgi.WSGIServer((self.host, self.port), self.app, **options)
             logger.info('Listening on %s:%s..', self.host, self.port)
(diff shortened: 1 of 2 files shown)


mvalkon:
I am actually of the opinion to remove catching the exceptions at all, since we're now essentially converting the `ImportError` to an `Exception`.

connexion/apps/flask_app.py
```diff
@@ -97,7 +97,7 @@ def run(self, port=None, server=None, debug=None, host=None, **options):  # prag
                 import tornado.wsgi
                 import tornado.httpserver
                 import tornado.ioloop
-            except:
+            except ImportError:
                 raise Exception('tornado library not installed')
```
prawn-cake: I guess for this purpose raising `RuntimeError(...)` would be more appropriate

mvalkon: At least it is more specific than an `Exception`. However regardless of which exception is raised, the application terminates. Given that the `server` is a user provided parameter, isn't letting the `ImportError` bubble up to the cli the clearest way of informing the user that the library they want to use is not installed?

prawn-cake: Not always. `ImportError` may also indicate that `PYTHONPATH` is wrong. I think the biggest value here is the error description or the message the user sees and it's quite clear.

prawn-cake:
:+1:

connexion/apps/flask_app.py
mvalkon: Perhaps my wording was poor, but that is exactly what I meant, the library cannot be _found_. 
```
>>> import tornado
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'tornado'
```

vs

```
>>> raise RuntimeError("Module 'tornado' cannot be found")
  File "<stdin>", line 1, in <module>
RuntimeError: Module 'tornado' cannot be found
```

I think we're splitting hairs now :smile:

prawn-cake: I just pointed to the fact the there are at least two reasons why module can't be found: it's not installed or it's installed but PYTONPATH is wrong and should we do `pip install` or check our environment variables. So it's kind of make sense to be precise here when it comes to debugging.
To me, giving the message that it's not installed (which is more probable) is better than just the module is not found, purely for debug reasons

hjacobs:
:+1:

Issue #427, opened 2017-04-05
Local Token Validation (in-process)


horomena:
### Description
In a micro-service architecture where the bearer token is passed to each service, it's much more efficent for each micro-service to validate the token locally than calling a centralized authorization service.

### Expected behaviour
Point "x-tokenInfoUrl" (or a different attribute) to a local function that will validate the bearer token locally.

### Actual behaviour
"x-tokenInfoUrl" only accepts URLs.

hjacobs:
@horomena we (at Zalando) actually support local token validation via HTTP, e.g. by running https://github.com/zalando/planb-tokeninfo as a sidecar in your Kubernetes pod :smirk: 

But in general it sounds like a good idea and it's related to #124.

MaicoTimmerman:
Would it be a solution to add something like `x-tokenInfoFunc` in which we pass a function similar to the `operationId`?

We'd then adapt the [security_decorator](https://github.com/zalando/connexion/blob/c30ba1f571c6a5af2800dcd3ef06b089f994c651/connexion/operation.py#L48) to check whether any of the two is present and return the respective partial. Where `x-tokenInfoUrl` returns the implementation which checks the remote endpoint and `x-tokenInfoFunc` returns the implementation which calls a local function.

horomena:
Yep, that would be great

hjacobs:
@MaicoTimmerman sounds good, would you be willing to implement this? :smile:

MaicoTimmerman:
We can close this issue.

hjacobs:
Closing as it was released here: https://github.com/zalando/connexion/releases/tag/1.3

PR #559, opened 2018-01-10
Added vendor extension for local token validation.

MaicoTimmerman:
Implements a method to locally validate the Bearer token passed to endpoints. See #427.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -6,6 +6,7 @@ import textwrap
 
 import requests
 
+from connexion.utils import get_function_from_name
 from ..exceptions import OAuthProblem, OAuthResponseProblem, OAuthScopeProblem
 
 logger = logging.getLogger('connexion.api.security')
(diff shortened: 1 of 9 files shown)


connexion/decorators/security.py
```diff
@@ -17,14 +18,27 @@
...
+def get_tokeninfo_func(security_definition):
+    """
+    :type security_definition: dict
+    :rtype: function
+
+    >>> get_tokeninfo_url({'x-tokenInfoFunc': 'foo.bar>'})
+    '<function foo.bar>'
+    """
+    token_info_func = (security_definition.get("x-tokenInfoFunc") or
+                       os.environ.get('TOKENINFO_FUNC'))
```
hjacobs: passing an (internal) function name via environment variable does not make sense I think

MaicoTimmerman: This function name is the function in the users application, not in the library. So I think we should offer the same methods for the tokenInfoFunc as the tokenInfoUrl.

What is the reason you'd not want to include this?

hjacobs: @MaicoTimmerman I can just not a find a good reason for it: the app and the swagger file should both be static (i.e. compile-time), environment variables are for runtime configuration and I don't see a good argument why you would want to switch the function at runtime (startup). We could leave it in, but I stumbled over it and it immediately raised the question "why? what's the benefit?".

hjacobs: On the other hand I could imagine having a "noop" token validation function for local testing, so maybe your `TOKENINFO_FUNC` makes sense to overwrite it with empty value "" (no token validation).

MaicoTimmerman: I'd personally prefer to keep it like it is currently, so can we proceed to merge the PR?

PR #560, opened 2018-01-13
Fix tests

hjacobs:
Follow up to #559:
* fix isort
* no longer test dev dependencies (for now) as they are breaking tests


hjacobs:
:+1:

PR #565, opened 2018-02-08
Allow token_info response to use 'scopes' key

skaasten:
Some oauth libraries use the key 'scopes' instead of 'scope' in the token info response.

https://github.com/doorkeeper-gem/doorkeeper/wiki/API-endpoint-descriptions-and-examples

 - Allow token info responses that use 'scopes' instead of 'scope'

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -73,10 +73,11 @@ def validate_token_info(token_info, allowed_scopes):
     :type token_info: dict
     :return: None
     """
-    if isinstance(token_info['scope'], list):
-        user_scopes = set(token_info['scope'])
+    scope = token_info.get('scope') or token_info.get('scopes')
+    if isinstance(scope, list):
+        user_scopes = set(scope)
     else:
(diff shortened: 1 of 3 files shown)


hjacobs:
I don't understand your change proposal, the docs you are linking to are talking about the RFC 7662 OAuth 2.0 Token Introspection which returns the `scope` field, see https://tools.ietf.org/html/rfc7662#section-2.2

skaasten:
Here's the example, in that linked document further down, of the response from the ```token/info``` call implemented by doorkeeper:
```
curl -H "Authorization: Bearer 53cff8f4a549beb1c38704158b0f6608a2382f094b6947ecc35c2eed4146a17c" 
localhost:3000/oauth/token/info

### command output
{"resource_owner_id":1, "scopes":[], "expires_in_seconds":7178, "application":{"uid":null}, "created_at":1440460991}
```

hjacobs:
@skaasten OK, I see and now it makes sense!

hjacobs:
:+1:

PR #568, opened 2018-02-18
Support displaying swagger.json when intermediate path routing is present

stopmachine:
For example if you have an application load balancer that redirects:
    
    http://myhost.example.com/my-first-app/resource/in/app

to:

    http://internal.host/resource/in/app

Then assuming your flask/connexion app listens on `internal.host` and your basePath is `/resource`, then accessing `http://myhost.example.com/my-first-app/resource/ui` will not work, since it will try to fetch `http://myhost.example.com/resource/swagger.json`, unknown to the load balancer.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -119,9 +119,12 @@ OAuth 2 Authentication and Authorization
 
 Connexion supports one of the three OAuth 2 handling methods. (See
 "TODO" below.) With Connexion, the API security definition **must**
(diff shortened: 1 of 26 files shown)


stopmachine:
@hjacobs Any chance this gets merged soon? Any changes required?

huesforalice:
@stopmachine thanks for this PR. We would also like to use it this way and see it get merged. If there is any help needed, I can contribute.

dtkav:
closing this because swagger-ui is no longer vendored in connexion (as of v2.0)

PR #587, opened 2018-05-25
query parameters only explode if collectionFormat is 'multi'

### LINKED ISSUE #323: Support for type array with collectionFormat multi

dtkav:
Fixes #323 .
This changes the behavior from #500 to adhere to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7

While there is no defined HTTP spec on this behavior (see https://stackoverflow.com/a/1746566), the openapi3 spec provides some insight on the intent of the swagger2 spec here:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-values

> Form style parameters defined by RFC6570. This option replaces collectionFormat with a csv (when explode is false) or multi (when explode is true) value from OpenAPI 2.0.

 - support for `collectionFormat: multi`
 - query parameters that are defined multiple times use the last (right-most) one, unless the collectionFormat is multi

for example, if collectionFormat == csv then
`https://myapi.com/api?items=a&items=b,c`
would result in `items = ["b", "c"]`

for more concrete examples, see the unit tests.

@lhannest may have some context that I'm missing. It might be a contentious change because it could potentially break users who are relying on this behavior.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -111,13 +111,12 @@ def parameter_to_arg(parameters, consumes, function, pythonic_params=False):
             for k, v in request.query.to_dict(flat=False).items():
                 k = sanitize_param(k)
                 query_param = query_types.get(k, None)
-                if query_param is not None and query_param["type"] == "array":
-                    if query_param.get("collectionFormat", None) == "pipes":
-                        request_query[k] = "|".join(v)
-                    else:
-                        request_query[k] = ",".join(v)
+                if (query_param is not None
+                        and query_param["type"] == "array"
+                        and query_param.get("collectionFormat") == "multi"):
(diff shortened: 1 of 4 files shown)


dtkav:
It would be great to get some feedback on this.
@hjacobs do you have any thoughts?

dtkav:
closing in favor of #613

PR #595, opened 2018-06-06
Pass kwargs from RestyResolver to Resolver

u8sand:
Allows for `function_resolver` override

In my particular use case I choose to use a loaded python module. `RestyResolverEx` below is `RestyResolver` with my proposed change.

```python
class ModuleResolver(RestyResolverEx):
  def __init__(self, module, collection_endpoint_name='search'):
    """
    :param module: Default module name for operations
    :type module: any
    """
    RestyResolverEx.__init__(
      self,
      default_module_name=module.__name__,
      collection_endpoint_name=collection_endpoint_name,
      function_resolver=self.get_function_from_name,
    )
    self.module = module

  def get_function_from_name(self, function_name):
    """
    Tries to get function by fully qualified name (e.g. "mymodule.myobj.myfunc")
    :type function_name: str
    """
    if function_name is None:
      raise ValueError("Empty function name")
    function_name_parts = function_name.split('.')
    assert function_name_parts[0] == self.module.__name__
    return utils.deep_getattr(self.module, '.'.join(function_name_parts[1:]))
```

 - add kwargs to RestyResolver constructor
 - pass to Resolver initialization

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -76,6 +76,15 @@ class RestyResolver(Resolver):
         """
         :param default_module_name: Default module name for operations
         :type default_module_name: str
+        :param \**kwargs:
+            Pass arguments to Resolver.__init__
+
+        :Keyword Arguments:
+            * *function_resolver* (``types.FunctionType``) --
+              Function that resolves functions using an operationId
+
+        :param kwargs: Keyword arguments to pass on to Resolver.__init__
+        :type kwargs: dict
         """
         Resolver.__init__(self, **kwargs)
         self.default_module_name = default_module_name
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -72,12 +72,12 @@ class RestyResolver(Resolver):
     Resolves endpoint functions using REST semantics (unless overridden by specifying operationId)
     """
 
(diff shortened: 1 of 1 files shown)


hjacobs:
Can you add the `**kwargs` to the init's docstring? (to mention that it passes through the parameters)
Please also add an unit test for your use case.

u8sand:
@hjacobs thanks for your response. I will write those tests within the next day.

In terms of the use case for this PR it's just being able to modify Resolver's constructor via RestyResolver's constructor. I can write a test for that.

But my use case involves passing an object as opposed to a string (referring to a string). Let me know if I should make another pull request given that it's in the scope of this project.

E.g. instead of `add_api(spec, resolver=RestyResolver('my.path'))` you could do `import my.path; add_api(spec, resolver=RestyResolver(my.path))` passing the object itself.

dtkav:
Hey @u8sand - just wanted to check in if this is still something you're interested in, or if I should close it out.
There have been a lot of changes to connexion since June !

u8sand:
While I've worked around this problem in other ways, i think it's still a good idea for better extension; I'll address these things hopefully later today

PR #599, opened 2018-06-12
RestyResolver for nested paths

### LINKED ISSUE #206: Handle nested paths in swagger definitions
Connexion does not seem to handle nested paths.
Tried loading the example Uber swagger spec provided at [editor.swagger.io](editor.swagger.io).
Using RestyResolver to perform view resolution with 'api' as module.
Added placeholder views for all path actions.
Python 3.5.1 on MacOS
The spec has the following paths defined:
- /estimates/time
- /estimates/price
  Fails with `Attribute error: module 'api.estimates' has no attribute 'get'`
  There is no `get` action defined at the /estimates path although this seems to be what the resolver is looking for.
  I would expect it to use /api/estimates/time.py with get method and /api/estimates/price.py with get method instead.

e4r7hbug:
Fixes #206.

There has been some discussion about fixing `RestyResolver` to work with nested paths. I've bumped into this issue multiple times trying to create a new API with Connexion.

After looking over the regex, it seems that parsing would become extremely complicated. Going the more basic route, I feel `str.split()` can cover most of the situations and using conditionals can be more maintainable and easier to read for newcomers.

Hopefully the new test cases can clearly demonstrate what is handled by this code change. Do let me know if this is not the intended behaviour or if something must be changed. Feel free to take whichever parts of the code or test cases that would be beneficial to the project.

 - Support nested paths
 - Define a custom module name separator

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -1,5 +1,4 @@
 import logging
-import re
 import sys
 
 import connexion.utils as utils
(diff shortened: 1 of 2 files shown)


dtkav:
Hey @e4r7hbug is this issue something you're still interested in working on, or can I close it out?
There have been a lot of changes to connexion since June.

ntangsuratupgrade:
Sorry for the late reply! Since it sounds like I need to spend some time getting myself updated on the latest changes, I agree on closing this for now. I'll try opening another Pull Request in the future if it's still needed. Thanks for checking in!

paulkernstock:
I am very interested in this happening! Go go go!

connexion/resolver.py
```diff
@@ -98,33 +98,52 @@ def resolve_operation_id_using_rest_semantics(self, operation):
...
-        )
+        path_fragments = operation.path.split('/')
+
+        tail_parameter = ''
+        has_trailing_slash = False
+        if path_fragments:
+            tail = path_fragments[-1]
+            if tail == '':
+                has_trailing_slash = True
+            elif tail.startswith('{'):
```
reece: What about intermediate path segments with path parameters, like `/path/{id}/subpath`?

reece: Nevermind. I see it later on. Deleting comment.

PR #600, opened 2018-06-13
fix on utils.has_coroutine function

dutradda:
Fixed coroutinefunction verification when has async def.
   It is necessary because there's no @asyncio.coroutine decorator to validate

Fixes # .

 -
 -
 -

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -148,7 +148,6 @@ class Jsonifier(object):
 def has_coroutine(function, api=None):
     """
     Checks if function is a couroutine.
- 
     If ``function`` is a decorator (has a ``__wrapped__`` attribute)
     this function will also look at the wrapped function.
     """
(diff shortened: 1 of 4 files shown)


connexion/utils.py
```diff
@@ -148,12 +148,20 @@ def loads(self, data):
 def has_coroutine(function, api=None):
```
jmcs: Since the logic of this function is getting more complex can you add a docstring explaining it?

dutradda: I will do that @jmcs

dtkav: hey @dutradda , any chance you can pick this back up?

dtkav:
It looks like this has been broken for several months.
If you want to merge as-is, I'd be happy to go in and comment it.

jmcs:
@dtkav @dutradda I added the docstring. If you are ok with my description I would merge this.

dtkav:
lgtm!

jmcs:
:+1:

PR #605, opened 2018-06-15
Add capability to pass framework's request context to handler functions

### LINKED ISSUE #603: Unable to get request headers using AioHttpApp
### Description
Connexion does not pass in http headers to handler functions.  Instead one is supposed to use `connexion.request.headers` to access them. This is currently just an alias to `flask.request.headers` but this does not work with AIOHTTP.  Maybe the `aiohttp.web.Request` should be passed to the handler function.

### Additional info:

Output of the commands:

- `python --version`: 3.6.2
- `pip show connexion | grep "^Version\:"`: 1.4.2

skwong2:
Fixes #603

 - Add new argument (currently named `pass_context_arg_name` but could probably use a nicer name) to `connexion.apps.abstract.AbstractApp.add_api` to inject the framework's request context to the handler functions.

### DIFF
--- a/tests/aiohttp/test_aiohttp_simple_api.py
+++ b/tests/aiohttp/test_aiohttp_simple_api.py
@@ -16,7 +16,7 @@ def aiohttp_app(aiohttp_api_spec_dir):
     app = AioHttpApp(__name__, port=5001,
                      specification_dir=aiohttp_api_spec_dir,
                      debug=True)
-    app.add_api('swagger_simple.yaml', validate_responses=True)
+    app.add_api('swagger_simple.yaml', validate_responses=True, pass_context_arg_name='request_ctx')
     return app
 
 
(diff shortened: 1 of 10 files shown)


tests/aiohttp/test_aiohttp_simple_api.py
```diff
@@ -214,3 +214,10 @@ def test_create_user(test_client, aiohttp_app):
     user = {'name': 'Maksim'}
     resp = yield from app_client.post('/v1.0/users', json=user, headers={'Content-type': 'application/json'})
     assert resp.status == 201
+
+
+@asyncio.coroutine
+def test_access_request_context(test_client, aiohttp_app):
```
hjacobs: why is the parameter called `aiohttp_app` here? shouldn't it be the request context?

jmcs:
:+1:

Issue #420, opened 2017-03-22
OpenAPI v3 support


tomas-fp:
Just curious about OpenAPI 3 support plans.

See related post about Swagger 2 and OpenAPI 3 introduction: 
https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/

rafaelcaricio:
I have this blog post in my read buffer. 😄 

To be honest I haven't thought too much about the OpenAPI 3 in Connexion so far. But definitely it is time to start considering it. I will follow up in this thread later when I give this topic some more thought. Thank you for opening this issue.

If anyone have ideas/suggestions or is willing to help in this I will be glad to support.

tomas-fp:
Cool. I'm open to help in a near future, especially when in my team we can start considering OpenAPI 3 adoption across our services.

danballance:
I need to get up to speed on version 3 and would then like to help.

JuxhinDB:
Just went over the new OpenAPI 3.0 which looks quite nice. I'm also curious as to when we may consider supporting it in Connexion. :-)

danballance:
Will we try to simultaneously support both versions 2 and 3? Although more work, my assumption is this would be the way to go, otherwise we will break a lot of APIs currently built with Connexion.

rafaelcaricio:
@danballance Yes, we will support both. We will try to reuse as much as possible the codebase we have.

As side note, I am working on a sketch of the support for OpenAPI v3.

JuxhinDB:
Hey @rafaelcaricio just wanted to ask if you guys have made any progress on this and if you might need any form of testing.

rafaelcaricio:
@JuxhinDB Unfortunately no. I'm currently busy with some personal stuff and with little free time to take care of this right now. :(

JuxhinDB:
That's fine, @rafaelcaricio! Feel free to ping me in the future should you need any help with this.

smith-m:
support for 3.0 would be awesome. Parameter and anchor references are much flexible and it can mean less repeated code to document the same API.

mozbhearsum:
The OpenAPI 3.0 spec was finalized and released a couple of weeks ago: https://www.openapis.org/blog/2017/07/26/the-oai-announces-the-openapi-specification-3-0-0

JuxhinDB:
[...]

Issue #406, opened 2017-02-17
Remove Swagger UI vendor code


rafaelcaricio:
Today Connexion hosts a vendor of Swagger UI code. This is a highly unmaintainable code which is even customized for Connexion needs. We have to review this approach.

We need to find an alternative way to have the support to Swagger UI in Connexion projects which must not need to have the whole vendor code inside Connexion.

rafaelcaricio:
Currently we have a script called [`update.sh`](https://github.com/zalando/connexion/blob/master/connexion/vendor/swagger-ui/update.sh) that applies a small [patch](https://github.com/zalando/connexion/blob/master/connexion/vendor/swagger-ui/index.html.patch) to the original Swagger UI code.

We should think if it would be possible to use the Swagger UI code without modifications and completely remove the Swagger UI code from the Connexion repo.

An possible option would be to create an external Python package that bundles the Swagger UI. That Swagger UI package would be optionally installed by users of Connexion if they want to use it.

jmcs:
> An possible solution would be to create an external Python package that bundles the Swagger UI. That Swagger UI package would be optionally installed by users of Connexion if they want to use it.

We can evaluate:

- https://pypi.python.org/pypi/flask-swagger-ui/
- https://pypi.python.org/pypi/flask-swaggerui/

rafaelcaricio:
Opened an issue in the first project: https://github.com/sveint/flask-swagger-ui/issues/2

bobh66:
The flask-restplus package includes a nice swagger-ui implementation.  Would it make sense to use that if the package is installed?

enezhadian:
@rafaelcaricio An alternative approach could be requiring users to provide the code for Swagger UI as a configuration for Connexion app. I don't think simply bundling Swagger UI in an external package will solve the issue of maintainability.

rafaelcaricio:
@enezhadian I agree, users should provide the path for the Swagger UI code (if they want to use it). And we should remove the Swagger UI code completely from Connexion codebase.

The external package idea is a call for anyone that wants to create it and maintain. So we (Connexion) can recommend to users to install that package and not rely on Connexion to provide the Swagger UI.

beardedeagle:
@rafaelcaricio I have developed a flask plugin called Flask-SimplySwagger that does most if not all of what you are asking for. I am currently in the process of getting it open sourced through my company, but once it is available, I would be more than happy to have your team use it:
```
What it does:

[...]

PR #606, opened 2018-06-20
updated swagger-ui to latest for oas3 support

### LINKED ISSUE #418: swagger-ui 3.x support
swagger-ui 3.x recently dropped. It has enough changes that the ```update.sh``` script cannot automatically upgrade it. It would be cool if we could get this in here.

fehguy:
Fixes #418 .

Less code!  This update uses unpkg.com to host the swagger-ui artifacts rather than use the previous `update.sh` script that lived in the vendor folder.

If maintaining a specific version of swagger-ui is desired, I can scope the version of the three assets.  Please let me know

 - Update of the `index.html` to use the updated swagger-ui container
 - Removal of the `css`, `fonts`, `images`, `lang`, `lib` folders, as well as the `update.sh`, `o2c.html`, `*.js` files
 - Added links to `unpkg.com` for serving three static assets


fehguy:
This, or something like it, will be needed for #420

dtkav:
I definitely prefer vendoring swagger-ui (the current approach) over depending on a random 3rd party.
Also, I'd highly recommend always pinning your dependencies in general - using `latest` means that a connexion release could be broken by a swagger-ui update.

connexion/vendor/swagger-ui/index.html
```diff
@@ -5,104 +5,38 @@
...
-
-      function log() {
-        if ('console' in window) {
-          console.log.apply(console, arguments);
-        }
-      }
-  });
-  </script>
+  <script src="https://unpkg.com/swagger-ui-dist/swagger-ui-standalone-preset.js"></script>
+  <script src="https://unpkg.com/swagger-ui-dist/swagger-ui-bundle.js"></script>
```
dtkav: also pin these? looks like you used https://unpkg.com/swagger-ui-dist@3.12.1/ below.

fehguy:
Thanks @dtkav.  I'll remove the reference to `unpkg.com` and pull in the js artifacts with an updated script.  Will update this PR shortly.

dtkav:
Hey @fehguy - There was more context in #406 that I didn't realize. Maybe weigh in there? Also as a disclaimer - I just wanted to share my personal opinion above - the core connexion folks might have different opinions from me :smile:

fehguy:
@hjacobs what direction do you suggest on this front?  I'll redo the PR as fit 👍

rafaelcaricio:
I would vote for the #614 approach as optional (see my comment there).

jmcs:
We are going with the #614 approach so I'm closing this PR. Thanks for PR anyway.

PR #610, opened 2018-06-29
Correct validator response on bad JSON

positron96:
Right now request validator returns 415 Unsupported Media Type "Invalid Content-type (application/json), expected JSON data" when Content-Type is set to "application/json" (correctly) but the payload itself is not valid JSON. This error message is misleading into thinking that there is something wrong with Content-Type header (but it's not)

This patch fixes it by separating two cases:
 - Request content-type is not JSON, in this case old behaviour is retained
 - Request content-type is JSON but payload cannot be parsed into a valid JSON, in this case 400 Bad Request "Request body is not valid JSON" is returned.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -10,7 +10,7 @@ from werkzeug import FileStorage
 
 from ..exceptions import ExtraParameterProblem
 from ..problem import problem
-from ..utils import all_json, boolean, is_null, is_nullable
+from ..utils import all_json, boolean, is_null, is_nullable, is_json_mimetype
 
 logger = logging.getLogger('connexion.decorators.validation')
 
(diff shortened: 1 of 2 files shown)


dtkav:
Nice! You might want to update `tests/api/test_responses.py::test_post_wrong_content_type` to match.

connexion/decorators/validation.py
```diff
@@ -107,12 +107,24 @@ def wrapper(request):
...
                 if data is None and len(request.body) > 0 and not self.is_null_value_valid:
-                    # the body has contents that were not parsed as JSON
-                    return problem(415,
-                                   "Unsupported Media Type",
-                                   "Invalid Content-type ({content_type}), expected JSON data".format(
-                                       content_type=request.headers.get("Content-Type", "")
-                                   ))
+                    try:
+                        ctype_is_json = is_json_mimetype(request.headers.get("Content-Type", "") )
+                    except:
```
dtkav: What exceptions are you expecting here, and why?
FYI bare excepts are a [python-anti-pattern](https://docs.quantifiedcode.com/python-anti-patterns/correctness/no_exception_type_specified.html#no-exception-type-s-specified)

positron96: is_json_mimetype is throwing ValueError when passing empty string (in case Content-Type is not present). I'll change to a specific except clause.

jmcs:
@positron96 can you fix the tests?

dtkav:
Hey @positron96 - you just need to rebase onto master. The swagger errors were caused by a mix of unpinned dependencies and a new release of swagger-spec-validator. The problem was fixed upstream.
See [this build](https://travis-ci.org/dtkav/connexion/builds/402057231) where I've copied your code and rebased onto master.

You just have one `isort` error to fix up, and you should be good to go once you rebase:
```
ERROR: /home/travis/build/dtkav/connexion/connexion/decorators/validation.py Imports are incorrectly sorted.
--- /home/travis/build/dtkav/connexion/connexion/decorators/validation.py:before	2018-07-10 04:41:46.965250
+++ /home/travis/build/dtkav/connexion/connexion/decorators/validation.py:after	2018-07-10 04:45:22.764291
@@ -10,7 +10,7 @@
 
 from ..exceptions import ExtraParameterProblem
 from ..problem import problem
-from ..utils import all_json, boolean, is_null, is_nullable, is_json_mimetype
+from ..utils import all_json, boolean, is_json_mimetype, is_null, is_nullable
 
 logger = logging.getLogger('connexion.decorators.validation')
```

jmcs:
:+1:

jmcs:
Thanks for the PR

positron96:
no problems)

PR #611, opened 2018-07-02
Upgrade to swagger-spec-validator 2.3.1 (fixes #612)

### LINKED ISSUE #612: Lastest version of swagger-spec-validator (2.3.1) breaks master build
### Description
The latest version of swagger-spec-validator now validates that defaults are the correct type.
Since the library is not pinned, the release has broken master.

(tests/test_api.py::test_other_errors_stop_application_to_setup) relies on the old behavior (don't exit on default type errors if `app(debug=true)`)
Also, one of the test fixtures has a (type: array) without an `items: ...` block, which is now invalid.

### Additional info:
#611 fixes the build by upgrading to the latest version.
An alternate approach would be to pin to 2.0.2 of swagger-spec-validator
Let me know if that's preferred and I can put up a PR.

dtkav:
Master is boken because of an update to swagger-spec-validator. This upgrades to the latest version and should fix the build.

 - upgrade to swagger-spec-validator 2.3.1
 - `debug=true` will no longer continue if an operation has invalid defaults (change was in swagger-spec-validator)


hjacobs:
Is there any way to get the `debug=true` behavior back? It's pretty useful for quick iterations during development when the spec is being changed and temporarily (partly) broken.

dtkav:
@hjacobs I can hard-pin (==) to 2.0.1 instead to preserve the previous behavior if you prefer that.

The only real change here is that default validation happens earlier. Personally I'd prefer for it to crash as early as possible when the spec is wrong, because then you know the problem is unrelated to connexion, and there are several tools available for debugging invalid specs.

The previous behavior would have avoided crashing in Operation.validate_defaults(), which also made sense given the lack of functionality in the spec validator.

Another way to handle you request would be to ignore _all_ validation errors in apis/abstract.py, but that seems like a recipe for heaps of stack traces :slightly_smiling_face: .

hjacobs:
:+1:

jmcs:
:+1:

PR #613, opened 2018-07-02
URI parsing decorator

### LINKED ISSUE #608: Default parameters break with collectionFormat csv
### Description

Can no longer define a `default` value for a query parameter with `type: array`, `items: {type: string}` and `collectionFormat: csv`. Possibly this is because of a change in swagger-spec-validator.

### Expected behaviour

In connexion version 1.3 (and swagger-spec-validator 2.1.0), a string default value worked in the above case.
An array default value would be fine too, and be more consistent with the `collectionFormat`.

### Actual behaviour

With a string default, eg. ```default: foobar```,
```
swagger_spec_validator.common.SwaggerValidationError:
("'foobar' is not of type 'array'\n\n
Failed validating 'type' in schema:\n
    {'collectionFormat': 'csv',\n
     'default': 'foobar',\n
     'in': 'query',\n
     'items': {'type': 'string'},\n
     'name': 'bar',\n
     'type': 'array'}\n\n
On instance:\n
    'foobar'", <ValidationError: "'foobar' is not of type 'array'">)
```
With an array default, eg. ```default: ["foobar"]```
```
  File ".../connexion/decorators/validation.py", line 55, in validate_type
    parts = value.split(",")
AttributeError: 'list' object has no attribute 'split'
```

### Steps to reproduce

Use this `swagger.yaml` file: 
```
swagger: "2.0"
info:
  title: Thing
  version: "0"
paths:
  /foo:
    get:
      operationId: "test_app.api.foo"
      parameters:
        - $ref: "#/parameters/bar"
      responses:
        200:
          description: "Results."

parameters:
  bar:
    name: bar
    in: query
    type: array
    items:
      type: string
    collectionFormat: csv
    default: foobar
```
Use this `api.py`:
```
import connexion
def foo(bar):
  return bar
```
then execute:
```
import os
swagger_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'swagger.yaml')
app.add_api(swagger_path, base_path='/v0', strict_validation=True, validate_responses=False)
app.run()
```

### Additional info:

Output of the commands:

- `python --version`
Python 3.6.5 :: Anaconda, Inc.
Version: 1.4.2

Also using swagger-spec-validator version 2.3.1.

dtkav:
Fixes #608 
Supersedes #587 

 - array logic refactored into one place.
 - validation.py and parameter.py no longer try to join the array, and the split it again.
 - validation of defaults now works, because the validator is given the correct type.
 - some additional classes that change the behavior of deduplicating query parameters that are defined multiple times
    - **AlwaysMultiURIParser** that is backwards compatible, warts and all (used by default)
    - **Swagger2URIParser** that adheres to the spec's definition of `collectionFormat: multi` and uses the last-defined query parameter value (ex. `query?a=1&a=2` => `a = 2`)
    - **FirstValueURIParser** that behaves like Swagger2URIParser, except that the first-defined value is used (ex. `query?a=1&a=2` => `a=1`)

This cleanup is really helpful for OAS3 support as well (#420).

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -14,6 +14,7 @@ from .decorators.response import ResponseValidator
 from .decorators.security import (get_tokeninfo_func, get_tokeninfo_url,
                                   security_passthrough, verify_oauth_local,
                                   verify_oauth_remote)
+from .decorators.uri_parsing import AlwaysMultiURIParser
 from .decorators.validation import (ParameterValidator, RequestBodyValidator,
                                     TypeValidationError)
 from .exceptions import InvalidSpecification
@@ -389,6 +390,10 @@ class Operation(SecureOperation):
         for validation_decorator in self.__validation_decorators:
             function = validation_decorator(function)
 
+        uri_parsing_decorator = self.__uri_parsing_decorator
+        logging.debug('... Adding uri parsing decorator (%r)', uri_parsing_decorator)
+        function = uri_parsing_decorator(function)
+
         # NOTE: the security decorator should be applied last to check auth before anything else :-)
         security_decorator = self.security_decorator
         logger.debug('... Adding security decorator (%r)', security_decorator)
@@ -401,6 +406,16 @@ class Operation(SecureOperation):
         function = self._request_end_lifecycle_decorator(function)
 
         return function
+
+    @property
+    def __uri_parsing_decorator(self):
+        """
+        Get uri parsing decorator
+
+        This decorator handles query and path parameter deduplication and
+        array types.
+        """
+        return AlwaysMultiURIParser(self.parameters)
 
     @property
     def __content_type_decorator(self):
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -142,7 +142,7 @@ class Operation(SecureOperation):
                  path_parameters=None, app_security=None, security_definitions=None,
                  definitions=None, parameter_definitions=None, response_definitions=None,
                  validate_responses=False, strict_validation=False, randomize_endpoint=None,
-                 validator_map=None, pythonic_params=False):
+                 validator_map=None, pythonic_params=False, uri_parser_class=None):
         """
         This class uses the OperationID identify the module and function that will handle the operation
 
@@ -189,6 +189,8 @@ class Operation(SecureOperation):
         :param pythonic_params: When True CamelCase parameters are converted to snake_case and an underscore is appended
         to any shadowed built-ins
         :type pythonic_params: bool
+        :param uri_parser_class: A URI parser class that inherits from AbstractURIParser
+        :type uri_parser_class: AbstractURIParser
         """
 
         self.api = api
@@ -210,6 +212,7 @@ class Operation(SecureOperation):
         self.operation = operation
         self.randomize_endpoint = randomize_endpoint
         self.pythonic_params = pythonic_params
+        self.uri_parser_class = uri_parser_class
 
         # todo support definition references
         # todo support references to application level parameters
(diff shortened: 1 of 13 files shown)


tests/decorators/test_uri_parsing.py
```diff
@@ -0,0 +1,65 @@
...
+        (Swagger2URIParser, ['d', 'e', 'f'], QUERY1, CSV),
+        (FirstValueURIParser, ['a'], QUERY1, CSV),
+        (AlwaysMultiURIParser, ['a', 'b', 'c', 'd', 'e', 'f'], QUERY1, CSV),
+        (Swagger2URIParser, ['a', 'b', 'c', 'd', 'e', 'f'], QUERY1, MULTI),
+        (FirstValueURIParser, ['a', 'b', 'c', 'd', 'e', 'f'], QUERY1, MULTI),
+        (AlwaysMultiURIParser, ['a', 'b', 'c', 'd', 'e', 'f'], QUERY1, MULTI),
+        (Swagger2URIParser, ['d', 'e', 'f'], QUERY2, PIPES),
+        (FirstValueURIParser, ['a'], QUERY2, PIPES),
+        (AlwaysMultiURIParser, ['a', 'b', 'c', 'd', 'e', 'f'], QUERY2, PIPES)])
+def test_uri_parser_query_params(cls, expected, query_in, collection_format):
```
rafaelcaricio: It's a simple thing, but I would suggest to change the name of the first argument from "cls" to "parser_class", I was very confused seen this code at first due to https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments

dtkav: sure thing - thanks for the review!

dtkav:
hey @rafaelcaricio - any thoughts on backwards compatibility vs correctness? I've defaulted to a backwards compatible mode, but it would be great to switch to the Swagger2URIParser on a major version bump or something.

jmcs:
Thanks for the PR. How would I change the URI Parser for my application? This needs to be documented.

As for the default resolver the safest option will be to leave Swagger 2 as is and use do the "right thing" for OpenAPI 3.

dtkav:
@jmcs do you think users will want to override the URI Parser?
If so, how do you recommend adding the option? I can think of a couple ways:
1. Through ConnexionOptions
2. Allow the user to specify the `Operation` class to `add_operation(...)` in `apis/abstract.py`, so they can easily override methods, and specify their custom operation class.

If it won't be something that users will use, shall I just remove the other classes? Having the uri parser api is handy for having a swagger2 and openapi3 implementation, but I don't need to supply additional swagger2 classes if they aren't helpful. I'm happy to remove them - they were still a useful discussion aid for potentially changing the default behavior.

jmcs:
@dtkav I think that `Swagger2URIParser` might be of interest for new applications.

Since `add_operations` is usually not called by the user directly, it should be configurable in `ConnexionOptions`.

dtkav:
@jmcs ok, I've added uri_parser_class to the ConnexionOptions class, a test case, and some docs.
Along the lines of idea 2 above, I could expose the `Operation` class through `options` instead, which would allow a user to override more behaviors, and would be a more generic solution.

connexion/operation.py
```diff
@@ -209,6 +212,7 @@ def __init__(self, api, method, path, operation, resolver, app_produces, app_con
         self.operation = operation
         self.randomize_endpoint = randomize_endpoint
         self.pythonic_params = pythonic_params
+        self.uri_parser_class = uri_parser_class
```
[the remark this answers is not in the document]
dtkav: Ah yeah, that's a better spot. The uri parser becomes an important interface of the Operation class with oas3, so my preference would be to leave the decorator, but I can also remove it now and we can discuss that patch separately.

dtkav:
here's an example of changing the Operation class.
https://github.com/dtkav/connexion/compare/uri_parsing_decorator...dtkav:uri_parsing_decorator_operation_override

dtkav:
I think the build is broken for python 3.4 - I'm looking into that separately, but it's getting near the end of my working day. Thanks for your thoughts and comments.

jmcs:
:+1:

hjacobs:
It's difficult for me to assess the impact of this change, but I'm OK with merging if @jmcs says :+1:.

PR #614, opened 2018-07-02
Remove vendored swagger-ui, add swagger-ui-bundle as optional pip extra

dtkav:
Fixes(?) #406 

 - add [swagger_ui_bundle](https://github.com/dtkav/swagger_ui_bundle) package as a new optional dependency
 - remove connexion/vendor/ directory

I'm putting this up as a demo to aid the discussion. I mentioned in #406 that this was a potential solution, but I'm not set on it. Ideally I'd love to have the maintainers decide on a path forward. Figuring this out will help make my openapi v3 support PR (#420) significantly smaller and easier to review (I'm currently just vendoring swagger-ui-3 alongside swagger-ui-2).

Thanks!


setup.py
```diff
@@ -27,6 +27,7 @@ def read_version(package):
     'requests>=2.9.1',
     'six>=1.9',
     'swagger-spec-validator>=2.3.1',
+    'swagger-ui-bundle==0.0.2',
```
rafaelcaricio: Could you please make this optional (to live in `extras_require`) ?

rafaelcaricio:
Thank you for this PR anyway.

dtkav:
@rafaelcaricio - thanks - I've made the dependency optional, have a look at the implementation and let me know what you think.
I've left the default options the same (connexion will attempt to use swagger-ui by default, warn on missing library unless disabled), but I could also change the behavior so it must be explicitly turned on.
My preference is to have "batteries included", i.e. if the library is installed then it works out of the box by default.

rafaelcaricio:
@dtkav Looks good to me. Thanks for the changes. That is what I was envisioning when I created the #406 .

dtkav:
Hey @rafaelcaricio are we waiting on review from Zalando folks, or what's the next step?

jmcs:
I say we merge this, but we should increase Connexion version to 1.5 to make this transparent (or even 2.0 but then I would also like to release OpenAPI 3 support at the same time). What do you think @hjacobs?

rafaelcaricio:
@jmcs I also think this is a major breaking change and should be in the 2.0.0 version. Also agree we could put this in together with OpenAPI 3.0.0 support and that seems to be moving in a good pace.

hjacobs:
@rafaelcaricio @jmcs SGTM

rafaelcaricio:
@jmcs @hjacobs I've created now a new temporary label `v2.0.0` so we can mark PRs we want to be merged for the version `2.0.0`.

hjacobs:
@rafaelcaricio milestones are better for this I guess. I created: https://github.com/zalando/connexion/milestone/10

rafaelcaricio:
Ok, I deleted the label then.

dtkav:
Would it be possible to create a 2.0 staging branch? That way this change can get merged into the 2.0 staging branch, and then the oas3 branch can be reviewed as a PR against that branch without also containing all of these changes.

dtkav:
Also, what is the best way to discuss changes for the milestone release? There are a few cleanup things we could include - removing "old style" connexion options, changing some of the names of options from "swagger" to "openapi", and changing the default URI parser for swagger 2 specs - to name a few.

rafaelcaricio:
We don't have an single place to discuss cross-issues topics. For now every issue have it's own context of discussion. Maybe it's time for a Connexion Slack community!?

rafaelcaricio:
Agree, I would vote for a branch `dev-2.0.0` or something to streamline all the code for that release.

jmcs:
I created a [dev-2.0](https://github.com/zalando/connexion/tree/dev-2.0) branch. We can merge this PR there.

jmcs:
Superseded by #618 . Thanks for the PR.

PR #618, opened 2018-07-10
Remove vendored swagger-ui, add swagger-ui-bundle as optional pip extra

### LINKED ISSUE #406: Remove Swagger UI vendor code

dtkav:
Fixes #406 

- add swagger-ui-bundle package as a new optional dependency
- remove connexion/vendor/ directory


jmcs:
:+1:

PR #619, opened 2018-07-10
Connexion 2.0

jmcs:
Changes:
- Removed vendored swagger ui and added swagger-ui-bundle as an optional dependency.


dtkav:
Hey @jmcs, do you know if merging from a staging branch like this will squash the commits, and remove the contributor information? If it would be possible to keep big changes as separate commits and contributor info preserved, that would be nice (Whenever you are ready to merge).
I'm not sure if that would mean cherry-picking vs. merging or what, but it would mean a lot to me.

jmcs:
@dtkav both rebase and squash show the changes as commited by me but authored by you:
image

If you prefer I can create a merge commit instead.

czchen:
Hi,

I has tried commit 62582643cbfca65b0cf0d54755f81f0e1a7d1f82 with `swagger-ui-bundle==0.0.2`. The `base_path` does not work in UI.

For example, the following code add `my_api.yml` under `/v0`, and add swagger ui in `/v0/ui`. However, if I click `Try it out` and the `Execute` in swagger ui, the request is sent to API without `base_path` (`/v0` in this case).

```
app.add_api('my_api.yml', base_path='/v0', strict_validation=True, validate_responses=True)
```

dtkav:
Thanks @czchen for trying out dev-2.0 and finding that bug. Do you mind creating a ticket instead of commenting on here? I'll take a look at it as soon as I can.

jmcs:
I see this as potential blockers:
- https://github.com/zalando/connexion/issues/729
- https://github.com/zalando/connexion/pull/732

dtkav:
AFAICT #732 is backwards compatible. Couldn't it be in the 2.1 release?
#735 should address #729

jmcs:
I released `2.0.0rc5` because I had forgotten #735 :man_facepalming:.

dtkav:
IMHO we should call it and do the release now. JWT support is backwards compatible, so it could be in 2.1.
I'm trying to practice patience, but at the same time this release cycle has been very long.
What do you think?

jmcs:
:+1:

marcinzaremba:
:+1:

PR #621, opened 2018-07-11
Support OpenAPI 3.0.0 specs

### LINKED ISSUE #420: OpenAPI v3 support

dtkav:
Fixes #420 

 - support openapi version 3

Architecture:
 - creates an AbstractOperation class, which defines the interface for what an Operation is
 - Implements this interface with Swagger2Operation and OpenAPIOperation classes
 - apis/abstract.py uses the correct operation class based on the version of the spec

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -8,23 +8,40 @@ from typing import AnyStr, List  # NOQA
 import jinja2
 import six
 import yaml
-
-from openapi_spec_validator import validate_v2_spec as validate_spec
 from openapi_spec_validator.exceptions import OpenAPIValidationError
-from swagger_ui_bundle import swagger_ui_2_path
+from six.moves.urllib.parse import urlparse
 
 from ..exceptions import InvalidSpecification, ResolverError
 from ..jsonref import resolve_refs
-from ..operations import Swagger2Operation
+from ..operations import OpenAPIOperation, Swagger2Operation
 from ..options import ConnexionOptions
 from ..resolver import Resolver
 from ..utils import Jsonifier
 
 MODULE_PATH = pathlib.Path(__file__).absolute().parent.parent
-SWAGGER_UI_PATH = swagger_ui_2_path
 SWAGGER_UI_URL = 'ui'
+NO_SPEC_VERSION_ERR_MSG = """Unable to get the spec version.
+You are missing either '"swagger": "2.0"' or '"openapi": "3.0.0"'
+from the top level of your spec."""
 
 logger = logging.getLogger('connexion.apis.abstract')
+
+
+def _get_spec_version(spec):
+    try:
+        version_string = spec.get('openapi') or spec.get('swagger')
+    except AttributeError:
+        raise InvalidSpecification(NO_SPEC_VERSION_ERR_MSG)
+    if version_string is None:
+        raise InvalidSpecification(NO_SPEC_VERSION_ERR_MSG)
+    try:
+        version_tuple = tuple(map(int, version_string.split(".")))
+    except TypeError:
+        err = ('Unable to convert version string to semantic version tuple: '
+               '{version_string}.')
+        err = err.format(version_string=version_string)
+        raise InvalidSpecification(err)
+    return version_tuple
 
 
 class AbstractAPIMeta(abc.ABCMeta):
@@ -71,13 +88,6 @@ class AbstractAPI(object):
         self.validator_map = validator_map
         self.resolver_error_handler = resolver_error_handler
 
-        self.options = ConnexionOptions(options)
-
-        logger.debug('Options Loaded',
-                     extra={'swagger_ui': self.options.openapi_console_ui_available,
-                            'swagger_path': self.options.openapi_console_ui_from_dir,
-                            'swagger_url': self.options.openapi_console_ui_path})
-
         logger.debug('Loading specification: %s', specification,
                      extra={'swagger_yaml': specification,
                             'base_path': base_path,
@@ -92,6 +102,15 @@ class AbstractAPI(object):
 
         self.specification = compatibility_layer(self.specification)
         logger.debug('Read specification', extra={'spec': self.specification})
+
+        self.spec_version = _get_spec_version(self.specification)
+
+        self.options = ConnexionOptions(options, oas_version=self.spec_version)
+
+        logger.debug('Options Loaded',
+                     extra={'swagger_ui': self.options.openapi_console_ui_available,
+                            'swagger_path': self.options.openapi_console_ui_from_dir,
+                            'swagger_url': self.options.openapi_console_ui_path})
 
         # Avoid validator having ability to modify specification
         self.raw_spec = copy.deepcopy(self.specification)
@@ -115,6 +134,7 @@ class AbstractAPI(object):
         logger.debug('Security Definitions: %s', self.security_definitions)
 
         self.definitions = self.specification.get('definitions', {})
+        self.components = self.specification.get('components', {})
         self.parameter_definitions = self.specification.get('parameters', {})
         self.response_definitions = self.specification.get('responses', {})
 
@@ -144,6 +164,10 @@ class AbstractAPI(object):
             self.add_auth_on_not_found(self.security, self.security_definitions)
 
     def _validate_spec(self, spec):
+        if self.spec_version < (3, 0, 0):
+            from openapi_spec_validator import validate_v2_spec as validate_spec
+        else:
+            from openapi_spec_validator import validate_v3_spec as validate_spec
         try:
             validate_spec(spec)
         except OpenAPIValidationError as e:
@@ -152,10 +176,17 @@ class AbstractAPI(object):
     def _set_base_path(self, base_path):
         # type: (AnyStr) -> None
         if base_path is None:
-            self.base_path = canonical_base_path(self.specification.get('basePath', ''))
-        else:
-            self.base_path = canonical_base_path(base_path)
-            self.specification['basePath'] = base_path
+            if self.spec_version < (3, 0, 0):
+                base_path = self.specification.get('basePath', '')
+            else:
+                # TODO variable subsitution in urls for oas3
+                servers = self.specification.get('servers', [])
+                for server in servers:
+                    # TODO how to handle multiple servers in an
+                    #      oas3 spec with different paths?
+                    base_path = urlparse(server['url']).path
+                    break
+        self.base_path = canonical_base_path(base_path)
 
     @abc.abstractmethod
     def add_openapi_json(self):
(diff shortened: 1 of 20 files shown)


tests/decorators/test_parameter.py
```diff
@@ -16,14 +16,20 @@ def test_injection(monkeypatch):
...
+
+    parameter_to_arg(Op(), handler)(request)
 
     func.assert_called_with(p1='123')
 
 
 def test_query_sanitazion(query_sanitazion):
     app_client = query_sanitazion.app.test_client()
-    l = LogCapture()
+    #l = LogCapture()
```
dtkav: This tests is pretty nasty - I'm thinking of just removing it.

tests/api/test_parameters.py
```diff
@@ -280,11 +290,13 @@ def test_nullable_parameter(simple_app):
         '/v1.0/nullable-parameters?time_start={}'.format(time_start))
     assert json.loads(resp.data.decode('utf-8', 'replace')) == time_start
 
-    resp = app_client.post('/v1.0/nullable-parameters', data={"post_param": 'None'})
-    assert json.loads(resp.data.decode('utf-8', 'replace')) == 'it was None'
+    # XXX this has not valid OAS3 equivalent as far as I can tell
```
dtkav: I could use help figuring out the correct behavior here.
It seems like there should be something that iterates through form parameters and replaces `"null"` or `"None"` with `None`? Does this need to be recursive? Is it only for form encoded request bodies? Can form parameters have recursive definitions, or are they guaranteed to be flat?

connexion/operations/openapi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        self._definitions_map = {
+            'components': {
+                'schemas': component_get('schemas'),
+                'requestBodies': component_get('requestBodies'),
+                'parameters': component_get('parameters'),
+                'securitySchemes': component_get('securitySchemes'),
+                'responses': component_get('responses'),
+                'headers': component_get('headers'),
+            }
```
rafaelcaricio: It seems to be missing the "examples", "links", and "callbacks". All those are important components to be reachable within Connexion.

connexion/operations/openapi.py
```diff
@@ -0,0 +1,376 @@
...
+                    ref = self._resolve_reference(resp)
+                    del resp['$ref']
+                    resp = ref
+
+                content = resp.get("content", {})
+                for mimetype, resp in content.items():
+                    # check components/examples
+                    examples = resp.get("examples", {})
+                    for _, example in examples.items():
+                        example = self._resolve_reference(example)
```
rafaelcaricio: What is the intention here? the `examples` variable seems to be overridden every time with no use, even right after the loop on line 132.

connexion/operations/openapi.py
dtkav: Here's a library that resolves json references lazily https://github.com/gazpachoking/jsonref
what do you think?

rafaelcaricio: Sure, I agree we should indeed skip the additional features that only apply to OAS3.

 I would vote for eager evaluation of references. But due the case of loops we need to be able to handle that with some lazy reference resolver. Maybe a custom `dict` would do. Seems like the library you mentioned here does that. Sounds good to me! 👍I was trying the code here with some complex spec I have, some fixes here: https://github.com/dtkav/connexion/commit/ea54567150cafce39b80968b030a6ef2b4964cff feel free to cherry-pick.

dtkav: I'm very confused at what/where https://github.com/dtkav/connexion/commit/ea54567150cafce39b80968b030a6ef2b4964cff actually is. It seems to be a dangling commit, and I can't pull it down locally to cherry-pick. Maybe you committed that against my branch, and then I force-pushed it away? Any ideas on how to recover it?

dtkav: (thanks for the fixes by the way)

rafaelcaricio: Oh, sorry. This is the correct place https://github.com/rafaelcaricio/connexion/commit/ea54567150cafce39b80968b030a6ef2b4964cff

It currently lives in this branch https://github.com/rafaelcaricio/connexion/commits/oas3-support

dtkav: Thanks! I've cherry-picked your changes.

dtkav:
Another architecture improvement that would simplify the oas3 work in #627

connexion/operations/openapi.py
dtkav: hey @rafaelcaricio , any chance you can provide a test case for raw examples?
The spec says that examples can can be either a string primitive or an object, so it looks like it needs to support both.
 https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#examples-object

dtkav: If you have an example that is json, then it makes sense to have it be an object in the spec (this way it can be validated). If not, then connexion won't validate the object (if it's xml or something), and send it to the client as is.

I don't really understand the use case of having raw json. It seems significantly worse, in that your editor can't lint it, and the validator can't validate it. What am I missing?

connexion/decorators/validation.py
```diff
@@ -128,6 +135,33 @@ def __init__(self, schema, consumes, api, is_null_value_valid=False, validator=N
         self.api = api
         self.strict_validation = strict_validation
 
+    def validate_formdata_parameter_list(self, request):
+        request_params = request.form.keys()
+        spec_params = self.schema.get('properties', {}).keys()
+        return validate_parameter_list(request_params, spec_params)
+
+    def coerce_types(self, param_schema, value, parameter_name=None):
```
dtkav: I think I can refactor this code out a bit.

connexion/decorators/validation.py
```diff
@@ -90,10 +94,15 @@ def extend_with_nullable_support(validator_class):
...
         null_properties = {}
         for property_, subschema in six.iteritems(properties):
-            if isinstance(instance, collections.Iterable) and \
-                    property_ in instance and \
-                    instance[property_] is None and \
-                    subschema.get('x-nullable') is True:
+            # check equal to True, because x-nullable is an extension
+            # whereas nullable value is validated by 3.0 spec
+            nullable = (subschema.get('x-nullable') is True or
+                        subschema.get('nullable'))
```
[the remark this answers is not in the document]
JuxhinDB: Approving either way as it's a very minor issue. :-)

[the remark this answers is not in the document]
dtkav: Yeah, that's what I was trying to get across with the comment - the 3.0 spec validation will enforce that `nullable` is a bool, but `x-nullable` is an spec extension, so the type will never be validated.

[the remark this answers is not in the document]
JuxhinDB: Interesting, thanks for clarifying!

connexion/apis/abstract.py
```diff
@@ -24,6 +24,21 @@
 logger = logging.getLogger('connexion.apis.abstract')
 
 
+def _get_spec_version(spec):
```
ioggstream: raise InvalidSpecification('Invalid Spec Version: %r' % version_string)

connexion/apis/abstract.py
```diff
@@ -24,6 +24,21 @@
...
+        version_string = spec.get('openapi') or spec.get('swagger')
+    except AttributeError:
+        raise InvalidSpecification('Unable to get spec version')
+    if version_string is None:
+        raise InvalidSpecification('Unable to get spec version')
+    try:
+        version_tuple = tuple(map(int, version_string.split(".")))
+    except TypeError:
+        # unable to convert to semver tuple
+        raise InvalidSpecification('Invalid Spec Version')
```
ioggstream: raise InvalidSpecification('Invalid Spec Version: %r' % version_string)

connexion/apis/abstract.py
```diff
@@ -150,6 +157,13 @@ def _set_base_path(self, base_path):
         # type: (AnyStr) -> None
         if base_path is None:
             self.base_path = canonical_base_path(self.specification.get('basePath', ''))
+            if self.spec_version >= (3, 0, 0):
```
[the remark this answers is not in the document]
dtkav: Sorry, I'm not sure what you are suggesting here. Could you elaborate?
Technically this version check doesn't actually need to be here (the `servers` field won't exist in any valid swagger 2 specs), but it is more explicit. I'm trying not to just casually mix swagger2/openapi3 specific code. Most of that separation happens in the Swagger2Operation/OpenAPIOperation classes, but this is one of the few places in AbstractAPI that needs to be "version aware" in order to maintain the current API.

connexion/apis/abstract.py
```diff
@@ -281,6 +306,20 @@ def load_spec_from_file(self, arguments, specification):
             swagger_string = jinja2.Template(swagger_template).render(**arguments)
             return yaml.safe_load(swagger_string)  # type: dict
 
+    def _get_spec_version(self, spec):
+        try:
+            version_string = spec.get('openapi') or spec.get('swagger')
+        except AttributeError:
+            raise InvalidSpecification('Unable to get spec version')
```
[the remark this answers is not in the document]
dtkav: I really disagree with this. If we get this exception, then it means we couldn't find either of `openapi` / `swagger` keys in the spec. Printing the whole spec here may dump several pages of text to their terminal, and it won't be useful because it won't show them the _absence of the required field_.
I'll happily make this error more verbose though - let me know if that would address your concern.

connexion/decorators/uri_parsing.py
```diff
@@ -120,14 +127,72 @@ def coerce_dict(md):
...
+
+    def resolve_form(self, body):
+        if self._body_schema is None or self._body_schema.get('type') != 'object':
+            return body
+        for k in body:
+            encoding = self._body_encoding.get(k, {"style": "form"})
+            defn = self.form_defns.get(k, {})
+            # TODO support more form encoding styles
+            body[k] = self._resolve_param_duplicates(body[k], encoding)
+            if defn and defn["type"] == "array":
```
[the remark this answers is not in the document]
dtkav: The spec validation ensures that the type field will exist. I don't want to silently hide a KeyError here and continue. It should be impossible, and would just fail elsewhere in a potentially more confusing way if we were to continue. Using `get()` here would violate the fail-fast principle.

dtkav:
FYI the Codacy problems are due to the use of "assert" in the new unit tests. I've added a separate PR to remove those codacy checks.
I'm going to to ahead and merge this (yay!), but it will only be in dev-2.0, so I'm happy to continue to address suggestions and fixes there. We'll make sure that branch is rock solid cutting the next major release.

connexion/decorators/uri_parsing.py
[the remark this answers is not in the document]
ioggstream: Ok. It wasn't clear to me that failing was in the scope.
I'd add a comment to explicit that.

PR #623, opened 2018-07-11
Remove old_style_options

dtkav:
 - retire old_style_options

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,10 @@ Other Sources/Mentions
 - Connexion listed on Swagger_'s website
 - Blog post: `Crafting effective Microservices in Python`_
 
+New in Connexion 2.0:
+---------------------
+- App and Api options must be provided through the "options" argument (``old_style_options`` have been removed).
+
 How to Use
 ==========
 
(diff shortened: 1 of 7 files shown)


jmcs:
Can you adapt the tests and put a note in the readme (make a "Backward Incompatibilities in Connexion 2" section or something similar)?

dtkav:
I can't figure out why code coverage is down. Did I inadvertently disable coroutines somehow?

dtkav:
maybe related to #600 , @dutradda any thoughts?

jmcs:
You removed 42 lines and added 29 (4 of them in the Readme) so there are less covered lines.

jmcs:
👍

jmcs:
Thanks for the PR

dtkav:
What surprised me was a %73 drop in the coverage of `coroutine_wrappers.py`.
Here's the coveralls build link: https://coveralls.io/builds/17946733

dutradda:
@dtkav 

I don't think this is related to #600, because that issue is occurring just with "async def"  functions.

I think that the response decorator only uses the old_style_options on coroutine functions, so when you removed it, the validate_response option is not passed anymore, so the code is never running.

I will take a look on that.

PR #625, opened 2018-07-12
Minor change in readme about non-required parameters

positron96:
I think "non-named" word was ill-placed here (at least I did not understand what it meant). I tried to clarify this better and propose a solution to not get a exception.

Slightly changed wording in README


hjacobs:
:+1:

Issue #624, opened 2018-07-12
Idea: replace custom json reference resolver with off-the-shelf one


dtkav:
Some of the code for resolving references if fairly custom, and kind of nasty (esp. in #621  the openapi3 branch)
I wanted to float the idea of using something like https://github.com/gazpachoking/jsonref 
It will lazily resolve the references as they are used.

I'm happy to put up a demo - let me know what you think.

rafaelcaricio:
I think this is a great suggestion. We should be try to use ready components to put together Connexion features and bring in maintainable pieces together. This improves community engagement, cleaner design, and _less code to maintain_ (yay!).

jmcs:
I also agree with using libraries were we can.

dtkav:
Here's a branch with jsonref https://github.com/dtkav/connexion/tree/jsonref
Playing around with it a little bit, I don't think we actually want lazy evaluation. Also, the tests run _really slowly_.

Each resolve_reference call is doing two things:
1. Resolving json references at the provided "view" into the spec
2. Attaching definitions to that view["schema"] to be passed to the json validator

Because the json validator can already follow recursive references, I think we just need to resolve the references once (not recursively) early on, and then continue to attach the definitions for the various "views" into the spec (ex. body_schema).

I'll keep playing around, and hopefully put up a PR for discussion tonight, or early next week.

rafaelcaricio:
I think we need to investigate why it is so slow. I will try to take a look at this branch you setup to investigate.

dtkav:
I've included a PR that leverages the jsonschema RefResolver.
After playing with the lazy reference resolver, I don't think it's a great idea anymore.

This above PR will instead resolve the spec early on (right after validation), but still provides recursive schema validation by allowing definitions to be "attached" to a schema, providing a "view" into the spec that is self-contained.

An example of a "view" would be something like a body schema.

dtkav:
hey @rafaelcaricio or @jmcs can you take a look at #627 ? I think that's the last architectural change that I can think of to make the oas3 branch cleaner, and then I'll probably start looking to merge that. Thanks!

PR #627, opened 2018-07-17
resolve json references earlier, generically

dtkav:
relates to #624 

This branch factors out reference resolution, and applies it to the spec immediately after it is validated.
Previously this behavior was (1) in operation.py, and (2) tied to the swagger spec implementation.

 - make reference resolver general purpose
 - resolve references eagerly in `apis/abstract.py` once instead of many times within `operation.py`

Note that the resolve_references function had two main features:
1. resolving references
2. attaching the definitions for use by the validator (this was useful for recursive definitions).

There is now a `with_definitions()` function that attached the definitions to a schema, to be used with the jsonschema validator.
This allows for both eager evaluation, and recursive definitions.

### DIFF
--- a/connexion/jsonref.py
+++ b/connexion/jsonref.py
@@ -4,38 +4,21 @@ from copy import deepcopy
 from jsonschema import RefResolver
 from jsonschema.exceptions import RefResolutionError  # noqa
 
-from .utils import deep_get
 
-
-def resolve_refs(spec, defns=None, store=None):
+def resolve_refs(spec, store=None):
     """
     Resolve JSON references like {"$ref": <some URI>} in a spec.
-    Optionally takes a defns dictionary to look up references.
-    If no defns dictionary is specified, references will be resolved within
-    the spec, or resolve remotely from the web. Passing defns is useful
-    if you have a namespace collision between your spec and definitions.
     Optionally takes a store, which is a mapping from reference URLs to a
     dereferenced objects. Prepopulating the store can avoid network calls.
     """
     spec = deepcopy(spec)
     store = store or {}
-    if defns:
-        defns = deepcopy(defns)
     resolver = RefResolver('', spec, store)
 
     def _do_resolve(node):
         if isinstance(node, collections.Mapping) and '$ref' in node:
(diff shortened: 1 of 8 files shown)


connexion/jsonref.py
```diff
@@ -0,0 +1,48 @@
...
+        defns = deepcopy(defns)
+    resolver = RefResolver('', spec, store)
+
+    def _do_resolve(node):
+        if isinstance(node, collections.Mapping) and '$ref' in node:
+            path = node['$ref'][2:].split("/")
+            try:
+                # resolve known references
+                store = defns if defns else spec
+                node.update(deep_get(store, path))
```
rafaelcaricio: This method `deep_get` needs to be able to resolve JSON references. In it's current form it can resolve basic paths, but there's some specific scenarios we need to take in consideration. We can simply use the `RefResolver` from jsonschema lib, no?

```python
node.update(resolver.resolve_from_url(path))
```

I took a look at the source code and seems like it can resolve local refs too: https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L669-L704

rafaelcaricio: I think it would need to receive the full path starting from `#/`

dtkav: note that on KeyError it will fall back to the jsonschema reference resolver.
I can remove the deep_get functionality, I'm realizing now that I kept it so that you could resolve things in case of a namespace conflict, but this only happens when resolving operations (not entire specs).

dtkav: I appreciate the review @rafaelcaricio . thanks!

dtkav: hmm.. actually removing the deep_get resolution didn't work because jsonschema RefResolver will get break on recursive definitions. I think the old behavior is actually ideal (try deep_get resolution, fall back to RefResolver).

rafaelcaricio: Hm.. ok, I haven't considered this scenario. Thank you for taking the time to work on this. I really appreciate. 👍

rafaelcaricio:
👍

Issue #593, opened 2018-06-04
Response validator fails to determine mimetype when several are provided in spec.


positron96:
### Description

I have an endpoint that produces either `application/json` or `text/csv`, depending on Accept header. 

This is reflected in swagger file as 
```
/endpoint:
    get:
      summary: get all calibrations from server
      operationId: endpoint
      produces:
        - application/json
        - text/csv
      responses:
        '200':
          description: 'OK'
          schema:
            type: array
            items:
              $ref: '#/definitions/CalibrationData'  ## this in an array of records of 3 floats
          examples:
            text/csv: |-
                  1, 0.12, -12
                  2, 1, 2
            aplication/json:
              Series: 1
              a: 0.12
              b: -12
      parameters: []
```

I have turned on response validation with ```app.add_api('api.yaml', validate_responses=True)```

### Expected behaviour

A correct response to a request with `Accept: text/csv` header should be sent to browser. The ResponseValidator should pass text/csv as is.

### Actual behaviour

Validator produces this error:

```
ERROR:connexion.decorators.validation:http://localhost:8080/api/endpoint validation error: '123,0.345000,1.234000\n4,0.345000,1.234000' is not of type 'array'

Failed validating 'type' in schema: 
...
On instance:
    '123,0.345000,1.234000\n4,0.345000,1.234000'

```

Seems that validator does not favour additional mimetypes when they are present.

### Additional info:

Output of the commands:

- `python --version`

Python 3.5.2

- `pip show connexion | grep "^Version\:"`

Version: 1.4.2

positron96:
It seems to me that there are several causes of this problem:
 * ResponseValidator stores only one mimetype from spec and does not check actual response mimetype
 * Operation.get_mimetype returns only 1 mimetype from `produces` list. If there are several entries, they are discarded and json is returned (https://github.com/zalando/connexion/blob/master/connexion/operation.py#L333)

RobbeSneyders:
Fixed in https://github.com/spec-first/connexion/issues/590

PR #629, opened 2018-07-18
Fix body validator different error codes for bad Content-Types.

### LINKED ISSUE #628: Request validator passes www-multipart-data when schema is $ref'ed
### Description

Request validator passes a request with Content-type: application/x-www-form-urlencoded when spec only accepts JSON and schema is $ref'erenced.

Interestingly, when schema is not $ref'ed, validation fails.

### Expected behaviour

Validator should reject such request, it contains bad content-type.

### Actual behaviour

Validator passes execution to application code, mayhem ensues.

### Steps to reproduce

Minimal files to reproduce:

#### Swagger spec:

```
swagger: '2.0'

info:
  description: a
  version: "0"
  title: a
    
host: localhost

paths:
            
  /test:
    post:
      operationId: api.test
      consumes:
      - application/json
      parameters:
        - name: body
          description: Just a testing parameter in the body
          in: body
          required: true
          schema:
            type: object
            properties:
              body1:
                type: string
              body2:
                type: string
      responses:
        200:
          description: OK
          
  /test2:
    post:
      operationId: api.test2
      consumes:
      - application/json
      parameters:
        - name: $body
          description: Just a testing parameter in the body
          in: body
          required: true
          schema:
            $ref: '#/definitions/Test'
            
      responses:
        200:
          description: OK
      

definitions:

  Test:
    properties:
      body1:
        type: string
      body2:
        type: string

```

#### API code:
```
#!/usr/bin/env python3

import flask
from flask import request, Response

import connexion
    
def test(body):
    print('body is:', body)
    return "OK"
    
def test2(body):
    print('body is:', body)
    return "OK"


app = connexion.FlaskApp(__name__)
app.add_api('api.yaml', validate_responses=True)

app.run(port=8080, debug=True)
```

#### Test code:
```
#!/usr/bin/env python3

import requests

r = requests.post( "http://localhost:8080/test", headers={'Content-Type':'application/x-www-form-urlencoded'}, data="a=1" )
print( r.status_code ) ## returns 4xx error, somewhat correctly

r = requests.post( "http://localhost:8080/test2", headers={'Content-Type':'application/x-www-form-urlencoded'}, data="a=1" )
print( r.status_code ) ## returns 200

```

#### Requirements:
```
flask
connexion
requests
```

### Additional info:

Output of the commands:

- `python --version`  Python 3.5.2
- `pip show connexion | grep "^Version\:"`  Version: 1.4.2

positron96:
Fixes #628 .

 - Added a test for this bug.
 - Fixed it by checking for non-empty HTTP POST payload by considering request.body, request.form and request.files (only request.body was checked)

### DIFF
--- a/.zappr.yaml
+++ b/.zappr.yaml
@@ -10,5 +10,3 @@ approvals:
       # a public zalando org member
       # (any org in here counts)
       - zalando
-    # OR a collaborator of the repo
-    collaborators: true
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -12,6 +12,8 @@ logger = logging.getLogger('connexion.decorators.uri_parsing')
 
 @six.add_metaclass(abc.ABCMeta)
 class AbstractURIParser(BaseDecorator):
+    parsable_parameters = ["query", "path"]
+
     def __init__(self, param_defns):
         """
         a URI parser is initialized with parameter definitions.
@@ -25,7 +27,7 @@ class AbstractURIParser(BaseDecorator):
         """
         self._param_defns = {p["name"]: p
                              for p in param_defns
-                             if p["in"] in ["query", "path"]}
+                             if p["in"] in self.parsable_parameters}
 
     @abc.abstractproperty
     def param_defns(self):
@@ -97,18 +99,20 @@ class AbstractURIParser(BaseDecorator):
 
         @functools.wraps(function)
         def wrapper(request):
+            def coerce_dict(md):
+                """ MultiDict -> dict of lists
+                """
+                try:
+                    return md.to_dict(flat=False)
+                except AttributeError:
+                    return dict(md.items())
 
-            try:
-                query = request.query.to_dict(flat=False)
-            except AttributeError:
-                query = dict(request.query.items())
-
-            try:
-                path_params = request.path_params.to_dict(flat=False)
-            except AttributeError:
-                path_params = dict(request.path_params.items())
+            query = coerce_dict(request.query)
+            path_params = coerce_dict(request.path_params)
+            form = coerce_dict(request.form)
 
             request.query = self.resolve_params(query, resolve_duplicates=True)
+            request.form = self.resolve_params(form, resolve_duplicates=True)
             request.path_params = self.resolve_params(path_params)
             response = function(request)
             return response
@@ -121,6 +125,7 @@ class Swagger2URIParser(AbstractURIParser):
     Adheres to the Swagger2 spec,
     Assumes the the last defined query parameter should be used.
     """
+    parsable_parameters = ["query", "path", "formData"]
 
     @property
     def param_defns(self):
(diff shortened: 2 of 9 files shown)


dtkav:
nice, I noticed this too when working OpenAPI3 support.
Since this may break existing clients, it might make sense to stage it for the 2.0 release.
Just my 2c though.

dtkav:
my bad @positron96 , I think i am confusing it with a different issue where connexion accepts form data even if the mimetype is wrong.

positron96:
That sounds serious. Is there a description of the bug?

dtkav:
I dug into this a little bit to refresh my memory. I couldn't remember if it was a problem with my branch, or also on master.
It turns out that form parameters won't be rejected based on the "consumes" part of the operation.
For example, the spec below still accepts the form parameters, even though `consumes = ["application/json"]`.
```yaml
  /test-formData-param:
    post:
      summary: Test formData parameter
      operationId: fakeapi.hello.test_formdata_param
      consumes:
        - application/json  # connexion assumes application/json if this is blank
      # - application/x-www-form-urlencoded
      parameters:
        - name: formData
          type: string
          in: formData
          required: true
      responses:
        200:
          description: OK
```
The client still sends the correct content type alongside the form data, but the spec doesn't specify that the endpoint consumes form content types, and connexion isn't strict - it let's the form data through.

It's an oddity of the swagger 2.0 spec that you can specify `in: formData` separately from the `consumes` block. This is fixed in the 3.0 spec with the introduction of `requestBody`. The existing behavior may actually make sense for the 2.0 spec - I'm not sure. If you were to reject formData parameter if the `consumes` block didn't match, that should/would probably be in a spec validator. What do you think?

You can reproduce this by running:
```bash
cd tests
PYTHONPATH=$PWD connexion run fixturesimple/swagger.yaml -p 9090
# in another terminal...
curl -X POST -F "formData=somedata" http://localhost:9090/v1.0/test-formData-param
```

Either way, your change looks good to me!

positron96:
@dtkav Yes, I've reproduced this behavior. It seems that Swagger 2.0 vagueness in this point caused it since it's not clear what to do in this case. Probably keeping formData is OK.

Regarding MIME types, #593 is more serious for me, I'll look into it sometime.

connexion/decorators/validation.py
```diff
@@ -101,7 +101,8 @@ def wrapper(request):
             if all_json(self.consumes):
                 data = request.json
 
-                if data is None and len(request.body) > 0 and not self.is_null_value_valid:
+                empty_body = len(request.body) == 0 and len(request.form) == 0 and len(request.files) == 0
```
jmcs: I think `not(request.body or request.form or request.files)` would be more readable.

jmcs:
@positron96 I agree with @dtkav this should target the `dev-2.0` branch (and be included in the changes in the README).

dtkav:
@jmcs It turns out I was conflating this change with some other behavior (elaborated on above).
FWIW I now think this change is relatively benign.

connexion/decorators/validation.py
positron96: yes, it definitely will

positron96:
BTW, I've changed the code as requested.

PR #638, opened 2018-07-25
parse form parameters in Swagger2URIParser

### LINKED ISSUE #634: FormData array parameters are not split correctly
It seems that formData array parameters are not being split correctly in the 1.5.0 release. The issue is not present in 1.4.2.

The string is split on every character instead of each `,`. It looks like this issue is related to the changes in  #613 

The following parameter definition was used:

```yaml
    - name: data
      in: formData
      required: true
      type: array
          collectionFormat: csv
          items:
              type: string
```

### Additional info:
- `python --version`: 3.6.5
- Connexion version 1.5.0

dtkav:
Fixes #634 

 - add formData to parsed parameters list in Swagger2URIParser


dtkav:
whoops...
thanks @jmcs

jmcs:
:+1:

PR #639, opened 2018-07-25
Refactor Operation into AbstractOperation and Swagger2Operation

dtkav:
Relates to #420 

In an attempt to break up the OpenAPI3 PR, this includes all of the refactor that went into that branch, but leaves out other openapi3 support.

 - Establish interface for what an AbstractOperation does
 - Implements interface with Swagger2Operation
 - Defaults to Swagger2URIParser (*not* AlwaysMultiURIParser)

Plan for OpenAPI integration:
Once this is merged, the OAS3 branch should basically contain
1. The OpenAPIOperation class
2. The OpenAPIURIParser class
3. requestBody form validation
4. modifications to apis/abstract to get the spec version, and instantiate the correct operation class


dtkav:
Landing #641 should make this diff a little smaller.
I'll remove the WIP from this PR once that's in.

jmcs:
@dtkav can you resolve the conflicts?

tests/api/test_bootstrap.py
```diff
@@ -41,10 +41,10 @@ def test_app_with_different_server_option(simple_api_spec_dir, spec):
 
 
 def test_app_with_different_uri_parser(simple_api_spec_dir):
-    from connexion.decorators.uri_parsing import Swagger2URIParser
```
dtkav: I'll factor this out into a separate PR

dtkav:
#653 will make this smaller (factors out switch to openapi_spec_validator)

connexion/operations/swagger2.py
```diff
@@ -0,0 +1,291 @@
+import logging
```
JuxhinDB: Curious, was this just stripped out from an older file into its own file to ensure backwards compatibility?

jmcs:
:+1:

connexion/operations/swagger2.py
dtkav: Yup, the swagger2.py file is for backwards compatibility with the swagger2 spec. The idea with this PR is to create an Abstract Base Class (abc), which is essentially the interface that describe what an "Operation" is. Previously we only had one implementation of this interface (swagger 2 Operation), so it wasn't well defined. Now, after this refactor, the OpenAPIOperation can implement the exact same interface as a Swagger2Operation. Together they encapsulate the differences between the Swagger2 spec and the OpenAPI3 spec. Hope that makes sense.

PR #641, opened 2018-07-26
parameterize tests

dtkav:
related to #420 
This includes some of the changes to tests from #639 
I've broken them out because they could be reviewed separately.

 - add test parameterization for different spec files (i.e. swagger.yaml, openapi.yaml)


dtkav:
@rafaelcaricio do you mind having a look at this one?
It should be a `nop`, but will reduce the number of lines changed in future oas3 diffs.

jmcs:
👍

PR #642, opened 2018-07-26
Add x-nullable support for body validators

### LINKED ISSUE #439: x-nullable support in response body validation
### Description

I want to x-nullable support in response body validation.

### Expected behaviour

supporting  response object incuding null value as property. such as this person schema.

```yaml
definitions:
   person:
     properties:
       name:
         type: string
       nickname:
         type: string
          x-nullable: true
```

### Actual behaviour

validation error (input data is null and schema has x-nullable=true).  connexion's x-nullable support is not comprehensive.

### Additional info:

supporting x-nullable with jsonschema library(used by connexion) is  like a below code.
https://gist.github.com/podhmo/2259fa63b58f6bbc26d618976b7662db

dmksnnk:
Closes #439 

Add `x-nullable` support in properties by adding property validator to default `Draft4Validator`. It skips null-property with `"x-nullable": true` and then adds it back.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -5,7 +5,7 @@ import logging
 import sys
 
 import six
-from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
+from jsonschema import Draft4Validator, ValidationError, draft4_format_checker, validators
 from werkzeug import FileStorage
 
 from ..exceptions import ExtraParameterProblem
@@ -73,6 +73,36 @@ def validate_parameter_list(request_params, spec_params):
     return request_params.difference(spec_params)
 
 
+def extend_with_nullable_support(validator_class):
+    """Add support for null values in body.
+
+    It adds property validator to given validator_class.
+
+    :param validator_class: validator to add nullable support
+    :type validator_class: jsonschema.IValidator
+    :return: new validator with added nullable support in properties
+    :rtype: jsonschema.IValidator
+    """
+    validate_properties = validator_class.VALIDATORS['properties']
+
+    def nullable_support(validator, properties, instance, schema):
+        null_properties = {}
+        for property_, subschema in six.iteritems(properties):
+            if property_ in instance and \
+                    instance[property_] is None and \
+                    subschema.get('x-nullable') is True:
+                # exclude from following validation
+                null_properties[property_] = instance.pop(property_)
+        for error in validate_properties(validator, properties, instance, schema):
+            yield error
+        # add null properties back
+        instance.update(null_properties)
+    return validators.extend(validator_class, {'properties': nullable_support})
+
+
+Draft4ValidatorSupportNullable = extend_with_nullable_support(Draft4Validator)
+
+
 class RequestBodyValidator(object):
     def __init__(self, schema, consumes, api, is_null_value_valid=False, validator=None):
         """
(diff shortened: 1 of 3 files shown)


connexion/decorators/validation.py
```diff
@@ -73,6 +73,36 @@ def validate_parameter_list(request_params, spec_params):
...
+    :param validator_class: validator to add nullable support
+    :type validator_class: jsonschema.IValidator
+    :return: new validator with added nullable support in properties
+    :rtype: jsonschema.IValidator
+    """
+    validate_properties = validator_class.VALIDATORS['properties']
+
+    def nullable_support(validator, properties, instance, schema):
+        null_properties = {}
+        for property_, subschema in six.iteritems(properties):
```
dtkav: It looks like this is breaking the build by not throwing the right exception when the function is given an integer as input.
See the travis build for the failure.

dtkav:
This looks really close - there are some minor changes needed to fix the build, plus a rebase on master. Thanks for putting this together!

tests/decorators/test_uri_parsing.py
```diff
@@ -1,5 +1,6 @@
-import pytest
 from werkzeug.datastructures import MultiDict
+
```
dtkav: Unfortunately this breaks the isort step in the build. See the end of the python 3.6 travis run for details.

dmksnnk:
Can you please check for Pypy build? It can't download Pypy binaries. Otherwise all other builds are passed.

dtkav:
hey @dmksnnk , sorry this fell off my radar. I kicked the build - seems like the problem was with travis-ci.
You code looks great to me, and I'd love to use this in the dev-2.0 branch for nullable support in openapi 3 validation.

dtkav:
Only zalando employees can merge PRs into master, so we'll have to wait for either @jmcs or @hjacobs to have a look.

I've merged it into `dev-2.0` for my own purposes, but I think it would make sense to merge it into master too (so it can get out sooner). I can sort out any rebase/merge conflicts that may come up on the `dev-2.0` branch.

Moving the code into another file is totally optional, I just wanted to raise the idea. I don't have a strong preference.

dtkav:
actually, @dmksnnk if you can rebase on master, that would help.

dtkav:
Hey @jmcs - This is looking good from my perspective. I recommend we merge this into master, what do you think?

jmcs:
:+1:

jankontrik:
Isn't this a bug ? Why x-nullable does not work on $ref ? For example:
```
changes:
      type: object
      properties:
        Bla:
          $ref: "#/definitions/Muhehe"
          **x-nullable: true**
      additionalProperties:
        $ref: "#/definitions/Muhehe"
        **x-nullable: true**
```

dmksnnk:
@jankontrik  you need to set `x-nullable: true` inside referenced object:
```
Muhehe:
    title: Muhehe
    type: object
    x-nullable: true
```

jankontrik:
Indeed I had that this way (x-nullable: true inside referenced object) but always ends with: 
`validation error: None is not of type 'object'`
So that's why I've tried to define x-nullable alongside $ref.

jankontrik:
I'm using python 3.6, connexion==1.5.2.

dmksnnk:
if you check [releases](https://github.com/zalando/connexion/releases), this feature was added for version 1.5.3

jankontrik:
@dmksnnk Yep. Sorry I know about which release. Of course I tried that on 1.5.3 but with the result above :(. Then I returned to 1.5.2 and forgot about the right version.

dmksnnk:
I'm investigating this, but have you considered using `{}` for empty object instead of `None/null`?

jankontrik:
Of course it is not good I absolutely agree. But this is some legacy stuff, no way to change.

jankontrik:
BTW the problem is always on additionalProperties when there is a x-nullable ref and None comes.

PR #644, opened 2018-07-27
remove ref resolution from mock code

dtkav:
This case is no longer possible since jsonref resolution happens much earlier in api/abstract.py

 - remove jsonref resolution code from mock code, tests


jmcs:
👍

PR #651, opened 2018-08-01
Rename Operation to Swagger2Operation

dtkav:
This is an attempt to shrink #639 to make it easier to review.
Many of the lines changed in #639 are purely because of the name change from `Operation` to `Swagger2Operation`.

The idea of this PR is to have this PR be _purely_ a name change with no changes in functionality at all.
Then, #639 can be rebased on this PR, and reviewed for functionality changes without the extra overhead of all of the name changes.

Let me know what you think.

 - Rename Operation to Swagger2Operation

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -131,7 +131,7 @@ class SecureOperation(object):
         return EndOfRequestLifecycleDecorator(self.api, self.get_mimetype())
 
 
-class Operation(SecureOperation):
+class Swagger2Operation(SecureOperation):
 
     """
     A single API operation on a path.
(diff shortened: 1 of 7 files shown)


connexion/operation.py
```diff
@@ -131,7 +131,7 @@ def _request_end_lifecycle_decorator(self):
         return EndOfRequestLifecycleDecorator(self.api, self.get_mimetype())
 
 
-class Operation(SecureOperation):
+class Swagger2Operation(SecureOperation):
```
jmcs: Can you add a `Operation = Swagger2Operation` in the end of the module in case someone is importing it?

jmcs:
:+1:

PR #656, opened 2018-08-07
Fix cli.py app options

dtkav:
cli.py currently passes old_style_options to the app_cls


dtkav:
@jmcs - quick fix here from removing old_style_options if you have a spare few minutes

PR #657, opened 2018-08-07
Accept only 👍 from Zalando Employees

jmcs:
Merges to master should only be allowed for members of the Zalando Organization


hjacobs:
:+1:

PR #659, opened 2018-08-08
update some names from swagger to openapi

dtkav:
another chunk of #639 

 - rename references to the "swagger json" to "openapi json"
 - ensure we serve up the spec as it exists on disk ("raw_spec"), not the resolved spec.

Note that the UI component is actually called "swagger-ui".
> ```
> The easiest way to understand the difference is:
> OpenAPI = Specification
> Swagger = Tools for implementing the specification
> ```

More information on naming here: https://swagger.io/blog/api-strategy/difference-between-swagger-and-openapi/

### DIFF
--- a/connexion/options.py
+++ b/connexion/options.py
@@ -52,14 +52,12 @@ class ConnexionOptions(object):
 
         Default: True
         """
-        serve_spec = self._options.get('serve_spec', False)
+        deprecated_option = self._options.get('swagger_json', True)
+        serve_spec = self._options.get('serve_spec', deprecated_option)
         if 'swagger_json' in self._options:
             deprecation_warning = ("The 'swagger_json' option is deprecated. "
                                    "Please use 'serve_spec' instead")
             logger.warning(deprecation_warning)
-            serve_spec = serve_spec or self._options.get('swagger_json')
-        # override if swagger UI is enabled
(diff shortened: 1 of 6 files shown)


dtkav:
my bad - should be fixed now.

jmcs:
:+1:

connexion/options.py
```diff
@@ -44,11 +48,10 @@ def openapi_spec_available(self):
         # type: () -> bool
         """
         Whether to make available the OpenAPI Specification under
-        `openapi_console_ui_path`/swagger.json path.
+        `openapi_spec_path`.
 
         Default: True
         """
-        # NOTE: Under OpenAPI v3 this should change to "/openapi.json"
```
dtkav: Thoughts on renaming the option? Backwards compatibility?

jmcs: We can rename it. Maybe using `swagger_json` as a fallback (with a deprecation warning) if `openapi_json` is not set.

dtkav: I'm signing off for the night - If you could please take a look at https://github.com/zalando/connexion/pull/656 also, that would be great - It should be a quick bugfix. Have a good evening.

jmcs:
:+1:

PR #671, opened 2018-08-31
Add validator kind to error logs extra in validators

bartekwojcicki:
To easily discern response and request validation errors logs
programmatically (e.g. in logging filter).


bartekwojcicki:
Use case: I'd like to be able to log response validation errors to sentry, but ignore request validation errors. Now there is no obvious way to do it. I've went with adding extra and not changing message or using separate loggers to avoid breaking existing logging configs of other connexion users.

Please let me know if you consider it worth merging.

jmcs:
👍

jmcs:
Thanks for the PR.

PR #672, opened 2018-09-06
Improve metrics reporting

### LINKED ISSUE #665: Metrics: report correct HTTP status code
### Description

Currently the [UWSGIMetricsCollector](https://github.com/zalando/connexion/blob/53908d30e566ccb81d73d3e27b461f407a8a8d31/connexion/decorators/metrics.py#L33) reporting will create a metrics key with status_code=500 for all exception thrown in the underlying call (BadRequest, Unauthorized, etc.)

### Expected behaviour
If an HTTPException is thrown in the underlying function, use the status code of the exception to report in the metrics. If an unknown exception is thrown map it to a specific "unkown exception" metrics key.

E.g.:
```json
"timers": {
    "connexion.response.401.GET.state.{id}": {
        "count": 2
    },
    "connexion.response.unknown-exception.GET.state.{id}": {
        "count": 2
    }
}
```

### Actual behaviour
**Any** exception is mapped to status code 500 as that is the default:
```json
"timers": {
    "connexion.response.500.GET.state.{id}": {
        "count": 2
    }
}
```

### Solution suggestion
Maybe something like this would help?
```python
from werkzeug.exceptions import HTTPException

@functools.wraps(function)
def wrapper(request):
    status = 'unknown-exception'
    start_time_s = time.time()
    try:
        response = function(request)
        status = response.status_code
    except HTTPException as http_e:
        status = str(http_e.code)
        raise http_e
    finally:
        end_time_s = time.time()
        delta_s = end_time_s - start_time_s
        delta_ms = delta_s * 1000
        key = '{status}.{suffix}'.format(status=status, suffix=self.key_suffix)
        uwsgi_metrics.timer(self.prefix, key, delta_ms)
    return response

### PR DESCRIPTION
**Pull request by gavronek**
Fixes #665 .

 - In case of an HttpException use the status code from the exception to report in the metrics

### DIFF
--- a/connexion/decorators/metrics.py
+++ b/connexion/decorators/metrics.py
@@ -32,7 +32,7 @@ class UWSGIMetricsCollector(object):
 
         @functools.wraps(function)
         def wrapper(request):
-            status = 'unknown-exception'
+            status = 500
             start_time_s = time.time()
             try:
                 response = function(request)
--- a/connexion/decorators/metrics.py
+++ b/connexion/decorators/metrics.py
@@ -2,6 +2,8 @@ import functools
 import os
 import time
 
+from werkzeug.exceptions import HTTPException
+
 try:
     import uwsgi_metrics
     HAS_UWSGI_METRICS = True  # pragma: no cover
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: connexion/decorators/metrics.py [hunk 1]
```diff
@@ -30,16 +32,19 @@ def __call__(self, function):
 
         @functools.wraps(function)
         def wrapper(request):
-            status_code = 500
+            status = 'unknown-exception'
```
dtkav: FWIW I can imagine this being a big pain for someone trying parse logs who expects the status to always be an integer.

dtkav: I think I'd still default this to 500, what do you think?

jmcs: I also expect it to be 500, since that's what makes sense from an end-2-end visibility point of view.

PR #679, opened 2018-09-13
Codacy skip assert warnings

dtkav:
Codacy issues warning when the `assert` keyword is used.

The concern from the bandit linter is that people might use `assert` in production code as a way of checking function inputs or something (no idea why anyone would ever do this).

In the case of connexion, it spews warnings any time you add a unit test! This is pretty annoying, and gives a false sense of a PR's code quality.

This PR disables that warning.


jmcs:
:+1:

dtkav:
I don't think this is actually working =(

PR #680, opened 2018-09-13
Codacy fixes

dtkav:
 - avoid using python keywords as variable names
 - use enumerate() instead of range(len(...))
 - convert to staticmethod when not using self (but still want to keep as method, not function)


jmcs:
:+1:

PR #682, opened 2018-09-14
WIP: rely on openapi-spec-validator to safely load and resolve yaml files

### LINKED ISSUE #681: [dev-2.0] RefResolver cannot resolve yaml
```
connexion/jsonref.py:from jsonschema import RefResolver
```

because it can't parse yaml schema files like https://opensource.zalando.com/problem/schema.yaml

I workaround'd it in the `jsonschema` module

```

+++ b/jsonschema/validators.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 import contextlib
 import json
+import yaml
 import numbers
 
 try:
@@ -466,12 +467,15 @@ class RefResolver(object):
             # Requests has support for detecting the correct encoding of
             # json over http
             if callable(requests.Response.json):
-                result = requests.get(uri).json()
-            else:
-                result = requests.get(uri).json
+                try:
+                    result = requests.get(uri).json()
+                except ValueError:
+                    result = None
+            if not result:
+               result = yaml.load(requests.get(uri).content)
         else:
             # Otherwise, pass off to urllib and assume utf-8
-            result = json.loads(urlopen(uri).read().decode("utf-8"))
+           result = yaml.load(urlopen(uri).read().decode("utf-8"))
 
         if self.cache_remote:

```

But on https://github.com/Julian/jsonschema/issues/420 it seems yaml is currently out of scope.

Can we fix it in `connexion` subclassing RefResolver?

_Originally posted by @ioggstream in https://github.com/zalando/connexion/issues/678#issuecomment-421061800_

dtkav:
Fixes #681 

 - Replace the "compatibility layer" that rewrites response codes to strings (to be jsonschema compatible) with a custom safe loader from openapi-spec-validator library
 - Use openapi-spec-validator UrlHandler in jsonschema RefResolver to be able to resolve yaml files
 - Rename some mentions of "swagger" to "openapi" following the openapi branding guidelines


dtkav:
Here is the relevant source for how these are implemented in openapi-spec-validator.
https://github.com/p1c2u/openapi-spec-validator/blob/master/openapi_spec_validator/constructors.py
https://github.com/p1c2u/openapi-spec-validator/blob/master/openapi_spec_validator/loaders.py
https://github.com/p1c2u/openapi-spec-validator/blob/master/openapi_spec_validator/handlers.py

dtkav:
I get why codacy doesn't like yaml.load(), but note that we are using `loader=ExtendedSafeLoader` which extends  `SafeLoader`, so we still get all of the same safety guarantees.

PR #684, opened 2018-09-17
Nullable

cziebuhr:
 - Set validator_map per api 
 - Allow setting validator on ResponseValidator
 - Change nullable handling to also work on non-properties
 - Add support for readOnly, writeOnly/x-writeOnly


connexion/decorators/json_validator.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if instance not in enums:
+        yield ValidationError("%r is not one of %r" % (instance, enums))
+
+
+def validate_required(validator, required, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    for property in required:
```
dtkav: `property` is a keyword, can you rename this please?

connexion/decorators/json_validator.py
```diff
@@ -0,0 +1,62 @@
+from jsonschema import Draft4Validator, _utils
```
dtkav: This isn't a decorator, IMHO it should be up one directory. What do you think?

dtkav:
Excellent work! This is much more readable.
It would be great if you added a unit test that uses the validator map at the API level.
In addition, please update the README section "New in Connexion 2.0" describing your changes.

connexion/decorators/json_validator.py
cziebuhr: What do you think about renaming jsonref.py (only contains one function) to jsonschema.py and adding it there?

cziebuhr:
> It would be great if you added a unit test that uses the validator map at the API level.
Do you mean to specify a custom validator or to test readOnly/writeOnly as real api?

connexion/decorators/json_validator.py
dtkav: sounds good to me

dtkav:
> Do you mean to specify a custom validator or to test readOnly/writeOnly as real api?

Yeah, I was thinking something in `tests/api/test_bootstrap.py` that instantiates an API and sets `validator_map` to something other than the default (you could try Draft6Validator for example).

Your tests so far are great low-level unit tests of the code you've added, but exercising the functionality from a higher level would be really great too because it would "lock-in" the interface that users are relying on.

For example, if there had been tests against **App**(validator_map=...), you would have known your changes would break users (backwards incompatible) because you would have had to adapt those tests.

Hopefully that makes sense, my brain is a little fried today.

(Adding a readOnly / writeOnly test case would be amazing too, but definitely "extra credit")

cziebuhr:
PR has been updated. I couldn't use jsonschema.py as a name, because of python 2.7 implicit relative imports.

tests/test_json_validation.py
```diff
@@ -0,0 +1,75 @@
...
+        def __init__(self, *args, **kwargs):
+            super(MyRequestBodyValidator, self).__init__(*args, validator=MinLengthRequestValidator, **kwargs)
+
+    validator_map = {'body': MyRequestBodyValidator}
+
+    app = App(__name__, specification_dir=json_validation_spec_dir)
+    app.add_api(spec, validate_responses=True, validator_map=validator_map)
+    app_client = app.app.test_client()
+
+    res = app_client.post('/v1.0/minlength', json={'foo': 'bar'}) # type: flask.Response
```
dtkav: I think you might have to use `data=json.dumps({"foo": "bar"})` and `headers={"Content-Type": "application/json"}`

dtkav: These lines are causing CI to fail: https://travis-ci.org/zalando/connexion/jobs/430130461#L984
I said "might", because I haven't tested it (or maybe there's a better fix).

cziebuhr: Hmm, tox installed flask=0.10.1, on my maschine tox gave me 1.0.2

cziebuhr:
Fixed json handling (also for responses) and updated README

PR #697, opened 2018-09-25
Fix bad parsing of query params for AioHttp

### LINKED ISSUE #696: String query params are being chopped off leaving only last character standing
### Description

Given https://github.com/zalando/connexion/blob/master/examples/helloworld/swagger/helloworld-api.yaml swagger file, let's add query param to the endpoint

```
- name: someId
  in: query
  required: true
  type: string
```

And modify the handler accordingly to accept the parameter. 
Now when the flow reaches the function `someId` will be set but it will be always the last character of the string passed in the request (i.e. via curl or httpie). 

Also I've added some printing to the `AioHttpApi#get_request` function to examine the content of the `request.query` stuff. Did the same in the `couroutine_wrapper` wrappers. In both cases I could see the full string as passed in query parameters of the request. That puzzles me a lot.... After the code reaches the decorators execution, i.e. this line:

```
2018-09-25 09:18:54,127 - ERROR - connexion.decorators.parameter - parameter:wrapper:154 - query_arguments={'someId': '7'}, from_request={'someId': '7'} - {}
```
We can see that the value is already chopped down.

One last thing. I think that only `query_parameters` are being truncated. Everything seems to be ok for `path_parameters`.

### Expected behaviour

`curl 127.0.0.1:3000/v1/gretting/kornicameister?someId=123-456-789` results in `someId` being set to `123-456-789`

### Actual behaviour

`someId` is `9` (i.e. last character of the string). 

### Steps to reproduce

!!! will provide link to the repro rep later, can't really do it now !!!
- https://github.com/kornicameister/aio_connexion_bug
- pip install -r requirements.txt
- PYTHONPATH=./ python bug/main.py
- curl "127.0.0.1:3000/greeting/kornicameister?queryA=query_a&queryB=query_b"

### Additional info:

Output of the commands:

- `python --version` => `3.7.0`
- `pip show connexion | grep "^Version\:"` => `1.5.1`
- `pip show aiohttp | grep "^Version\:"` => `3.4.4`

kornicameister:
Fixes #696 .

- do not join query parameters values with `,` for AioHttpApi. `connexion.decorators.uri_parsing.AbstractURIParser` requires those to be an array, given the code (i.e. `{'queryParam': ['aValue']}`). However the dictionary that reaches the parses always have single string as value for each key.

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -167,7 +167,7 @@ class AioHttpApi(AbstractAPI):
         logger.debug('Getting data and status code',
                      extra={'has_body': req.has_body, 'url': url})
 
-        query = {k: ','.join(v) for k, v in parse_qs(req.rel_url.query_string).items()}
+        query = {query_key: query_values for query_key, query_values in parse_qs(req.rel_url.query_string).items()}
         headers = {k.decode(): v.decode() for k, v in req.raw_headers}
         body = None
         if req.can_read_body:
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -167,7 +167,7 @@ class AioHttpApi(AbstractAPI):
         logger.debug('Getting data and status code',
                      extra={'has_body': req.has_body, 'url': url})
 
-        query = {query_key: query_values for query_key, query_values in parse_qs(req.rel_url.query_string).items()}
+        query = parse_qs(req.rel_url.query_string)
         headers = {k.decode(): v.decode() for k, v in req.raw_headers}
         body = None
         if req.can_read_body:
--- a/tests/fakeapi/aiohttp_handlers.py
+++ b/tests/fakeapi/aiohttp_handlers.py
@@ -47,20 +47,17 @@ def aiohttp_access_request_context(request_ctx):
 
 @asyncio.coroutine
 def aiohttp_query_parsing_str(query):
-    assert query == 'query'
-    return ConnexionResponse(status_code=204)
+    return ConnexionResponse(body={'query': query})
 
 
 @asyncio.coroutine
 def aiohttp_query_parsing_array(query):
-    assert query == ['queryA', 'queryB']
-    return ConnexionResponse(status_code=204)
+    return ConnexionResponse(body={'query': query})
 
 
 @asyncio.coroutine
 def aiohttp_query_parsing_array_multi(query):
-    assert query == ['queryA', 'queryB', 'queryC']
-    return ConnexionResponse(status_code=204)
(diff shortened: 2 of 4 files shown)


kornicameister:
For reviewing people. I am not sure about it, but shouldn't the minimal version of `aiohttp` and other `aio-ish` libraries be bumped in `requirements-aiohttp.txt`. Frankly speaking the latest available `aiohttp` is `3.4.4`. Even if the requirements file says `>=`, sounds to me like this might be a good idea.

hjacobs:
@kornicameister can you add a test please? So we can avoid introducing a similar regression in the future.

connexion/apis/aiohttp_api.py
```diff
@@ -167,7 +167,7 @@ def get_request(cls, req):
         logger.debug('Getting data and status code',
                      extra={'has_body': req.has_body, 'url': url})
 
-        query = {k: ','.join(v) for k, v in parse_qs(req.rel_url.query_string).items()}
+        query = {query_key: query_values for query_key, query_values in parse_qs(req.rel_url.query_string).items()}
```
dtkav: can this just be `query = parse_qs(req.rel_url.query_string)`?

kornicameister: Yeah, nice catch ;-).

kornicameister:
@hjacobs pushed the tests. Hopefully, I wrote them correctly (i.e. consistently with the rest).

dtkav:
I'd recommend making the test fixture just echo, so the test contains all the logic. reading the test it's difficult to know why a 204 response is good.

dtkav:
otherwise, looking good. thanks for adding tests!

kornicameister:
@dtkav I hope that this is what you wanted me to do

tests/fakeapi/aiohttp_handlers.py
```diff
@@ -45,6 +45,21 @@ def aiohttp_access_request_context(request_ctx):
     return ConnexionResponse(status_code=204)
 
 
+@asyncio.coroutine
+def aiohttp_query_parsing_str(query):
+    return ConnexionResponse(body={'query': query})
```
dtkav: FYI you could just `return query`, but this is totally fine too.

dtkav:
Hey @hjacobs , this is looking good from my perspective. Can you have a look?

tests/fakeapi/aiohttp_handlers.py
kornicameister: Yeah, I know I could've but just wanted to be consistent with the rest of the handlers here. I am pretty picky about this point thanks to my colleagues from FE team :D

jmcs:
:+1:

PR #698, opened 2018-09-27
 Setting base_path updates served spec

### LINKED ISSUE #692: [dev-2.0] setting Api(base_path=...) does not work with swagger-ui-bundle
Setting `Api(base_path=...)` doesn't work with `swagger-ui-bundle==0.0.2`. Swagger-ui does not get the correct base_path, and so cannot hit the API.

For example, add `my_api.yml` under `/v0`, and add swagger ui in `/v0/ui`.
```
app.add_api('my_api.yml', base_path='/v0', strict_validation=True, validate_responses=True)
```

I you click `Try it out` and the `Execute` in swagger ui, the request is sent to API without `base_path` (`/v0` in this case).

_Adapted from a post by @czchen in https://github.com/zalando/connexion/pull/619#issuecomment-423217816_

dtkav:
Fixes #692 

This is very much a WIP, I just wanted to push it to show approximately what needs to be done (for the sake of discussion).
Sorry for the low quality PR.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -9,7 +9,7 @@ import jinja2
 import six
 import yaml
 from openapi_spec_validator.exceptions import OpenAPIValidationError
-from six.moves.urllib.parse import urlparse
+from six.moves.urllib.parse import urlsplit
 
 from ..exceptions import InvalidSpecification, ResolverError
 from ..json_schema import resolve_refs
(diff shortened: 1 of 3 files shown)


dtkav:
note to self: add a unit test to prevent regressions

connexion/apis/abstract.py
```diff
@@ -174,20 +174,43 @@ def _validate_spec(self, spec):
...
+                servers = self.specification.get('servers', [])
+                for i, server in enumerate(servers):
+                    parsed = urlsplit(server['url'])
+                    self.raw_spec['servers'][i] = server = \
+                        urlunsplit(
+                            (parsed.schem, parsed.netloc, self.base_path) +
+                            ('', '')
+                        )
+                else:
+                    raise RuntimeError(
```
JuxhinDB: Codacy seems to be complaining about this line. Maybe embed it as part of the loop, or put an earlier check to see if any server blocks where specified.

jmcs:
👍

PR #709, opened 2018-10-03
disable threading in sqlalchemy demo (in-memory db)

### LINKED ISSUE #706: sqlalchemy example raises "sqlite3.OperationalError: no such table: pets"
### **Description**

The sqlalchemy example (under WSL with Ubuntu distribution) raises the following error when queried for the content of `pets` (or any other entry): 

`sqlite3.OperationalError: no such table: pets`

### **Expected behaviour**

Should return an empty list.

### **Actual behaviour**

Returns 500, "Internal Server Error".
```
ERROR:flask.app:Exception on /pets [GET]
  File "/home/user/conda/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1193, in _execute_context
    context)
  File "/home/user/conda/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such table: pets
```

### **Steps to reproduce**

```
cd connexion/examples/sqlalchemy
python app.py
curl -X GET --header 'Accept: application/json' 'http://localhost:8080/pets'
```

### **Additional info**

This issue has been adressed in: [https://gehrcke.de/2015/05/in-memory-sqlite-database-and-flask-a-threading-trap/](url) The in-memory database can be created on the "wrong" thread and thus, cannot be accessed.

**Fix:**

The simplest solution would be to not use an in-memory database, i.e., replace

https://github.com/zalando/connexion/blob/2b6169aade03e8a9dc7109c7ba7a224b3888e51f/examples/sqlalchemy/app.py#L50

with, for instance,
```
db_session = orm.init_db('sqlite:////tmp/db.sqlite') 
```

Output of the commands:

- `python --version`:  Python 3.6.6 :: Anaconda, Inc.
- `pip show connexion | grep "^Version\:"` Version: 1.5.3

Can also be reproduced for 
- `python --version`:  Python 2.7.12
- `pip show connexion | grep "^Version\:"` Version: 1.5.3

dtkav:
disable threading in sqlalchemy demo because it uses an in-memory database, which is not shared across threads.

Fixes #706 

 - disable threading in sqlalchemy demo

### DIFF
--- a/examples/sqlalchemy/app.py
+++ b/examples/sqlalchemy/app.py
@@ -60,4 +60,7 @@ def shutdown_session(exception=None):
 
 
 if __name__ == '__main__':
-    app.run(port=8080)
+    app.run(
+        port=8080,
+        threaded=False  # in-memory database isn't shared across threads
+    )


nmusatti:
I'm not an expert myself, even though I originally wrote this example. I prefer your solution as it's automatically portable, whereas creating a physical temporary file would have made it a little more complicated.

dtkav:
@jmcs one-line bugfix if you have a second

jmcs:
:+1:

jmcs:
:+1:

PR #710, opened 2018-10-03
Add some set attributes of Flask application like customer static_folder

thinksource:
Fixes # .

 I would like customer Flask app like:
```
connex_app = connexion.FlaskApp(__name__, static_folder="static/dist", \
    template_folder="static", specification_dir=basedir+'\\swagger')
```
But I found the connex_app.app do not change anything and I can not use specialization my personal static_folder and template_folder.

I want when I initialized the app I can use same Flask accept parameters to customer my Flask.

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -19,9 +19,12 @@ logger = logging.getLogger('connexion.app')
 class FlaskApp(AbstractApp):
     def __init__(self, import_name, server='flask', **kwargs):
         super(FlaskApp, self).__init__(import_name, FlaskApi, server=server, **kwargs)
-
+        self.flask_dict={}
+        for key value in kwargs.items():
+            self.flask_dict[key] = value
+            
     def create_app(self):
-        app = flask.Flask(self.import_name)
+        app = flask.Flask(self.import_name, **self.flask_dict)
         app.json_encoder = FlaskJSONEncoder
         return app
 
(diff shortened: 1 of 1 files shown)


connexion/apps/flask_app.py
```diff
@@ -19,9 +19,12 @@
 class FlaskApp(AbstractApp):
     def __init__(self, import_name, server='flask', **kwargs):
         super(FlaskApp, self).__init__(import_name, FlaskApi, server=server, **kwargs)
-
+        self.flask_dict={}
```
dtkav: kwargs is already a dict.
why not `self.flask_args = kwargs`?

dtkav:
Cool suggestion.
Maybe wait until the Connexion 2.0 is release, as the `kwargs` are currently interpreted as `old_style_options` within AbstractApp.

connexion/apps/flask_app.py
thinksource: just make sure the kwargs parameters, I am not deeply in this connexion, afraid  **kwargs change, like: 
```
kwargs.pop('foo')
```
in your oldstyle or AbstractApp initialization.

dtkav: for future reference, in that case you can use `copy`

dtkav:
@thinksource is this still something you're interesting in picking up again, or should I close this out?

fgsalomon:
> @thinksource is this still something you're interesting in picking up again, or should I close this out?

I'm also interested in this. Is there any workaround available?

fgsalomon:
I'm trying to workaround this by setting the `static_folder` and the `static_url_path` after instantiating the Connexion App like this:

```python
        self.app = connexion.App(__name__, specification_dir=SWAGGER_DIR)
        self.app.app.static_url_path = STATIC_PATH + "/static"
        self.app.app.static_folder = "dist/static"
```

But it's not working, static files are served in "/".

@dtkav do you know if it would be possible to set them in another way?

Thanks

RobbeSneyders:
Closing as a similar change was merged in #1173.

PR #712, opened 2018-10-05
bugfix: get securitySchemes in openapi 3 APIs (thanks @cziebuhr)

### LINKED ISSUE #711: [dev-2.0] securitySchemes not passed to operation
### Description
@cziebuhr found that "auth all paths" wasn't working for OpenAPI specs in dev-2.0

### Expected behaviour
Providing securitySchemes should cause them to be used when auth_all_paths is true

### Actual behaviour
securitySchemes is always being set to {} in oas3

### Steps to reproduce
create API with auth_app_paths=True
include securitySchemes
use API as though there's no auth

### Additional info:
#703 includes this, and more
#710 includes just the fix, refactor/cleanup to come

Output of the commands:

- `python --version` 3.6.5
2.0.0rc1

dtkav:
bugfix: get securitySchemes in openapi 3 APIs
Fixes #711 

Extracts just the bugfix from #703 
Thanks @cziebuhr

 - include securitySchemes in self.security_definitions
 - ensures test runs on both swagger 2 and openapi 3

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -128,12 +128,17 @@ class AbstractAPI(object):
         # API calls.
         self.consumes = self.specification.get('consumes', ['application/json'])  # type: List[str]
 
-        self.security = self.specification.get('security')
-        self.security_definitions = self.specification.get('securityDefinitions', dict())
-        logger.debug('Security Definitions: %s', self.security_definitions)
-
         self.definitions = self.specification.get('definitions', {})
         self.components = self.specification.get('components', {})
+
+        self.security = self.specification.get('security', dict())
+        _security_schemes = self.components.get('securitySchemes', dict())
+        self.security_definitions = self.specification.get(
+            'securityDefinitions',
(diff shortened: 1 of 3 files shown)


jmcs:
:+1:

connexion/apis/abstract.py
```diff
@@ -128,12 +128,17 @@ def __init__(self, specification, base_path=None, arguments=None,
...
         self.consumes = self.specification.get('consumes', ['application/json'])  # type: List[str]
 
-        self.security = self.specification.get('security')
-        self.security_definitions = self.specification.get('securityDefinitions', dict())
-        logger.debug('Security Definitions: %s', self.security_definitions)
-
         self.definitions = self.specification.get('definitions', {})
         self.components = self.specification.get('components', {})
+
+        self.security = self.specification.get('security', dict())
```
cziebuhr: Default should be `list()`. But the change isn't critical as there is a 'boolean' check before the list gets accessed.

PR #713, opened 2018-10-08
Introduce Specification classes

dtkav:
This is a proof of concept for a discussion with @cziebuhr in #703 

 - factor out spec version switching
 - introduce Specification, Swagger2Specification, and OpenAPISpecification classes

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -308,7 +308,6 @@ class AbstractAPI(object):
             # update spec to include user-provided base_path
             self.specification.base_path = base_path
             self.base_path = base_path
-            assert self.base_path == self.specification.base_path
         else:
             self.base_path = self.specification.base_path
 
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -59,6 +59,10 @@ class Specification(collections_abc.Mapping):
     def version(self):
         return self._get_spec_version(self._spec)
 
+    @property
+    def security(self):
+        return self._spec.get('security')
+
     def __getitem__(self, k):
         return self._spec[k]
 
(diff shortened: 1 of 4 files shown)


connexion/apis/abstract.py
```diff
@@ -262,18 +364,19 @@ def add_operation(self, method, path, swagger_operation, path_parameters):
...
+                                          security_definitions=self.specification.security_definitions,
+                                          definitions=self.specification.definitions,
+                                          parameter_definitions=self.specification.parameter_definitions,
+                                          response_definitions=self.specification.response_definitions,
                                           **shared_args)
         else:
             operation = OpenAPIOperation(self,
-                                         components=self.components,
+                                         components=self.specification.components,
                                          **shared_args)
```
cziebuhr: Can't we just pass the specification object as a whole?

connexion/apis/abstract.py
```diff
@@ -27,30 +32,190 @@
...
+        return self._spec[k]
+
+    def __iter__(self):
+        return self._spec.__iter__()
+
+    def __len__(self):
+        return self._spec.__len__()
+
+    @classmethod
+    def from_file(cls, spec, arguments=None):
```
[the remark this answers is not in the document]
dtkav: yeah, what's a better name?

connexion/apis/abstract.py
dtkav: yeah, or see my TODO comment above.

connexion/apis/abstract.py
```diff
@@ -262,18 +364,19 @@ def add_operation(self, method, path, swagger_operation, path_parameters):
             "pass_context_arg_name": self.pass_context_arg_name
         }
 
-        if self.spec_version < (3, 0, 0):
+        # TODO refactor into AbstractOperation.from_spec(Specification, method, path)
```
dtkav: > Can't we just pass the specification object as a whole?

This way it will create the right `Operation` from the right `Specification` without needing the explicit version check here.

What do you think?

connexion/apis/abstract.py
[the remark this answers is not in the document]
cziebuhr: What about just `load`?

jmcs:
@dtkav is this PR ready?

dtkav:
Sorry @jmcs - I've been traveling in the US, so I haven't been super attentive.
FWIW, I don't think this PR is blocking a 2.0 release at all since it's mostly cosmetic.
Let me put some work in today and try to get rid of the WIP.

PR #726, opened 2018-10-18
simplify Operation class construction

dtkav:
Follow up to #713 

 - Introduce make_operation that creates the correct Operation class from a Specification
 - Move Specification classes to separate file (spec.py)

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -1,33 +1,19 @@
 import abc
-import copy
 import logging
 import pathlib
 import sys
-from typing import AnyStr, List  # NOQA
 
-import jinja2
 import six
-import yaml
-from openapi_spec_validator.exceptions import OpenAPIValidationError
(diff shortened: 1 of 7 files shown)


dtkav:
Hey @cziebuhr , some more refactor - let me know what you think!

cziebuhr:
I would have gone even further, to pass and save the spec object in the operation instead of single parameters in the operation constructor.

dtkav:
I think the current PR is as far as I can go without breaking interfaces or changing tons of tests.

dtkav:
do you want to put up a PR of what you're thinking?

cziebuhr:
> do you want to put up a PR of what you're thinking?

I can do that

cziebuhr:
> do you want to put up a PR of what you're thinking?

I can do that

cziebuhr:
> do you want to put up a PR of what you're thinking?

I can do that

jmcs:
👍

PR #728, opened 2018-10-21
Add plugins support

### LINKED ISSUE #719: Plugin Mechanism
### Description

I would like to propose a plugin mechanism to the request handling flow.
The idea is to enable the registration of functions / decorators on predefined hooks,
to later be triggered at certain points of the request life-cycle.
I would also like o be able to set the order of these plugins execution.

Some examples of possible plugins:

* Logging 
* Authentication
* Metric collection
 
The plugins could be initialized at the app definition.
The implementation could be very similar to the current custom validators, but free of the current assumptions (the validators are only triggered when needed, and not on every incoming request).

sharbov:
Fixes #719 .

This is an initial version to see if I'm on the right track, 
I'll be happy to hear some feedback, 
then I'll add the tests.

FYI @dtkav

### DIFF
--- a/connexion/operation.py
+++ b/connexion/operation.py
@@ -6,6 +6,7 @@ from .decorators.decorator import (BeginOfRequestLifecycleDecorator,
                                    EndOfRequestLifecycleDecorator)
 from .decorators.metrics import UWSGIMetricsCollector
 from .decorators.parameter import parameter_to_arg
+from .decorators.plugins import PluginsDecorator
 from .decorators.produces import BaseSerializer, Produces
 from .decorators.response import ResponseValidator
 from .decorators.security import (get_tokeninfo_func, get_tokeninfo_url,
@@ -139,7 +140,7 @@ class Operation(SecureOperation):
                  definitions=None, parameter_definitions=None, response_definitions=None,
                  validate_responses=False, strict_validation=False, randomize_endpoint=None,
                  validator_map=None, pythonic_params=False, uri_parser_class=None,
-                 pass_context_arg_name=None):
+                 pass_context_arg_name=None, plugins=None):
         """
         This class uses the OperationID identify the module and function that will handle the operation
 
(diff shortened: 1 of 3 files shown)


connexion/operation.py
```diff
@@ -387,6 +392,10 @@ def function(self):
         logger.debug('... Adding uri parsing decorator (%r)', uri_parsing_decorator)
         function = uri_parsing_decorator(function)
 
+        for plugins_decorator in self.__plugins_decorator:
```
eran-stratoscale: Why the for loop?
self.__plugins_decorator yields once.

dtkav:
I think this PR as-is is significantly less flexible than overriding `AbstractApi.function()`.
If you override `AbstractApi.function()` you can insert user code anywhere, or even replace middleware, but the proposed PR forces any plugins to run before uri parsing and validation. This seems somewhat arbitrary to me.
Django has a pretty good middleware concept (https://docs.djangoproject.com/en/2.1/topics/http/middleware/#activating-middleware), but it's more holistic. (example below)
```
MIDDLEWARE = [
    'django.middleware.security.SecurityMiddleware',
    'django.contrib.sessions.middleware.SessionMiddleware',
    'django.middleware.common.CommonMiddleware',
    'django.middleware.csrf.CsrfViewMiddleware',
    'django.contrib.auth.middleware.AuthenticationMiddleware',
    'django.contrib.messages.middleware.MessageMiddleware',
    'django.middleware.clickjacking.XFrameOptionsMiddleware',
]
```
I think the plugin concept is cool, but IMHO it should aim to basically replace the entire `AbstractApi.function()` method.

What do you think? Do you like the django-style approach?

BTW thanks for making the PR, I think it is perfect for starting concrete discussion around a plugin architecture!

sharbov:
Hey @dtkav

Thanks for the review. 

I see your point about the lack of flexibility in the current PR.
At the time I felt like refactring the entire `AbstractApi.function` would be a too drastic change. 

The Django middleware seems interesting, and a similar principle might apply here as well,
each of the decorators used in the `function` can be viewed as a kind of plugin / middleware.

Would you like me to have another go at it or would you rather wait for more input on this?

cognifloyd:
I prefer the approach in #760 for making pluggable deserialization/validation. I think the same approach will extend well to serialization.

What other kinds of plugins does this enable that content-type based plugins do not?

RobbeSneyders:
Closing due to lack of contributor feedback.

PR #730, opened 2018-10-23
[DRAFT] Directly use spec in operation classes

cziebuhr:
Follow up on #726.

- Code cleanup
- Directly use spec in operation classes
  - It saves a lot of parameter passing
  - Move some common code to base class
  - Future: Allow preprocessing of security definitions per spec and not per operation

Tests are not yet updated.
@dtkav what do you think?

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -1,33 +1,20 @@
 import abc
-import copy
 import logging
 import pathlib
 import sys
-from typing import AnyStr, List  # NOQA
+from typing import AnyStr  # NOQA
 
-import jinja2
 import six
-import yaml
-from openapi_spec_validator.exceptions import OpenAPIValidationError
-from six.moves.urllib.parse import urlsplit
 
-from ..exceptions import InvalidSpecification, ResolverError
-from ..json_schema import resolve_refs
-from ..operations import OpenAPIOperation, Swagger2Operation, make_operation
+from ..exceptions import ResolverError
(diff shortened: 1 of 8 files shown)


dtkav:
Hey @cziebuhr - if you have the bandwidth can you update the tests? I think this is a good change, and it's really just a question of timing w.r.t to the release.

cziebuhr:
I will update the tests, just wanted to get feedback before.
There are some mixtures of using "private" variables (prefixed with underscore) and respective property decorators. Do you have any preference on how to use them? Should subclasses directly use variables or properties? Some property decorators are not used at all or only for the tests.

cziebuhr:
IMHO this change shouldn't block dev-2.0.

dtkav:
> Should subclasses directly use variables or properties?

Yeah I think it's fine to use private variables in a subclass. The main thing is that the spec should be private to the Operation. The motivation behind this is to keep the interface for Operation well defined, and not also leak the Specification interface.

This was one of the primary motivators of my approach in #726 . The other driver was to maintain backwards compatibility. Because this diff (#730) changes the Operation interface, I think it either needs to make it into 2.0 or else it would have to be in 3.0.

dtkav:
FYI - I think @jmcs closed this because it's based on the dev-2.0 branch. It's still relevant to #764 and I'm still interested in this concept and cleanup.

If you're still interested and have time, then you'll need to re-submit a new PR against master.
We can continue to discuss on #764 a well.

Issue #607, opened 2018-06-21
Connexion jwt support


ctippur:
### Description

Wondering if there is support for jwt auth via connexion

### Expected behaviour

Should be a way to authenticate jwt token, I am unable to find a way to initialize jwt token via the flask app.

### Actual behaviour

I get an error:
```
raise RuntimeError("You must initialize a JWTManager with this flask "
RuntimeError: You must initialize a JWTManager with this flask application before using this method
```

I see an issue reported earlier but I was not sure if it led to a conclusion.

https://github.com/zalando/connexion/issues/124

Appreciate if I could get a working example of the same (if any)

### Additional info:

Output of the commands:

- `python --version`
3.6.2
- `pip show connexion | grep "^Version\:"`
Version: 1.1.9

bobh66:
Did you try flask_jwt_extended:

```
from flask_jwt_extended import JWTManager

app.config['JWT_SECRET_KEY'] = 'jwt-secret-string'

jwt = JWTManager(app)
```

ctippur:
Yes. I tried to use that.
I get an error trying to use the below line of code:
```access_token = create_access_token(identity=username)```
```
raise RuntimeError("You must initialize a JWTManager with this flask "
RuntimeError: You must initialize a JWTManager with this flask application before using this method
```

**Comment by ctippur** [comment]
I am using the below code in my controller

```
pp=connexion.FlaskApp(__name__).app
# Setup the Flask-JWT-Extended extension
app.config['JWT_SECRET_KEY'] = 'super-secret'  # Change this!
app.config['JWT_TOKEN_LOCATION'] = ['cookies','header']
jwt = JWTManager(app)
```

dtkav:
JWT auth was introduced with #732 and is available in connexion 2.0.x

PR #732, opened 2018-10-25
Add support for JWT authentication

### LINKED ISSUE #389: Add support for x-authentication-scheme and Bearer auth scheme in apiKey security definition

krise3k:
Fixes #389, #607 .

 - For swagger 2.0 add support for `x-authentication-scheme=bearer` in apiKey method.
 - For OpenAPI 3.0 add support for `scheme=bearer` in http method
 - Use `x-bearerInfoFunc` in security definition or `BEARERINFO_FUNC` env variable  to pass a reference to token validation function


examples/openapi3/jwt/app.py
```diff
@@ -0,0 +1,52 @@
...
+
+import time
+
+import connexion
+from werkzeug.exceptions import Unauthorized
+
+from jose import JWTError, jwt
+
+JWT_ISSUER = 'com.zalando.connexion'
+JWT_SECRET = '872f65270a71ca913766335276fb2c1b76b8ccdef749cdfc9a1598ac36d77c99'
```
dtkav: Can you make this something like `not_really_secret`? I can imagine some poor security engineer thinking these are leaked creds.

examples/openapi3/jwt/app.py
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    return jwt.encode(payload, JWT_SECRET, algorithm=JWT_ALGORITHM)
+
+
+def decode_token(token):
+    try:
+        return jwt.decode(token, JWT_SECRET, algorithms=[JWT_ALGORITHM])
+    except JWTError as e:
+        raise Unauthorized from e
```
dtkav: I don't think `raise ... from` is python 2.7 compatible.
You probably need to use `six.raise_from(...)`

tests/fixtures/secure_endpoint/openapi.yaml
```diff
@@ -121,3 +142,8 @@ components:
       name: X-Auth
       in: header
       x-apikeyInfoFunc: fakeapi.hello.apikey_info
+    jwt:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+      x-bearerInfoFunc: fakeapi.hello.jwt_info
```
dtkav: github is complaining about missing newline at end of file

dtkav:
great work! I had a few minor comments, but otherwise this is looking really good :)

examples/openapi3/jwt/app.py
krise3k: @dtkav It's not supposed to run on 2.7, please look at shebang.
BTW end of python 2.7 is coming :)

krise3k:
@dtkav I have fixed the issues.

examples/openapi3/jwt/app.py
dtkav: It's not a matter of my personal preference (I'm also on python 3). Connexion explicitly supports python 2.7, so I think making the examples backwards compatible is the right thing to do.

If you don't want to do it, I'm happy to. I'd prefer to follow the rule of least surprise now, because otherwise I'll have to respond to a ticket about this example being "broken" later.

krise3k: I have made it python 2.7 compatible.

But it looks like [helloworld example](https://github.com/zalando/connexion/blob/dev-2.0/examples/openapi3/helloworld/hello.py) is also broken on 2.7

dtkav:
Thanks @krise3k
Good call, I'm happy to fix the helloworld example in a separate diff.

dtkav:
@jmcs do you want to take a look as well, or are you happy for me to merge this?

connexion/decorators/security.py
```diff
@@ -222,6 +237,31 @@ def wrapper(request, required_scopes):
...
+def verify_bearer(bearer_info_func):
+    def wrapper(request, required_scopes):
+        authorization = request.headers.get('Authorization')
+        if not authorization:
+            return None
+
+        try:
+            auth_type, token = authorization.split(None, 1)
+        except ValueError:
+            raise OAuthProblem(description='Invalid authorization header')
```
Stranger6667: Is `OAuthProblem` class appropriate here? The code seems more generic and not only about `OAuth`

cziebuhr: Previously connexion only supported oauth. It's more generic now. We could rename the Exception, but it would be backward incompatible if someone has set up a custom error_handler.

cziebuhr:
Not sure if I missed something. Isn't this the same as oauth2 with an empty scope?

krise3k:
@cziebuhr When I have implemented JWT I found that verify function is a closure and guessed that it would be hard to avoid code duplication.
Today I take a second look at the code and  extract common function to verify oauth and bearer tokens.

cziebuhr:
Looks better now, but I still don't get the difference. IMHO the only difference is, that scope is not validated. Can't you achieve the same result when specifying oauth2 as type of the security scheme and an empty array as scope for security requirement?

dtkav:
I'm a bit out of the loop on this one, but It seems valuable to be able to define the correct security scheme in the spec, and not have to make everything work through oauth2. In addition, there may be a use case for someone supporting multiple security schemes, so they may need a handler for each one.
@cziebuhr thoughts?

cziebuhr:
One can't configure bearer method and oauth2 together, because if the first validator finds an Authorization header of type Bearer and can't validate its value, it will fail hard and not continue with the next validator.

dtkav:
@cziebuhr I'm not sure if you have blocking feedback for the PR, and if so how @krise3k can address your feedback.

I now get your point about oauth2 and http:bearer both using the `Bearer` keyword in the header. Thanks for clarifying that.

Still, the 3.x.x spec [0] supports http:bearer, so it seems wise to support it as well, even if it's just for the principle of least surprise (people wouldn't expect that they need to change auth methods in their spec to make auth work, even if they are similar/subsets/supersets).

AFAICT this diff adds support for the http:bearer options in the spec, and seems pretty DRY.

[0] https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-scheme-object

connexion/decorators/security.py
dtkav: You could also inherit from OAuthProblem in order to rename it and keep it backwards compatible.  (and leave a note to change this on the next major version bump).

dtkav:
I tried to update this PR to be against master (as connexion 2.0 was released), but it didn't work as cleanly as I expected (sorry!). @krise3k do you mind rebasing your changes onto master?

cziebuhr:
> @cziebuhr I'm not sure if you have blocking feedback for the PR, and if so how @krise3k can address your feedback.

I'm fine now.

krise3k:
@dtkav indeed, rebase was not so obvious. It was easier to start from current master and cherry pick the commits, I also squash them.

dtkav:
Great - thanks @krise3k and sorry about that.
I think the last eyes are probably @jmcs then.

jmcs:
Can you expand the [documentation](https://github.com/zalando/connexion/blob/master/docs/security.rst#apikey-authentication) and add doctstrings to the new functions?

krise3k:
@dtkav I will do it this weekend

krise3k:
@jmcs, @dtkav I push the changes.

jmcs:
:+1:

zout:
Hi @krise3k,

We are using your JWT code, and it works great, many thanks for your work. However, we are using scopes in the JWT token payload. In your example code you have set a scope that does nothing. We would like to contribute code for scope checking. Is there any reason you have skipped this? And is there any particular problem with that that we need to solve?

krise3k:
@zout it's because I didn't need to use scopes, also I wanted to make this example as simple as possible.

bjoluc:
@zout Do you still want to approach this? "Scope" (or should we say "claim"?) checking would be really helpful for us too. I can take a look at this otherwise.

zout:
@bjoluc We have build it in our project now, but as separate functions that decode de JWT to check for scope in the handling function. I could give you that code. But it is quite challenging to get to the scopes that are listed in the yaml file. So we decided not to do that.

bjoluc:
@zout Thanks for the quick reply! Yeah, we ended up implementing a custom solution too. Just thought it would be really nice to support a claims validation function (would have saved me a lot of time). I will try to figure it out and mention you in the PR in case I succeed...

PR #738, opened 2018-10-29
Change can_read_body to body_exists in AioHttpApi.get_request

### LINKED ISSUE #737: Request validation fails when request body is read in AIOHTTP middleware
### Description
Request validation fails when request body is read in AIOHTTP middleware.

### Expected behaviour
Able to read request body in AIOHTTP middleware and request validation passes.

### Actual behaviour
The request validation fails.

### Steps to reproduce
1. Call request.json() in an AIOHttp middleware
2. Make an HTTP request to an endpoint which expects a properly formatted JSON request body
3. The request is marked invalid

### Additional info:

Output of the commands:

- `python --version`
3.6.6
1.5.3

fijshion:
Fixes # .
fixes #737

 Use `request.body_exists` instead of `request.can_read_body` in `AioHttpApi.get_request`. `request.can_read_body` is set to False after the body is read once. This change makes it possible to read the request body in a middleware function before the validation code is called.

See https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.body_exists for details on the body_exists and can_read_body.


fijshion:
Here's a simple example that hits the bug in the non-streaming use case: https://github.com/fijshion/connexion_test

I had not considered the streaming use case. I assume there is a way to check if the request is streaming and use can_read_body instead. I don't have time right now test this and come up with a solution to support both. I'll take a look at it later if you don't beat me to it.

dtkav:
Thanks for providing the example! That is really helpful.

Your use of `request.json()` in your middleware consumes the 'stream', which is why it is marked as being 'at_eof' and so `can_read_body` returns `False`.

I'm wondering if it makes sense to ensure that your middleware doesn't consume the stream, rather than changing connexion to remove the check.

I tried to look around for best practice, and I couldn't find anything concrete. Let me know if you find anything.

fijshion:
Sure - thanks for looking at this. 

I don't have any specific examples of this (I'm new to AIOHttp) but this note in the doc makes me believe the request body is intended to be read multiple times:

> Note
The method does store read data internally, subsequent json() call will return the same value.

 https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.json

dtkav:
It seems the AioHttp API has changed to support re-reading text and json, so that's some evidence that this change could be ok (as long as we pin the minimum version of AioHttp).

On the other hand, [multipart requests](https://docs.aiohttp.org/en/v3.0.1/multipart.html#aiohttp-multipart) can still only be consumed once, so this change still seems like it would break that.

I'd really appreciate if someone with more AioHttp knowledge could weigh in. Support for AioHttp was added by @dutradda about a year ago, maybe he'd be willing to add his thoughts.

dtkav:
Sorry for the long period without any updates on this.
Without any other input, I'm inclined to advocate for merging this to unblock you.
I guess we'll see if any other aiohttp users run into problems with streaming or multipart down the line.

jmcs:
:+1:

PR #753, opened 2018-11-06
pass files to handler if filename in arguments _or_ has_kwargs

### LINKED ISSUE #750: Uploaded file not passed to endpoints taking kwargs
### Description
I am not 100% sure this should be considered a bug, but the behavior has changed between 1.2.0 and 2.0.0, so it might be a regression. Uploaded files are only passed as arguments to endpoints not taking kwargs in version 2.0.0.

### Expected behaviour
When uploading a file, the file should always be passed as a parameter to the endpoint.

### Actual behaviour
When uploading a file, the file is only passed to endpoints not taking kwargs.

### Steps to reproduce
```
flask 1.0.2
requests 2.20.0
connexion 1.2.0 and 2.0.0
```

api.yaml:
```
swagger: '2.0'
info:
  version: '1'
  title: File Upload

paths:
  /upload:
    post:
      operationId: app.upload
      consumes:
        - multipart/form-data
      parameters:
        - in: formData
          name: upload_file
          type: file
          required: false
      responses:
        '200':
          description: OK
        '400':
          description: Bad request.
        default:
          description: Error
```

app.py:
```
def upload(upload_file=None, **kwargs):
    if upload_file is None:
        return "Did not get file as arg"
    return "Got file as arg"
```

Run application and send request:
```
connexion run api.yaml
python -c 'import requests; print requests.post("http://localhost:5000/upload", files=dict(upload_file="test")).text'
```

Response with connexion 1.2.0: Got file as arg
Response with connexion 2.0.0: Did not get file as arg

This seems to be the relevant change since 1.2.0:

https://github.com/zalando/connexion/blob/44ea9336fe79ddb9592942b405a663d2e3aa612c/connexion/operations/abstract.py#L169-L171

### Additional info:

Output of the commands:

- `python --version`: Python 2.7.15rc1
- `pip show connexion | grep "^Version\:"`: Version: 2.0.0

dtkav:
Fixes #750 

 - pass file if filename in arguments _or_ has_kwargs

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -168,7 +168,7 @@ class AbstractOperation(SecureOperation):
 
     @staticmethod
     def _get_file_arguments(files, arguments, has_kwargs=False):
-        return {k: v for k, v in files.items() if not has_kwargs and k in arguments}
+        return {k: v for k, v in files.items() if k in arguments or has_kwargs}
 
     @abc.abstractmethod
     def _get_val_from_param(self, value, query_defn):
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -288,7 +288,7 @@ def test_formdata_missing_param():
     return ''
 
 
-def test_formdata_file_upload(formData):
+def test_formdata_file_upload(formData, **kwargs):
     filename = formData.filename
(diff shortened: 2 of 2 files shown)


jmcs:
:+1:

Issue #511, opened 2017-09-10
connexion.problem not compatible with add_error_handler


mxmo0rhuhn:
### Description
Returning an instance of `connexion.problem` from an error handler results in an application crash in `connexion == 1.1.9`

### Expected behaviour (`connexion == 1.0.129`)

Returning a nice error like:
```
HTTP/1.0 500 INTERNAL SERVER ERROR
Content-Type: application/problem+json
Content-Length: 129
Server: Werkzeug/0.12.2 Python/3.6.1

{
  "type": "about:blank",
  "title": "Internal Server Error",
  "detail": "Could not connect to LDAP server",
  "status": 500
}
``` 

### Actual behaviour in (`connexion == 1.1.9`)

Stacktrace 
```
Traceback (most recent call last):
  File "/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/lib/python3.6/site-packages/flask/app.py", line 1615, in full_dispatch_request
    return self.finalize_request(rv)
  File "/lib/python3.6/site-packages/flask/app.py", line 1630, in finalize_request
    response = self.make_response(rv)
  File "/lib/python3.6/site-packages/flask/app.py", line 1740, in make_response
    rv = self.response_class.force_type(rv, request.environ)
  File "/lib/python3.6/site-packages/werkzeug/wrappers.py", line 885, in force_type
    response = BaseResponse(*_run_wsgi_app(response, environ))
  File "/lib/python3.6/site-packages/werkzeug/wrappers.py", line 57, in _run_wsgi_app
    return _run_wsgi_app(*args)
  File "/lib/python3.6/site-packages/werkzeug/test.py", line 884, in run_wsgi_app
    app_rv = app(environ, start_response)
TypeError: 'ConnexionResponse' object is not callable
```

### Steps to reproduce

1. Define handler: 
```
def ldap_connection_error(the_error):
[...]

PR #759, opened 2018-11-07
Handle connexion.problem being directly returned in error handlers

samuelgrigolato:
I was able to reproduce #511 in the `dev-2.0` branch. This PR adds a regression test case for it and also propose a fix, by decorating the handlers before registering them with Flask, taking the opportunity to hand them to `FlaskApi.get_response` so the conversion magic can happen.

As this is my first contribution to this project it is possible that the fix and the test case would be better placed elsewhere in the code, so just let me know if that is the case.

### DIFF
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -410,6 +410,10 @@ def throw_problem_exception():
     )
 
 
+def throw_value_error():
+    raise ValueError('Something really wrong.')
+
+
 def unordered_params_response(first, path_param, second):
     return dict(first=int(first), path_param=str(path_param), second=int(second))
 
--- a/tests/fixtures/problem/openapi.yaml
+++ b/tests/fixtures/problem/openapi.yaml
@@ -79,6 +79,13 @@ paths:
       responses:
         '200':
           description: Problem exception
+  /value_error:
+    get:
+      description: Simply raise a ValueError
+      operationId: fakeapi.hello.throw_value_error
+      responses:
+        '200':
+          description: ValueError
 servers:
   - url: /v1.0
 components:
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -57,7 +57,9 @@ class FlaskApp(AbstractApp):
 
     def add_error_handler(self, error_code, function):
         # type: (int, FunctionType) -> None
-        self.app.register_error_handler(error_code, function)
+        def wrapper(error):
+            return FlaskApi.get_response(function(error))
+        self.app.register_error_handler(error_code, wrapper)
 
     def run(self, port=None, server=None, debug=None, host=None, **options):  # pragma: no cover
         """
--- a/tests/api/test_errors.py
+++ b/tests/api/test_errors.py
@@ -2,6 +2,10 @@ import json
 
 import flask
 
+import pytest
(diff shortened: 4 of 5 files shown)


tests/fixtures/problem/openapi.yaml
```diff
@@ -79,6 +79,13 @@ paths:
       responses:
         '200':
           description: Problem exception
+  /value_error:
+    get:
+      description: Simply raise a ValueError
+      operationId: fakeapi.hello.throw_value_error
+      responses:
+        '200':
```
dtkav: shouldn't this be 599?

tests/fakeapi/hello.py
```diff
@@ -410,6 +410,10 @@ def throw_problem_exception():
     )
 
 
+def throw_value_error():
+    raise ValueError('Something really wrong.')
```
dtkav: Does this message get lost?

dtkav:
Nice work! Thanks for the PR.
I added a few very minor comments, otherwise this is looking good!

tests/fixtures/problem/openapi.yaml
samuelgrigolato: yes.. absolutely! I've fixed this.

tests/fakeapi/hello.py
samuelgrigolato: Yes, but this is a user's decision, not connexion's. The exception data get lost when the custom error handler decides to not pass it to `connexion.problem` in any way. In the regression test case scenario this happens here:

```py
def the_error_handler(error):
    return problem(599, 'Problem', 'Something wrong.')
app.add_error_handler(ValueError, the_error_handler)
```

Note that ther `error` parameter is not used inside the function.

samuelgrigolato:
@jmcs would you be able to share the reason why this PR has been closed? I'm asking because I've fixed @dtkav requests and I don't know if it is desired for me to open another PR.

dtkav:
hey @samuelgrigolato , your PR is looking good, but it's against the dev-2.0 branch which was used for staging the 2.0 release. You can also reopen this PR and change the base branch to master, but it will probably be easier to cherry pick your commits to a new branch.

samuelgrigolato:
HI @dtkav! Thanks for the reply. I cherry-picked my work on top of the origin/master and pushed the resulting branch. I also updated the base branch of this PR to the `master` branch, but I'm not sure I can reopen it by myself (maybe I don't have permission to do that?).

samuelgrigolato:
No worries, I'm glad to help!

tests/fixtures/problem/openapi.yaml
samuelgrigolato: I think what is happening here is that GitHub does not automatically track updates to closed PRs. If you look at the commit in my own fork then you'll see that the line is changed: https://github.com/samuelgrigolato/connexion/blob/fix_returning_problem_in_error_handlers/tests/fixtures/problem/swagger.yaml#L112

samuelgrigolato: Just for the sake of completion, this is the reference to the same line in the `openapi.yaml` file (which you originally pointed out in your review): https://github.com/samuelgrigolato/connexion/blob/fix_returning_problem_in_error_handlers/tests/fixtures/problem/openapi.yaml#L82

PR #761, opened 2018-11-08
only pass body on PATCH/POST/PUT requests

### LINKED ISSUE #757: Additional body kwarg given on GET requests
### Description
When sending a GET request, an additional empty `body` keyword argument is sent to the handling function.

### Expected behaviour
Only the described parameters should be forwarded to the function. Also, according to the spec, it seems that there should be no body provided on GET and DELETE.

### Actual behaviour
An empty `body` keyword argument is added as a function parameter.

### Steps to reproduce
requirements.txt
```
flask==1.0.2
requests==2.20.0
connexion==2.0.1
```
api.yaml
```yaml
openapi: 3.0.0
info:
  version: '1'
  title: Sample HTTP GET endpoint

paths:
  '/{some_param}':
    get:
      description: Testing
      operationId: app.sample_get
      parameters:
        - name: some_param
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
        '400':
          description: Bad request.
        default:
          description: Error
```

app.py
```python
def sample_get(**kwargs):
    return kwargs
```

Run application and send request
```sh
connexion run api.yaml
python -c 'import requests; print requests.get("http://localhost:5000/testing").text'
```
Returns
```json
{
  "body": {},
  "some_param": "testing"
}
```

### Additional info:
- Python 2.7.10
- connexion 2.0.1

dtkav:
Fixes #757 

 - only read `body` and `files` on PATCH/POST/PUT requests


jmcs:
:+1:

fgsalomon:
Hi,

Would be possible to also pass the body on DELETE requests?

fgsalomon:
Ok,  thanks Daniel.

PR #762, opened 2018-11-08
Don't let openapi_spec_validator tamper raw spec

cziebuhr:
openapi_spec_validator adds `x-scope` to schemas with `$ref`, which then also get served in `swagger.json`.
```
"$ref" : "#/definitions/Container",
"x-scope" : [
   "",
   "#/definitions/InstanceStats"
]
```
In my case, bravado (which internally uses swagger_spec_validator) fails using the spec:
```
jsonschema.exceptions.RefResolutionError: unknown url type: ''
```


dtkav:
What about using the spec that's passed in, and moving validation before resolving refs?
That would avoid another deepcopy.
```
     def __init__(self, raw_spec):
         self._raw_spec = copy.deepcopy(raw_spec)
         self._set_defaults()
         self._validate_spec(raw_spec)  # change to a staticmethod
         self._spec = resolve_refs(raw_spec)
```

cziebuhr:
That's also possible, if it's ok that the passed dict will get changed.

dtkav:
it is already changed by the ref resolver, so I think the only risk is that
the validator and ref resolver have some interaction.

jmcs:
:+1:

PR #772, opened 2018-11-09
DEMO: also serve yaml spec

### LINKED ISSUE #769: Serve Swagger Spec in YAML format
### Description
Currently a JSON representation of the Swagger/OpenAPI spec can be found at /swagger.json. It would be helpful if the YAML format was also available by changing the extension (ie. /swagger.yaml).

### Expected behaviour
The configured swagger spec is returned in YAML format from the /swagger.yaml endpoint.

### Actual behaviour
HTTP/1.0 404 NOT FOUND
Content-Type: application/problem+json

{"type": "error", "code": "objectNotFound", "message": "The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again."}

dtkav:
Fixes #769 

 - also serve the spec as yaml

Note that this implementation is a bit of a hack - I wanted to get the discussion going.


kornicameister:
5 cents on this is that I don't fancy having one thing doing too many things in the same time without explicit request. I'd say that if user wishes to server specification as `yaml` he has good reasons for it. If that is his/her wish let's maybe check how `openapi_spec_path` looks like. If:

- ends with `.json` or there is no format suffix, let's servr `JSON`
- ends with `.yml` or `yaml`, let's serve `yaml`

User gets the flexibility and there are not that many extra endpoints. With this change, as proposed now, you get either `0` (if you disable serving the spec) or `3` endpoints, right?

dtkav:
I think we always want to serve json for the client / UI.

I'm not personally a fan of this feature, but I decided to put the PR up for comment / discussion.

Another way to do this would just be to put your spec somewhere that serves up static files, and then also import it from there. I'll add a comment about that on the ticket.

dtkav:
Thanks for the PR!
There was a discussion in #769 on whether this feature should be added (functionality vs. code bloat).
Please add your thoughts there if you think this is a must-have feature in connexion.

PR #773, opened 2018-11-09
rename bandit.yaml to bandit.yml

dtkav:
Fixes #679  (?)

We're still seeing the warning for using asserts in tests.
I think it might be because I named the file slightly wrong.


jmcs:
:+1:

PR #789, opened 2018-11-20
Fix handling additionalProperties in body

krise3k:
Currently when object is send as body parameter only properties defined in `properties` in body schema are passed to handler function. Additional keys are filtered out, which is opposite to [spec](https://github.com/OAI/OpenAPI-Specification/blame/3.0.2/versions/3.0.2.md#L2305). For me specification is counter intuitive :(

 - If `additionalProperties` is not set or is `True`, all properties not defined in `properties` are passed without type casting.
 - If `additionalProperties` declares value type,  unknown properties are cast according to https://github.com/zalando/connexion#type-casting.

The best explanation for `additionalProperties` I found in https://github.com/OAI/OpenAPI-Specification/issues/668#issuecomment-218829120

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -252,6 +252,10 @@ class OpenAPIOperation(AbstractOperation):
         body_props = {sanitize(k): {"schema": v} for k, v
                       in self.body_schema.get("properties", {}).items()}
 
+        # by OpenAPI specification `additionalProperties` defaults to `true`
+        # see: https://github.com/OAI/OpenAPI-Specification/blame/3.0.2/versions/3.0.2.md#L2305
+        additional_props = self.body_schema.get("additionalProperties", True)
+
         body = body or default_body
 
         if self.body_schema.get("type") != "object":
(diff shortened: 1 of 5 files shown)


jmcs:
Can you also update the documentation. Thanks for the PR.

dtkav:
IMHO the important part to document with respect to connexion is that the type-casting magic will only work for properties that are defined in the schema.
Otherwise I don't think this new behavior is surprising, I just implemented it wrong originally (so this is a bugfix).

cziebuhr:
Sanitizing body arguments is fine for individual object properties which are fed as its own argument to handler functions. IMHO the `body` property itself should be passed unchanged without any magic involved (also if `type` != `object` or `type` not specified at all). It better aligns to the current behavior for swagger 2 and to the spec.
Background: I'm currently working on supporting `discriminator` value (polymorphism). Schemas are quite complicated there and we should try to keep the amount of own validation magic (other than what jsonschema does) to a bare minimum.

krise3k:
@jmcs This pull request doesn't introduce new functionality, so I don't know what should be documented. 
It's only fix passing additional properties to handler function on openapi v3.x, which always works for  openapi v2 (swagger)

jmcs:
:+1:

krise3k:
thanks @jmcs

PR #794, opened 2018-11-24
Added OpenAPI 3 getting API Key from cookie support.

ErmakovDmitriy:
Added support for getting API key from client cookies.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -12,7 +12,7 @@ from connexion.utils import get_function_from_name
 from ..exceptions import (ConnexionException, OAuthProblem,
                           OAuthResponseProblem, OAuthScopeProblem)
 
-from http.cookies import SimpleCookie
+from six.moves import http_cookies
 
 logger = logging.getLogger('connexion.api.security')
 
@@ -117,8 +117,10 @@ def security_deny(function):
     :type function: types.FunctionType
     :rtype: types.FunctionType
     """
+
     def deny(*args, **kwargs):
         raise ConnexionException("Error in security definitions")
+
     return deny
 
 
@@ -183,6 +185,7 @@ def verify_authorization_token(request, token_info_func):
 
 
 def verify_oauth(token_info_func, scope_validate_func):
+
     def wrapper(request, required_scopes):
         token_info = verify_authorization_token(request, token_info_func)
         if token_info is None:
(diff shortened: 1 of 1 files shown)


connexion/decorators/security.py
```diff
@@ -228,12 +230,29 @@ def wrapper(request, required_scopes):
...
 
+def get_cookie_value(cookies, name):
+    '''
+    Returns cookie value by its name. None if no such value.
+    :param cookies: str: cookies raw data
+    :param name: str: cookies key
+    '''
+    cookie_parser = SimpleCookie()
+    cookie_parser.load(cookies)
+    if name in cookie_parser.keys():
```
dtkav: is it possible to do this in a pythonic way?
```
try:
    return cookie_parser[name].value
except KeyError:
    return None
```

dtkav:
Thanks for the PR! I had a few minor comments, but otherwise this is looking good.

connexion/decorators/security.py
```diff
@@ -12,6 +12,8 @@
 from ..exceptions import (ConnexionException, OAuthProblem,
                           OAuthResponseProblem, OAuthScopeProblem)
 
+from http.cookies import SimpleCookie
```
dtkav: you'll need to use six.http_cookies since we support python 2.7

connexion/decorators/security.py
ErmakovDmitriy: Hi! Thank you!
I made the change, as you had requested.

```suggestion
def get_cookie_value(cookies, name):
    '''
    Returns cookie value by its name. None if no such value.
    :param cookies: str: cookies raw data
    :param name: str: cookies key
    '''
    cookie_parser = http_cookies.SimpleCookie()
    cookie_parser.load(str(cookies))
    try:
        return cookie_parser[name].value
    except KeyError:
        return None

```
I added str(cookies) to your example to convert the "cookies" variable to string to make comparison in python Cookies.py possible:
```
def load(self, rawdata):
        """Load cookies from a string (presumably HTTP_COOKIE) or
        from a dictionary.  Loading cookies from a dictionary 'd'
        is equivalent to calling:
            map(Cookie.__setitem__, d.keys(), d.values())
        """
        if type(rawdata) == type(""):
            self.__ParseString(rawdata)
        else:
            # self.update() wouldn't call our custom __setitem__
            for k, v in rawdata.items():
                self[k] = v
        return
    # end load()
```

P.S. I am not good at git. I was not able to make correct addtion to your request with some kind of link to my changes. I am really sorry.
https://github.com/ErmakovDmitriy/connexion/commit/7e312eedacf80a7a3b267b079c8236a8e0b80b06

connexion/decorators/security.py
ErmakovDmitriy: Thank you!
I have done it.
It works as I can see.

PR #798, opened 2018-11-27
Add for support relative paths in schema refs -- updated.

crunk1:
Refs #461
Refs #254
Updates #543 

From #543:
"Fix to allow the json schema validation to handle relative refs by passing in the spec file path. This will only work when passing a file path to add_api and not a dict.
**Note:** This will only work with response validation disabled. There are more changes required to support local refs elsewhere in the code."

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -15,16 +15,17 @@ default_handlers = {
 }
 
 
-def resolve_refs(spec, store=None, handlers=None):
+def resolve_refs(spec, store=None, handlers=None, spec_url=''):
     """
     Resolve JSON references like {"$ref": <some URI>} in a spec.
     Optionally takes a store, which is a mapping from reference URLs to a
     dereferenced objects. Prepopulating the store can avoid network calls.
+    Passing a spec_url enables the use of references with relative URLs.
     """
     spec = deepcopy(spec)
     store = store or {}
     handlers = handlers or default_handlers
-    resolver = RefResolver('', spec, store, handlers=handlers)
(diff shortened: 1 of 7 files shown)


dtkav:
Thanks for the contribution!

What's needed for this to not break response validation?

Somewhat of an aside, when I was writing this I remember something about an "id" field that allows you to set the base_url from inside the spec. The code is [here](https://python-jsonschema.readthedocs.io/en/latest/_modules/jsonschema/validators/#RefResolver) if you want to look into it.

crunk1:
Sorry, that "Note: This will only work with response validation disabled. There are more changes required to support
local refs elsewhere in the code." was from the description from the original PR #543.

I'm new to the code base, so I'll have to lean on others to help answer that for me.  I'm thinking that note is no longer relevant, but I'm not sure what testing I need to do. Any help?

dtkav:
The main breakage I can think of is that the spec that is served to the UI/client is the raw unresolved json spec.

We'll need to figure out how to serve up all of the files so that the client can use them as well.

What I'm wondering is if the use case is (a) to help the developer break up a spec file, or (b) to serve multiple files to the client. I can't yet understand why serving multiple files to the client would be ideal.

Looking back through older tickets, I'll second that jinja templating is probably the best way of splitting up yaml specs if it's for developer convenience only.

cognifloyd:
Yes. I'd really like to improve working with swaggerui. I'm currently referencing https://zalando.github.io/problem/schema.yaml in my openapi specs, but swagger ui does not like that.

dtkav:
@cognifloyd is swagger-ui unable to resolve remote yaml?

PR #803, opened 2018-11-30
fix up array style defaults

### LINKED ISSUE #801: Array of string parameter not passed correctly to server
### Description
Using connexion 2.0.2 and openapi 3.0 YAML file, array of string parameters not passed correctly.

### Expected behaviour
When running auto-generated browser UI, the browser shows type 'array\[string\] (query)' and prompts user to add array items one by one.  If user does that, it is expected a `list of str` would be passed to the server processing method (defined by operationId in the YAML file).

### Actual behaviour
Server processing method gets a list with only one element, regardless of length of array entered by user into browser UI.  A work around exists:  if array is entered into the first item, with elements separated by commas, then server processing method gets `list of str` as expected.

### Steps to reproduce
1. YAML file contains:

`
openapi: 3.0.0

....

paths:

  /end_point:

    post:
      operationId: server_processing_method
      tags:
        - Tag
      summary: Do something
      parameters:
        - name: inParam
          in: query
          description: "list of str"
          schema:
            type: array
            items:
              type: string
          required: true  
`

2. Run server on localhost.  Point browser to UI:  127.0.0.1:5000/ui/#

3. Enter parameter for `POST end_point`:  use 'add item' to specify an array with >1 elements, say "first","second","third" (each string as a separate item)

4. server_processing_method(inParam) will get inParam==["first"].  The rest of the elements are missing.

5. back to step 3, enter "first","second","third" into first item.  This time server_processing_method(inParam) will get inParam==["first","second","third"]

### Additional info:

Output of the commands:

- `python --version`
Python 3.6.5 :: Anaconda, Inc.
Version: 2.0.2

dtkav:
Fixes #801 

 - include the default array parsing styles from the openapi 3.0.2 spec

### DIFF
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -63,7 +63,17 @@ class AbstractURIParser(BaseDecorator):
         """
 
     @abc.abstractmethod
-    def _resolve_param_duplicates(self, values, param_defn):
+    def resolve_query(self, query_data):
+        """ Resolve cases where query parameters are provided multiple times.
+        """
+
+    @abc.abstractmethod
+    def resolve_path(self, path):
+        """ Resolve cases where path parameters include lists
+        """
+
+    @abc.abstractmethod
+    def _resolve_param_duplicates(self, values, param_defn, _in):
         """ Resolve cases where query parameters are provided multiple times.
             For example, if the query string is '?a=1,2,3&a=4,5,6' the value of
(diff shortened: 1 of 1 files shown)


dtkav:
Actually, sorry @jmcs for requesting review already - I need to add some tests firsts.
I've removed the `ready` label.

jmcs:
:+1:

Issue #691, opened 2018-09-21
`oneOf` does not work on `requestBody` when `oneOf` is nested under `#/components/schemas`


czchen:
### Description

`oneOf` in schema does not work on `requestBody` when this `oneOf` is nested under other schema defined in `#/components/schemas/`.

### Expected behaviour

`oneOf` shall work when it is nested under other schema defined in `#/components/schemas/`.

### Actual behaviour

The body does not contain the block use `oneOf`.

### Steps to reproduce

The following is python and schema used to reprocedure this issue. 

When we send a `post` request to `/bad` endpoint with body `{"oneof":{"name":"one"}}`. The response becomes `{}` instead of `{"oneof":{"name":"one"}}`.

The response shall be the same as request because they use the same schema, and the request is not rejected due to schema error.

```
openapi: 3.0.0
info:
  title: title
  version: 0.0.0
paths:
  /bad:
    post:
      summary: test
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/nested'
      responses:
        200:
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/nested'
      operationId: test.post
components:
  schemas:
    nested:
      type: object
      properties:
        oneof:
          $ref: '#/components/schemas/oneof'
    oneof:
          discriminator:
            propertyName: name
            mapping:
              one: '#/components/schemas/one'
              two: '#/components/schemas/two'
          oneOf:
            - $ref: '#/components/schemas/one'
            - $ref: '#/components/schemas/two'
    one:
      type: object
      required:
        - name
      properties:
        name:
          type: string
          pattern: ^one$
    two:
      type: object
      required:
        - name
      properties:
        name:
          type: string
          pattern: ^two$
```

```
import connexion


def post(body):
    print('body = {}'.format(body))
    return body, 200, {}


app = connexion.App(__name__)
app.add_api('test.yml', strict_validation=True, validate_responses=True)
app.run()
```

### Additional info:

Output of the commands:

```
$ python --version
Python 3.6.4+
$ grep connexion Pipfile
connexion = {git = "https://github.com/zalando/connexion", ref = "62582643cbfca65b0cf0d54755f81f0e1a7d1f82"}
```

dtkav:
Interesting! Thanks for filing the issue. I'll try to reproduce it and report back.

dtkav:
It looks like the problem is here: 
[...]

PR #815, opened 2018-12-10
Fix passing empty object body

cziebuhr:
Having a json request with an empty object body `{}` resulted in having `None` passed as body argument.

I'm not quite sure about the fix for openapi 3, probably the whole funtion there needs some reworking. Why is this whole body argument sanitizing needed which is not needed for swagger 2? Is it because of form-data? Then we probably should distinguish between json and form-data by content-type header and not by content-inspection. Also see https://github.com/zalando/connexion/issues/691#issuecomment-444892054


hjacobs:
LGTM

connexion/operations/openapi.py
```diff
@@ -256,7 +256,8 @@ def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
         # see: https://github.com/OAI/OpenAPI-Specification/blame/3.0.2/versions/3.0.2.md#L2305
         additional_props = self.body_schema.get("additionalProperties", True)
 
-        body = body or default_body
+        if body is None:
+            body = default_body
```
dtkav: Can you move line 267-268 up here instead?
I think we may need the deepcopy.

dtkav:
Hey @cziebuhr - with regards to this bug, your fix makes sense to me, but I think you can move some code around instead of duplicating it (also I think you need the deepcopy).

In regards to your larger question about passing the object data in directly, I think a good approach would be to come up with several tests that represent the desired behavior, discuss them, and then come up with a fix that passes the tests.

 in swagger 2 `form` and `body` data were modeled as separate things, but in openapi 3 they are all part of the `requestBody`. In swagger2 it was impossible to even validate the body (only form data could be validated), but in openapi3 it is possible. This is a great feature, but it means more work for us!

Ideally all of the validation happens much further upstream than this. Maybe with `additionalProperties: False` this is possible (I haven't tried it yet). Then we could just pass in the body directly, and have confidence that it is correct.

It's also worth digging up the history of why the sanitation was added in the first place. It was before my time.

dtkav:
Ok, let's merge this as an incremental improvement and we'll cleanup that section with the added help of your tests.

dtkav:
Nice work!

PR #816, opened 2018-12-10
#594 - add source key name to json validation error message

### LINKED ISSUE #594: Get key of field causing the validation error
### Description
Is it possible to seethe field that caused the validation error in the error message?

### Expected behaviour
Example: The client entered an unknown enum 'C' for field 'letter'. Enum definition: ['A', 'B']
Error returned by Validation:
{
  "detail": "Validation of field 'letter' failed. 'C' is not one of ['A', 'B']",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}

### Actual behaviour
{
  "detail": "'C' is not one of ['A', 'B']",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}

### Steps to reproduce
Define a body like this:

...
letter:
        type: 'string'
        enum: ['A', 'B']
...

### Additional info:

Output of the commands:

- `python --version` Python 3.6.3
- `pip show connexion | grep "^Version\:"` 1.1.9

lumikanta:
Fixes #594  .

 - include json key with full path into validation error message


connexion/decorators/validation.py
```diff
@@ -182,10 +182,11 @@ def validate_schema(self, data, url):
         try:
             self.validator.validate(data)
         except ValidationError as exception:
-            logger.error("{url} validation error: {error}".format(url=url,
+            logger.error("{url} - '{path}' - validation error: {error}".format(url=url,
```
dtkav: can you reformat this to fit within 80 characters wide please?
something like
```
err = "{url} - '{path}' - validation error: {error}"
err_path = ".".join(exception.path)
err = err.format(url=url, path=err_path, error=exception.message)
```

connexion/decorators/validation.py
```diff
@@ -182,10 +182,11 @@ def validate_schema(self, data, url):
...
+                                                                  path=".".join(exception.path),
                                                                   error=exception.message),
                          extra={'validator': 'body'})
-            return problem(400, 'Bad Request', str(exception.message))
+            return problem(400, 'Bad Request', "'{path}' - {message}".format(path=".".join(exception.path), message=exception.message))
```
dtkav: same here

dtkav:
It looks like `path` is not always available (or is set to ''), maybe add it to the end as additional debug info if it has a value.

advance512:
Any chance to move this forward?

PR #821, opened 2018-12-15
fix openapi3 path and query validation and tests

### LINKED ISSUE #811: No validation for path and query parameters with schemas
### Description
There doesn't seem to be any validation occurring for path or query parameters -- the schema does not get validated. I traced this through, and it looks like the schema actually gets overlooked in jsonschema/validators.py (so this may be an issue for jsonschema), because there is no validator for key "schema" in Draft4Validator, and therefore the entire schema is not validated. 

I used the following code to test:

yaml:
```
openapi: "3.0.0"

info:
  title: Hello World
  version: "1.0"
servers:
  - url: http://localhost:5000

paths:
  /test/{test_path_param}:
    post:
      summary: Test
      description: Test
      operationId: helloworld.test_example
      parameters:
        - name: test_path_param
          in: path
          description: Test path param
          required: true
          schema:
            type: string
            enum: [test, testing, tester]
      responses:
        200:
          description: response
          content:
            text/plain:
              schema:
                type: string
```

app code:
```
import connexion

def test_example(test_path_param):
    return test_path_param

app = connexion.FlaskApp(__name__, specification_dir='openapi/')
app.add_api('helloworld.yaml')
app.run(port=5000)
```

### Expected behaviour
When a POST request is made to `http://localhost:5000/test/INVALID`, 400 BAD REQUEST should be returned.

### Actual behaviour
When a POST request is made to `http://localhost:5000/test/INVALID`, 200 SUCCESS is returned, as well as "INVALID", which is the path parameter that was passed in.

### Steps to reproduce
See code above.

### Additional info:

Output of the commands:

- `python --version` 
Python 2.7.15rc1
Version 2.0.2

dtkav:
Fixes #811 

 - use 'schema' in param_defn if available in validate_parameter(...)

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -240,6 +240,7 @@ class ParameterValidator(object):
                 return str(e)
 
             param = copy.deepcopy(param)
+            param = param.get('schema', param)
             if 'required' in param:
                 del param['required']
             try:
--- a/tests/decorators/test_validation.py
+++ b/tests/decorators/test_validation.py
@@ -18,15 +18,28 @@ def test_get_valid_parameter_with_required_attr():
     assert result is None
 
 
+def test_get_valid_path_parameter():
+    param = {'required': True, 'schema': {'type': 'number'}, 'name': 'foobar'}
+    result = ParameterValidator.validate_parameter('path', 20, param)
+    assert result is None
+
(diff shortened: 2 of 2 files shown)


jmcs:
:+1:

PR #823, opened 2018-12-16
Serve correct openapi spec basepath when path is altered by reverse-proxy

### LINKED ISSUE #392: HTTPD mod_wsgi aliases
### Description
When using connexion with httpd and mod_wsgi aliases the swagger.json URL is not being generated correctly.

In order to have more than one mod_wsgi script running on a single httpd instance aliases are used. For example...

 _WSGIScriptAlias /myalias /path/to/my/mod_wsgi/script.wsgi_

So the URL for this mod_wsgi script would be...

_www.mydomain.com/myalias_

Now making the swagger base_path='/api' the ui URL would be...

_www.mydomain.com/myalias/api/ui_

This all works perfectly fine accept when the swagger.json url is generated. Connexion does not know anything about  the alias that is configured in the httpd.conf file and so it is not using it to generates the swagger.json URL so the URL ends up looking like this...

_www.mydomain.com/api/swagger.json_

which is not correct. In this case it should be...

_www.mydomain.com/myalias/api/swagger.json_

If you manually set the swagger.json path correctly in the base_url input field things will work fine.

Everything also works fine when you set the WSGIScript Alias to something like this...

_WSGIScriptAlias / /path/to/my/mod_wsgi/script.wsgi_

This is effectively like not having an alias, However in many situations this is not possible as you can only set one of these blank alias per httpd virtual host.  

I have been reading through the connexion source but have not come up with a good fix for this yet.

### LINKED ISSUE #527: Mounting flask under subpath through nginx
Sorry if this is obvious and I just missed how to do it. My flask app is the rest API so http://localhost:5000/swagger.json is correct. However, if I mount it under /api via a proxy_pass in nginx, the swagger ui is still looking for /swagger.json instead of /api/swagger.json

Setting base_path relocates the whole api so I end up with /api/api/my-rest-call once mounted in nginx.

Is there a way to specify the path where the swagger.json file is located? For now I created a rewrite rule on nginx but that is not a long term solution in case multiple Connexion apps are running under the same nginx instance.

### LINKED ISSUE #820: Need variable to alter spec path in Swagger UI Explore bar
I'm deploying an API written in Swagger (OAS3) behind AWS API Gateway. I can't import it into API Gateway because it seems as if they still only support 2.0.

That's fine. I don't mind that so much, but the problem is it messes with my Swagger UI. API Gateway puts the stage name in the URL but the swagger/connexion app isn't able to "see" this. As a result the Explore bar has the wrong path to the spec file.

It loads up with `/swagger_url/openapi.spec` but needs to be changed to `/aws_stage_name/swagger_url/openapi.spec` in order to load properly in the console.

I took a look at the ui stuff myself but am not a front end guy so had absolutely no idea what I was looking at. :)

dtkav:
Fixes #820 
Fixes #392 
Fixes #527 

 - change swagger-ui, and openapi spec location to build the path using `url_for`
 - modify the openapi spec `basePath` or `servers` block to include the reversed api path
 - modify the swagger-ui page to load the reversed openapi spec using `url_for`
 - add a openapi3/reverseproxy example
 - add a openapi3/reverseproxy_aiohttp example

Thanks to @Jyhess for the collaboration, tests, and review!

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -35,6 +35,15 @@ class FlaskApi(AbstractAPI):
         logger.debug('Adding spec json: %s/%s', self.base_path,
                      self.options.openapi_spec_path)
         endpoint_name = "{name}_openapi_json".format(name=self.blueprint.name)
+
+        def spec_for_url():
+            """ Modify base_path in the spec based on incoming url
+                This fixes problems with reverse proxies changing the path.
+            """
+            openapi_json_route = ".".join([self.blueprint.name, endpoint_name])
+            base_path = flask.url_for(openapi_json_route).rsplit("/", 1)[0]
+            return flask.jsonify(self.specification.with_base_path(base_path).raw)
+
         self.blueprint.add_url_rule(self.options.openapi_spec_path,
                                     endpoint_name,
                                     lambda: flask.jsonify(self.specification.raw))
@@ -273,8 +282,14 @@ class InternalHandlers(object):
 
         :return:
         """
+        openapi_json_route_name = "{blueprint}.{prefix}_openapi_json"
+        escaped = self.base_path.replace(".", "_")
+        openapi_json_route_name = openapi_json_route_name.format(
+            blueprint=escaped,
+            prefix=escaped
+        )
         template_variables = {
-            'openapi_spec_url': (self.base_path + self.options.openapi_spec_path)
+            'openapi_spec_url': flask.url_for(openapi_json_route_name)
         }
         if self.options.openapi_console_ui_config is not None:
             template_variables['configUrl'] = 'swagger-ui-config.json'
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -27,6 +27,13 @@ class FlaskApi(AbstractAPI):
         self.blueprint = flask.Blueprint(endpoint, __name__, url_prefix=self.base_path,
                                          template_folder=str(self.options.openapi_console_ui_from_dir))
 
+    def _spec_for_prefix(self):
+        """ Modify base_path in the spec based on incoming url
+            This fixes problems with reverse proxies changing the path.
+        """
+        base_path = flask.url_for(flask.request.endpoint).rsplit("/", 1)[0]
+        return self.specification.with_base_path(base_path).raw
+
     def add_openapi_json(self):
         """
         Adds spec json to {base_path}/swagger.json
@@ -36,17 +43,12 @@ class FlaskApi(AbstractAPI):
                      self.options.openapi_spec_path)
         endpoint_name = "{name}_openapi_json".format(name=self.blueprint.name)
 
-        def spec_for_url():
-            """ Modify base_path in the spec based on incoming url
-                This fixes problems with reverse proxies changing the path.
-            """
-            openapi_json_route = ".".join([self.blueprint.name, endpoint_name])
-            base_path = flask.url_for(openapi_json_route).rsplit("/", 1)[0]
-            return flask.jsonify(self.specification.with_base_path(base_path).raw)
+        def get_json_spec():
+            return flask.jsonify(self._spec_for_prefix())
 
         self.blueprint.add_url_rule(self.options.openapi_spec_path,
                                     endpoint_name,
-                                    lambda: flask.jsonify(self.specification.raw))
+                                    get_json_spec)
 
     def add_openapi_yaml(self):
         """
@@ -56,7 +58,8 @@ class FlaskApi(AbstractAPI):
         if not self.options.openapi_spec_path.endswith("json"):
             return
 
-        openapi_spec_path_yaml = self.options.openapi_spec_path[:-len("json")] + "yaml"
+        openapi_spec_path_yaml = \
+            self.options.openapi_spec_path[:-len("json")] + "yaml"
         logger.debug('Adding spec yaml: %s/%s', self.base_path,
                      openapi_spec_path_yaml)
         endpoint_name = "{name}_openapi_yaml".format(name=self.blueprint.name)
@@ -66,7 +69,7 @@ class FlaskApi(AbstractAPI):
             lambda: FlaskApi._build_response(
                 status_code=200,
                 mimetype="text/yaml",
-                data=yamldumper(self.specification.raw)
+                data=yamldumper(self._spec_for_prefix())
             )
         )
 
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -300,11 +300,7 @@ class InternalHandlers(object):
         return flask.jsonify(self._spec_for_prefix())
 
     def get_yaml_spec(self):
-        lambda: FlaskApi._build_response(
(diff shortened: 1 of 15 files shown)


bhechinger:
That fixes the Explore bar nicely so the spec is loaded into the UI Console.

That also sets the link to the spec file correctly.

The `Try it out` section still has the wrong URL, however.

dtkav:
Ah ok - great thanks for the feedback. I need to set up a proper test environment :)
@jyhess also has a similar PR at #826 if you want to try that out and provide feedback.

dtkav:
hey @bhechinger - can you try out the latest example, and let me know if it is working for you?

dtkav:
The build is broken because of #844

dtkav:
I'm stuck on a single aiohttp test in python 3.4, where it seems to think that aiohttp is installed after calling sys.modules.pop("aiohttp").
Does anyone know what's going on there? I'm not able to reproduce it locally.

Jyhess:
> I'm stuck on a single aiohttp test in python 3.4, where it seems to think that aiohttp is installed after calling sys.modules.pop("aiohttp").
> Does anyone know what's going on there? I'm not able to reproduce it locally.

Not reproduced on my side either, but you can try another strategy with mocker:

```
def test_run_with_aiohttp_not_installed(mock_app_run, spec_file, mocker):
    import sys
    mocker.patch.dict(sys.modules, {'aiohttp': None})
    runner = CliRunner()

    # missing aiohttp
    result = runner.invoke(main,
                           ['run', spec_file, '-f', 'aiohttp'],
                           catch_exceptions=False)

    assert 'aiohttp library is not installed' in result.output
    assert result.exit_code == 1
```

dtkav:
@Jyhess thanks for your help. I was able to sort it out by using mock.

dtkav:
@jmcs can you take a look at this one please?

examples/openapi3/reverseproxy_aiohttp/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+This example demonstrates how to run a connexion application behind a path-altering reverse proxy.
+
+You can either set the path in your app, or set the ``X-Forwarded-Path`` header.
+
+Running:
+
+.. code-block:: bash
+
+    $ sudo pip3 install --upgrade connexion[swagger-ui]  # install Connexion from PyPI
```
Jyhess: Better to specify extra dependency:
`sudo pip3 install --upgrade connexion[swagger-ui] aiohttp-remotes`

dtkav: good catch! oops

ronlut:
Should that fix the last comment on #[692](https://github.com/zalando/connexion/issues/692#issuecomment-444708362) ?
I have the same problem exactly.

thomaschristensen:
I have the same problem as ronlut, but can't figure out how "url_for" should be applied to make this pull request version work. I have pulled it into my project, but it still does not work with the AWS staging environment.

ronlut:
BTW, I described my problem in details here: [link](https://community.smartbear.com/t5/Swagger-Open-Source-Tools/Can-t-get-url-configuration-to-work-with-Swagger-UI-on-AWS/m-p/178069#M120)

(if anyone needs more details)

dtkav:
@ronlut - yes, this is intended to fix that problem.
@thomaschristensen - did you see the example included in this PR? You need to wrap the wsgi app to make url_for work. In particular we need to populate the environment with `SCRIPT_PATH`. https://github.com/zalando/connexion/pull/823/files#diff-09f8c7fbc763384bc2fbf29a4a93cd19R15

examples/openapi3/reverseproxy/app.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+'''
+example of connexion running behind a path-altering reverse-proxy
+
+NOTE this demo is not secure by default!!
```
rafaelcaricio: People will not read this and copy and paste the code. I would suggest to use `logging.warn` in the code as well to log this information. Whoever will do copy&paste will have more chances of seeing this.

connexion/apis/flask_api.py
```diff
@@ -300,9 +312,16 @@ def console_ui_home(self):
 
         :return:
         """
+        openapi_json_route_name = "{blueprint}.{prefix}_openapi_json"
+        escaped = self.base_path.replace(".", "_")
```
rafaelcaricio: Maybe we should use `flask_utils.flaskify_endpoint` here?

connexion/apis/flask_api.py
```diff
@@ -35,9 +42,13 @@ def add_openapi_json(self):
         logger.debug('Adding spec json: %s/%s', self.base_path,
                      self.options.openapi_spec_path)
         endpoint_name = "{name}_openapi_json".format(name=self.blueprint.name)
+
+        def get_json_spec():
```
rafaelcaricio: Looks like to me that this fits in the class `InternalHandlers`, located in line 300 of this file.

connexion/apis/flask_api.py
```diff
@@ -300,9 +312,16 @@ def console_ui_home(self):
 
         :return:
         """
+        openapi_json_route_name = "{blueprint}.{prefix}_openapi_json"
```
rafaelcaricio: In line 44 and here we do similar formatting for the register and to get the URL for the endpoint. Maybe we can do it once? I would like to avoid duplicated "..._openapi_json" string. Something could tie this two together closer, so we don't break things later. I know we have tests covering this, but if possible would be nice to remove this duplication. Sorry if this seems over-engineering.

rafaelcaricio:
Maybe you can rebase with master? We fixed the tests. :)

Jyhess:
I rebased this branch on master, with some fix for @rafaelcaricio comments.
As I don't have permission to push to Daniel's repository, you can find this change on my own repository: https://github.com/Jyhess/connexion/tree/dynamic_ui_path

RobbeSneyders:
Is there a way to get these changes merged? Either by updating @dtkav branch, or opening a new PR for @Jyhess branch?

danie1k:
Aren't this PR and #948 a duplicates?

dtkav:
Nice to see some merge activity happening again.
I've rebased onto master and included the improvements from @Jyhess (thanks!).

connexion/apis/flask_api.py
dtkav: Hey @rafaelcaricio  - really appreciate your comments.
I agree with your point here, but I couldn't come up with a good idea for how to make it happen. Maybe you could provide a more specific example of what you have in mind?

dtkav:
Hey @hjacobs if you find some time - these changes will enable people to use connexion behind a path-altering reverse proxy (like AWS API Gateway).

dtkav:
@hjacobs can you please take a look at this fix for path-altering reverse proxies?

dtkav:
@rafaelcaricio are you happy with the changes to address your comments?

dtkav:
@hjacobs can you please have a look?

hjacobs:
:+1:

PR #825, opened 2018-12-17
Fix case sensitive headers

### LINKED ISSUE #824: Failed to get Authorization header when passed in lower case
### Description
If I setup an application with aiohttp using tests/fixtures/aiohttp/openapi_secure.yaml
I receive a 401 Unauthorized when I use authorization in lower case:
`curl -H "authorization Bearer 100" localhost:8080/v1.0/greeting/jsantos`

### Expected behaviour
authorization header is usable in lower case

### Actual behaviour
Only Authorization header is usable with first letter in capital

### Additional info:

Output of the commands:

- `python --version`
Python 3.6.7
Version: 2018.0.dev1

Jyhess:
Fixes #824 .

 - Use aiohttp headers (in case insensitive dict) instead of copying them in a standard dict.

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -171,7 +171,7 @@ class AioHttpApi(AbstractAPI):
                      extra={'has_body': req.has_body, 'url': url})
 
         query = parse_qs(req.rel_url.query_string)
-        headers = {k.decode(): v.decode() for k, v in req.raw_headers}
+        headers = req.headers
         body = None
         if req.body_exists:
             body = yield from req.read()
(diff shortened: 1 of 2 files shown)


connexion/apis/aiohttp_api.py
```diff
@@ -171,7 +171,7 @@ def get_request(cls, req):
                      extra={'has_body': req.has_body, 'url': url})
 
         query = parse_qs(req.rel_url.query_string)
-        headers = {k.decode(): v.decode() for k, v in req.raw_headers}
+        headers = req.headers
```
dtkav: Does this work in aiohttp 2.3.10, and if not should we bump the required version in setup.py?

dtkav: Looks this feature is in 2.3.10.
@dutradda any reason to not use `.headers` for the CIMultiDictProxy?

PR #836, opened 2019-01-03
Update quickstart.rst

drfeinberg:
I kept getting the error that home contained no app, so this defined it and allowed the sample code to work.
You need the code...
application = app.app

Fixes # .

 -
 -
 -

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -28,6 +28,7 @@ Put your API YAML inside a folder in the root path of your application (e.g ``op
 
     app = connexion.FlaskApp(__name__, specification_dir='openapi/')
     app.add_api('my_api.yaml')
+    application = app.app
     app.run(port=8080)
 
 


dtkav:
Thanks for the PR.
While I think it's useful to show something like
```
# flask_app = app.app
```
It's not required at all to make the example work, and seems to be specific to your use case.
Does that sound right?

docs/quickstart.rst
```diff
@@ -28,6 +28,7 @@ Put your API YAML inside a folder in the root path of your application (e.g ``op
 
     app = connexion.FlaskApp(__name__, specification_dir='openapi/')
     app.add_api('my_api.yaml')
+    application = app.app
```
dtkav: I think `flask_app = app.app` would probably be more helpful

dtkav:
Hey @drfeinberg I'm going to close this PR as I haven't heard back from you.
Please feel free to re-open if you're up for addressing my comments.

PR #844, opened 2019-01-09
set max aiohttp version to 3.5.1 (until build is fixed)

dtkav:
Fixes master

Master is broken because of an aiohttp update upstream:
https://travis-ci.org/zalando/connexion/builds/469438675

 - add an upper bound on the aiohttp version (<3.5.2)

I want to confirm that this fixes the build, and then we can investigate how to support the newest version.


hjacobs:
:+1:

PR #828, opened 2018-12-18
WIP: Run framework agnostic tests

panpann:
Hello,

# context
a couple weeks ago, I created a branch to add [quart](https://github.com/pgjones/quart) support to connexion. I managed to implement the app and api but I struggled when it came time to test my work.
I wasn't familiar at all with the connexion codebase and didn't feel able to add tests.  
Also this was frustrating because Flask and Quart follow almost the same API (except async for quart). I just wanted to use all tests that request the tests API via `test_client()`.  

I like your idea of API first principle, it makes APIs consistent by taking the time to think. And to write an openapi before starting coding helps to normalize data sent and returned by the API. It is easy to implement well documented API contracts and keep the code clean and business focused.
 
In the future, I would like to have the possibility to choose any python framework and to use connexion on it.
I think that to test the same way each implementation could help contributors to add more frameworks to connexion, if the contributor only has to implement the interface and see a concrete error if its code is wrong. It would also ensure that whatever server I'm using, the API is consistent and responds the same way.

# proposed changes
I've started working on it,
1. I added an API `AbstractApp.test_client()` which returns an `AbstractClient`, each web server has to implement 1 **synchronous** method to request the API and returns an enhanced version of ConnexionResponse (just added helpers like response.json, response.text).
For now the API to send request via this AbstractClient respects the client's flask API. 
2. I've changed all tests that were using app.app.test_client() to app.test_client() and all tests passed.
3. I created the test_client for the AioHttpApp
4. I am now trying to pass all tests with AioHttpApp, and I encountered [this issue](https://github.com/zalando/connexion/issues/578). Are you okay if I'm making changes to respect the same return API that flask ? 

Are you okay with changes I'm proposing in this pull request ?

The work currently won't pass the CI, I'm creating this pull request to share my idea with you and have your opinion on it

PR #760, opened 2018-11-08
WIP: redo validation by content type

dtkav:
This is a WIP, I'd like to make this a bit more pluggable, so that it is straightforward to add content-type validation handlers.

Related to #655 #755 

Changes proposed in this pull request:
 - remove `all_json` requirement that caused any non-json content types to preclude validating json

PR #849, opened 2019-01-14
Return tuple for aiohttp

Jyhess:
First step of #828

 - Support tuples as return values for aiohttp handlers

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -401,7 +401,7 @@ class AbstractAPI(object):
             status_code = status_code.value
 
         if data is not None:
-            body = cls._jsonify_data(data, mimetype)
+            body, mimetype = cls._serialize_data(data, mimetype)
         else:
             body = data
 
@@ -414,10 +414,11 @@ class AbstractAPI(object):
                          **extra_context
                      })
 
-        return body, status_code
+        return body, status_code, mimetype
 
     @classmethod
-    def _jsonify_data(cls, data, mimetype):
+    def _serialize_data(cls, data, mimetype):
+        # TODO: Harmonize with flask_api. Currently this is the backwards compatible with aiohttp_api._cast_body.
         if not isinstance(data, bytes):
             if isinstance(mimetype, str) and is_json_mimetype(mimetype):
                 body = cls.jsonifier.dumps(data)
@@ -427,7 +428,7 @@ class AbstractAPI(object):
                 body = str(data)
         else:
             body = data
-        return body
+        return body, mimetype
 
     def json_loads(self, data):
         return self.jsonifier.loads(data)
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -326,7 +326,7 @@ class AbstractAPI(object):
         if isinstance(response, ConnexionResponse):
             # If body in ConnexionResponse is not byte, it may not pass schema validation.
             # In this case, rebuild response with aiohttp to have consistency
-            if response.body is None or isinstance(response.body, six.binary_type):
+            if response.body is None or isinstance(response.body, bytes):
                 return response
             else:
                 response = cls._build_response(
@@ -395,10 +395,10 @@ class AbstractAPI(object):
 
     @classmethod
     def _jsonify_data(cls, data, mimetype):
-        if not isinstance(data, six.binary_type):
-            if isinstance(mimetype, six.string_types) and is_json_mimetype(mimetype):
+        if not isinstance(data, bytes):
+            if isinstance(mimetype, str) and is_json_mimetype(mimetype):
                 body = cls.jsonifier.dumps(data)
-            elif isinstance(data, six.text_type):
+            elif isinstance(data, str):
                 body = data
             else:
                 body = str(data)
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -2,6 +2,7 @@ import abc
 import logging
 import pathlib
 import sys
+import warnings
 from enum import Enum
 
 import six
@@ -425,6 +426,15 @@ class AbstractAPI(object):
             elif isinstance(data, str):
                 body = data
             else:
+                warnings.warn(
+                    "Implicit (aiohttp) serialization with str() will change in the next major version. "
+                    "This is triggered because a non-JSON response body is being stringified. "
+                    "This will be replaced by something that is mimetype-specific and may "
+                    "serialize some things as JSON or throw an error instead of silently "
+                    "stringifying unknown response bodies. "
+                    "Please make sure to specify media/mime types in your specs.",
+                    FutureWarning  # a Deprecation targeted at application users.
+                )
                 body = str(data)
         else:
             body = data
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -312,6 +312,9 @@ class AbstractAPI(object):
 
         if isinstance(response, tuple):
             len_response = len(response)
+            if len_response == 1:
+                data, = response
+                return cls._build_response(mimetype=mimetype, data=data, extra_context=extra_context)
             if len_response == 2:
                 if isinstance(response[1], (int, Enum)):
                     data, status_code = response
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -238,7 +238,9 @@ class AbstractAPI(object):
     @abc.abstractmethod
     def get_response(self, response, mimetype=None, request=None):
         """
-        This method converts the ConnexionResponse to a user framework response.
+        This method converts a handler response to a framework response.
+        The response can be a ConnexionResponse, an operation handler, a framework response or a tuple.
+        Other type than ConnexionResponse are handled by `cls._response_from_handler`
         :param response: A response to cast.
         :param mimetype: The response mimetype.
         :param request: The request associated with this response (the user framework request).
@@ -249,10 +251,31 @@ class AbstractAPI(object):
 
     @classmethod
     @abc.abstractmethod
+    def _response_from_handler(cls, response, mimetype):
+        """
+        Create a framework response from the operation handler data.
+        An operation handler can return:
+        - a framework response
+        - a body (str / binary / dict / list), a response will be created
+            with a status code 200 by default and empty headers.
+        - a tuple of (body: str, status_code: int)
+        - a tuple of (body: str, status_code: int, headers: dict)
+        :param response: A response from an operation handler.
+        :type response Union[Response, str, Tuple[str, int], Tuple[str, int, dict]]
+        :param mimetype: The response mimetype.
+        :type mimetype: str
+        :return A framwork response.
+        :rtype Response
+        """
+
+    @classmethod
+    @abc.abstractmethod
     def get_connexion_response(cls, response, mimetype=None):
         """
         This method converts the user framework response to a ConnexionResponse.
+        It is used after the user returned a response to give it to response validators.
         :param response: A response to cast.
+        :param mimetype: The response mimetype.
         """
 
     def json_loads(self, data):
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -245,12 +245,10 @@ class AbstractAPI(object):
         This method converts a handler response to a framework response.
         This method should just retrieve response from handler then call `cls._get_response`.
         It is mainly here to handle AioHttp async handler.
-        :param response: A response to cast.
+        :param response: A response to cast (tuple, framework response, etc).
         :param mimetype: The response mimetype.
+        :type mimetype: Union[None, str]
         :param request: The request associated with this response (the user framework request).
-
-        :type response: Framework Response
-        :type mimetype: str
         """
 
     @classmethod
@@ -259,12 +257,11 @@ class AbstractAPI(object):
         This method converts a handler response to a framework response.
         The response can be a ConnexionResponse, an operation handler, a framework response or a tuple.
         Other type than ConnexionResponse are handled by `cls._response_from_handler`
-        :param response: A response to cast.
+        :param response: A response to cast (tuple, framework response, etc).
         :param mimetype: The response mimetype.
+        :type mimetype: Union[None, str]
         :param extra_context: dict of extra details, like url, to include in logs
-
-        :type response: Framework Response
-        :type mimetype: str
+        :type extra_context: Union[None, dict]
         """
         if extra_context is None:
             extra_context = {}
@@ -299,7 +296,7 @@ class AbstractAPI(object):
         - a tuple of (body: str, status_code: int)
         - a tuple of (body: str, status_code: int, headers: dict)
         :param response: A response from an operation handler.
-        :type response Union[Response, str, Tuple[str, int], Tuple[str, int, dict]]
+        :type response Union[Response, str, Tuple[str,], Tuple[str, int], Tuple[str, int, dict]]
         :param mimetype: The response mimetype.
         :type mimetype: str
         :param extra_context: dict of extra details, like url, to include in logs
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -4,14 +4,18 @@ import pathlib
 import sys
 
 import six
+from enum import Enum
 
+from ..decorators.produces import NoContent
 from ..exceptions import ResolverError
 from ..http_facts import METHODS
 from ..jsonifier import Jsonifier
+from ..lifecycle import ConnexionResponse
 from ..operations import make_operation
 from ..options import ConnexionOptions
 from ..resolver import Resolver
 from ..spec import Specification
+from ..utils import is_json_mimetype
 
 MODULE_PATH = pathlib.Path(__file__).absolute().parent.parent
 SWAGGER_UI_URL = 'ui'
@@ -239,18 +243,43 @@ class AbstractAPI(object):
     def get_response(self, response, mimetype=None, request=None):
         """
         This method converts a handler response to a framework response.
-        The response can be a ConnexionResponse, an operation handler, a framework response or a tuple.
-        Other type than ConnexionResponse are handled by `cls._response_from_handler`
+        This method should just retrieve response from handler then call `cls._get_response`.
+        It is mainly here to handle AioHttp async handler.
         :param response: A response to cast.
         :param mimetype: The response mimetype.
         :param request: The request associated with this response (the user framework request).
 
-        :type response: ConnexionResponse
+        :type response: Framework Response
         :type mimetype: str
         """
 
     @classmethod
-    @abc.abstractmethod
+    def _get_response(cls, response, mimetype=None, url=None):
+        """
+        This method converts a handler response to a framework response.
+        The response can be a ConnexionResponse, an operation handler, a framework response or a tuple.
+        Other type than ConnexionResponse are handled by `cls._response_from_handler`
+        :param response: A response to cast.
+        :param mimetype: The response mimetype.
+        :param url: The url to write in logs
+
+        :type response: Framework Response
+        :type mimetype: str
+        """
+        logger.debug('Getting data and status code',
+                     extra={
+                         'data': response,
+                         'data_type': type(response),
+                         'url': url
+                     })
+
+        if isinstance(response, ConnexionResponse):
+            framework_response = cls._connexion_to_framework_response(response, mimetype)
+        else:
+            framework_response = cls._response_from_handler(response, mimetype)
+        return framework_response
+
+    @classmethod
     def _response_from_handler(cls, response, mimetype):
         """
         Create a framework response from the operation handler data.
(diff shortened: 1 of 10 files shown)


panpann:
Hi, I just reviewed your code, firsly thanks to help on this one ! :)
I don't know what @dtkav thinks about it maybe he will merge it but this is not going in the way I want in #828 .  
To explain myself, what I want in connexion is as little as possible work to do when you create a new App. 
For me validation of handler response is not something specific to FlaskApp or AioHttpApp. It has to be handled by one central and well tested piece of code, shared by all Apps to ensure consistency. This is why I've done it [here](https://github.com/panpann/connexion/blob/run-same-tests-for-all-apps/connexion/operations/validation.py).
Then `App._response_from_handler` is here to make some boilerplate before validating the handler's response and to create a Framework specific response from the data returned by this validation.

What I explained juste before applies also for the tests. In my PR, I want same tests to be applied for each App, to not create duplication.  In my point a view, This project needs a base of tests that can be run on all tests and very specific App tests (like using a custom encoder in Flask by passing it to the flask object)    

Here is how I see this functionality ! :)

Jyhess:
@panpan I got your point. I did more factorization between both Flask and AioHttp. Let me know what you think.
I let you working on test side ;)

dtkav:
@Jyhess I looks like the CI is failing because your are trying to read the response body (bytes) as a string. I think you can use `response.json` instead.

cognifloyd:
I would like to use tuple returns with aiohttp. What is needed to move this forward? More code review?

jmcs:
@cognifloyd sorry for dropping the ball here. Can you fix the merge conflicts?

cognifloyd:
Sure. Here's the merge commit:
https://github.com/cognifloyd/connexion/commit/b4b72e7805d915d7632868b5127a7596ddb09699
I don't have commit rights in the connexion repo, so I can't update this PR.
`git mergetool` was able to auto-resolve the differences in the two files.

Jyhess:

hjacobs:

dtkav:
Hi @Jyhess - I tried to resolve the conflicts against master. Sorry if this messed up your workflow!

connexion/apis/flask_api.py
```diff
@@ -278,6 +228,15 @@ def get_request(cls, *args, **params):
...
         return request
 
+    @classmethod
+    def _jsonify_data(cls, data, mimetype):
+        # TODO: to discuss: Using jsonifier for all type of data, even when mimetype is not json is strange. Why ?
+        if (isinstance(mimetype, six.string_types) and is_json_mimetype(mimetype)) \
+                or not (isinstance(data, six.binary_type) or isinstance(data, six.text_type)):
+            return cls.jsonifier.dumps(data)
+
+        return data
```
cognifloyd: Do we need to resolve this before merging this PR?

connexion/apis/aiohttp_api.py
```diff
@@ -316,66 +316,57 @@ def get_response(cls, response, mimetype=None, request=None):
...
 
-        if isinstance(response, web.StreamResponse):
-            logger.debug('Got stream response with status code (%d)',
-                         response.status, extra={'url': url})
-        else:
-            logger.debug('Got data and status code (%d)',
-                         response.status, extra={'data': response.body, 'url': url})
+        # TODO: I let this log here for full compatibility. But if we can modify it, it can go to _get_response()
+        logger.debug('Got stream response with status code (%d)',
+                     response.status, extra={'url': url})
```
cognifloyd: Why would we need to retain a log line for compatibility? Would we ever consider a debug-level log entry as "API"?

connexion/apis/abstract.py
```diff
@@ -238,23 +241,170 @@ def get_request(self, *args, **kwargs):
...
+                return cls._build_response(mimetype=mimetype, data=data, headers=headers)
+            elif len_response == 3:
+                data, status_code, headers = response
+                return cls._build_response(mimetype=mimetype, data=data, status_code=status_code, headers=headers)
+            else:
+                raise TypeError(
+                    'The view function did not return a valid response tuple.'
+                    ' The tuple must have the form (body), (body, status, headers),'
+                    ' (body, status), or (body, headers).'
+                )
```
cognifloyd: Where does this handle tuples like `(body)` with `len_response == 1`? Are those considered "framework responses" already?

cognifloyd:
I'm trying to figure out where the test error is coming from.

This test has `content_type` of `application/json` instead of `application/problem+json`:
https://github.com/zalando/connexion/blob/b0b83c4879d6fc3b5242b06a9659942decaa62ef/tests/aiohttp/test_aiohttp_errors.py#L99-L104

using this spec:
https://github.com/zalando/connexion/blob/1f734d56e814f81d57cd983a8ec3619852452fa4/tests/fixtures/problem/openapi.yaml#L61-L67

With this handler function:
https://github.com/zalando/connexion/blob/1f734d56e814f81d57cd983a8ec3619852452fa4/tests/fakeapi/hello.py#L358-L359

But I don't know what is supposed to raise a `ProblemException` for this kind of invalid response.

This is probably getting cast to string here (Line 403), but it did the same before this PR (line 311), so I don't know why these changes are triggering the error.
This PR:
https://github.com/zalando/connexion/blob/a9d6122423250d88afd286b68fb3673a701edd4a/connexion/apis/abstract.py#L395-L406
Before this PR:
https://github.com/zalando/connexion/blob/7a8403ffeaadb0ac849f7482bd12737c344dfb19/connexion/apis/aiohttp_api.py#L301-L313

Log lines on Travis
```
DEBUG:connexion.apis.abstract:Adding /v1.0/json_response_with_undefined_value_to_serialize...
DEBUG:connexion.operations.openapi3:consumes: ['application/json']
DEBUG:connexion.operations.openapi3:produces: ['application/json']
DEBUG:connexion.apis.aiohttp_api:... Adding GET -> fakeapi.hello.get_invalid_response
DEBUG:connexion.operations.abstract:... Response validation enabled.
DEBUG:connexion.operations.abstract:... Adding response decorator (<ResponseValidator>)
DEBUG:connexion.operations.abstract:... Produces: ['application/json']
DEBUG:connexion.operations.abstract:... Produces json
DEBUG:connexion.operations.abstract:... Adding produces decorator (<function AbstractOperation.__content_type_decorator.<locals>.<lambda> at 0x7f8a38ef5048>)
DEBUG:connexion.operations.secure:... Security: None
DEBUG:connexion.operations.abstract:... Adding security decorator (<function security_passthrough at 0x7f8a3a13bbf8>)
```
So validation is enabled, and it is suppose to produce `application/json` on success.
But the spec has no `content` or `schema` definition, so response validation always will pass.

It looks like this endpoint is actually returning `200`
```
DEBUG:connexion.apis.aiohttp_api:Getting data and status code
DEBUG:connexion.decorators.parameter:Function Arguments: []
DEBUG:connexion.apis.abstract:Getting data and status code
DEBUG:connexion.apis.aiohttp_api:Got stream response with status code (200)
INFO:aiohttp.access:127.0.0.1 [10/Nov/2019:18:20:56 +0000] "GET /v1.0/json_response_with_undefined_value_to_serialize HTTP/1.1" 200 179 "-" "Python/3.7 aiohttp/3.6.2"
```

So, I'm lost and don't know where else to look. Next step I guess would be to hook up a debugger to master, run this test, and see what throws the error to generate the 500 an `application/problem+json` response. I don't have time to do that right now, so I'll just leave this as a note for someone else to continue investigating.

connexion/apis/abstract.py
```diff
@@ -238,23 +241,170 @@ def get_request(self, *args, **kwargs):
...
+        else:
+            return cls._build_response(mimetype=mimetype, data=response)
+
+    @classmethod
+    def get_connexion_response(cls, response, mimetype=None):
+        """ Cast framework dependent response to ConnexionResponse used for schema validation """
+        if isinstance(response, ConnexionResponse):
+            # If body in ConnexionResponse is not byte, it may not pass schema validation.
+            # In this case, rebuild response with aiohttp to have consistency
+            if response.body is None or isinstance(response.body, six.binary_type):
```
cognifloyd: connexion is now python 3 only, so six is no longer needed.

connexion/apis/abstract.py
```diff
@@ -238,23 +241,170 @@ def get_request(self, *args, **kwargs):
...
+        else:
+            body = data
+        return body, status_code
+
+    @classmethod
+    def _jsonify_data(cls, data, mimetype):
+        if not isinstance(data, six.binary_type):
+            if isinstance(mimetype, six.string_types) and is_json_mimetype(mimetype):
+                body = cls.jsonifier.dumps(data)
+            elif isinstance(data, six.text_type):
```
cognifloyd: ditto - six is no longer needed, so this can be bytes and str just like it was in aiohttp_api.py.

cognifloyd:
The equivalent Flask test is passing:
https://github.com/zalando/connexion/blob/b0b83c4879d6fc3b5242b06a9659942decaa62ef/tests/api/test_errors.py#L66-L68

So maybe it has something to do with how `AbstractApi._jsonify_data()` does not `.encode()` strings but the old `AioHttpApi._cast_body()` did? This is the Flask equivalent:
https://github.com/zalando/connexion/blob/a9d6122423250d88afd286b68fb3673a701edd4a/connexion/apis/flask_api.py#L231-L238

cognifloyd:
@Jyhess I hope we can work on this PR next. Are you available to help resolve some of my questions?

Jyhess:
@cognifloyd I will ping you on gitter when I can go back on this subject. Maybe this afternoon or tomorrow.

connexion/apis/aiohttp_api.py
cognifloyd: Ah. At this point the response is a framework response. So if we want to log the status code, then it needs to be under the framework specific *_api.py

cognifloyd: resolved

connexion/apis/flask_api.py
```diff
@@ -136,113 +135,64 @@ def get_response(cls, response, mimetype=None, request=None):
...
+        return flask.current_app.response_class(**kwargs)  # type: flask.Response
 
     @classmethod
     def _jsonify_data(cls, data, mimetype):
-        if (isinstance(mimetype, six.string_types) and is_json_mimetype(mimetype)) \
-                or not (isinstance(data, six.binary_type) or isinstance(data, six.text_type)):
+        # TODO: to discuss: Using jsonifier for all type of data, even when mimetype is not json is strange. Why ?
+        if (isinstance(mimetype, str) and is_json_mimetype(mimetype)) \
+                or not (isinstance(data, bytes) or isinstance(data, str)):
             return cls.jsonifier.dumps(data)
```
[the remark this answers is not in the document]
dtkav: I think connexion started out as a tool focused on json apis, and so there is some very old and very weird coupling with json.
In my opinion, the number one improvement I'd like to see from connexion is to decouple content types (see my work here https://github.com/zalando/connexion/pull/760 ), but it would mean a backwards incompatible change, and so a major version bump.

With respect to this particular section, I agree - it doesn't think it makes any sense to jsonify everything.
My guess though, is that because connexion is so forgiving about not specifying response content type, there are a bunch of users who are not specifying a json content type for their response, and are relying on this (and other) bugs/assumptions to get json out the other end.

I think the only thing to do in the context of this PR is leave it for now, and then if we can muster the effort - work together to decouple content type properly and do a major version bump.

I'm very interested to pick #760 back up, and to collaborate on this, assuming @hjacobs will have enough bandwidth to review the changes and merge them.

connexion/apis/abstract.py
```diff
@@ -238,22 +242,195 @@ def get_request(self, *args, **kwargs):
...
-        This method converts the ConnexionResponse to a user framework response.
+        This method converts a handler response to a framework response.
+        This method should just retrieve response from handler then call `cls._get_response`.
+        It is mainly here to handle AioHttp async handler.
         :param response: A response to cast.
         :param mimetype: The response mimetype.
         :param request: The request associated with this response (the user framework request).
 
-        :type response: ConnexionResponse
+        :type response: Framework Response
```
dtkav: I think this is supposed to be an actual type/class.

connexion/apis/flask_api.py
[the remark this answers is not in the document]
dtkav: Regarding what to do about the mismatch in aiohttp vs flask - maybe we can think about it from the point of view of the user. What would be the principle of least surprise?

On the one hand, we value correctness and having the same behavior between aiohttp and flask.
On the other hand, if we change the behavior in a way that isn't backwards compatible, then we should probably either be staging a major version bump, or at least understanding the breakage risk very well.

Maybe one way to understand the risk better would be to come up with an example application that returns different data based on this change - is it hard to trigger? how many people might it affect?

dtkav:
This is fantastic work @Jyhess and @cognifloyd - I'm inspired by your work on this project, and excited to feel the momentum!

It looks like the longer-term plan is to firm up this interface to allow for other frameworks (like quart, sanic).
I think it's worth discussing a few high level things (not blocking this PR)
1. Designing an internal API like this is _hard_, I think it could be valuable to at least demo a implementation of a third framework with this new interface. Proving that it would work with 3 frameworks is a good sign that it will work for many. Making it work for two doesn't force you to generalize as much (which I think is what led to the current design).
2. It was pretty high cost to support both aiohttp and flask when I introduced openapi3 support. What are some approaches for unifying the test infrastructure, and supporting the sync/async differences more generally?

I'd be excited to collaborate on long term vision and improvements assuming there's bandwidth for review and merging by folks at zalando.

connexion/apis/abstract.py
cognifloyd: I cleaned up the type hint comments a bit.

tests/aiohttp/test_get_response.py
```diff
@@ -0,0 +1,150 @@
...
+        yield from api.get_response((web.Response(text='foo', status=201, headers={'X-header': 'value'}), 200))
+    assert str(e.value) == "Cannot return web.StreamResponse in tuple. Only raw data can be returned in tuple."
+
+
+@asyncio.coroutine
+def test_get_response_from_dict(api):
+    response = yield from api.get_response({'foo': 'bar'})
+    assert isinstance(response, web.Response)
+    assert response.status == 200
+    assert response.body == b"{'foo': 'bar'}"
```
dtkav: You wouldn't expect this to be json?

cognifloyd: Right. I would not expect this to be json. Train of thought follows:
- Everything in this file is testing against an api that has only one endpoint `/greeting/{name}` -> `fakeapi.aiohttp_handlers.aiohttp_post_greeting` which returns a dict.
- But it's not actually using the handler, it's accessing get_response directly.
- So mimetype is not coming from the spec here, which means there is no assumption that the content will be json.
- which means if you pass an object (in this case a dict) it will be stringified by `_jsonify_data`.
- the implementation of `_jsonify_data` in `abstract` is basically doing the same thing that `_cast_body` did in `aiohttp_api`
- Note that this body is using `'` not `"` so it is not valid json. it is just a stringified dict.
- So, yes I would expect (without changing behavior) this to merely stringify the dict as no mimetype / content type is defined.
- Consider the next test where a mimetype is explicitly defined, and therefore produces json data.

dtkav: I think I can follow your logic here - I have a minor pragmatic hangup which is that casting a dict to a string is kind of the least-useful way to serialize that object.
You might argue in practice that folks would notice this, and then include the right content-type, but it seems like an unfriendly default.
Consider the following flask app, which defaults to returning json
```
  import flask                                                                                                                                                                                                      
      
  app = flask.Flask(__name__)    
      
  @app.route("/")    
  def a():    
      return {"a": "b"}    
      
  app.run()
```
Does that mean in that example you'd expect to get a string representation of the dict as `text/plain`?

connexion/apis/flask_api.py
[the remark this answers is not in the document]
cognifloyd: Oh. This accomplishes the whole dict => json default. hmmm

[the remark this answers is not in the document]
cognifloyd: So, previous to this change no aiohttp users would be able to directly return a dictionary anyway. They would have had to return a ConnexionResponse or a web.StreamResponse. So, enabling passing a dictionary here should not be backwards incompatible. I think I have a change to add this. I'll push it momentarily.

cognifloyd:
K. So my change is a little more restrictive than flask's _jsonify_data, but hopefully it will be less surprising when people are trying to serialize with something other than json.
If the mimetype is provided and it is not json, then it will only run data through str() even if it is a dict or anything like that.

But, this doesn't automatically change the mimetype as there is no obvious place to "change" a requested mimetype/content_type. I suppose it could be done just for aiohttp in the aiohttp-specific version of get_response, but that feels like a weird place to start introspecting the data to change the content_type. If we do want to do that in get_response instead of in _jsonify_data, then we could rely on `is_json_mimetype` to trigger the jsonification in `_jsonify_data` instead of inspecting the data type there.

But if we start introspecting the data type in get_response, does that mean that every other backend is going to do that as well? Where is flask magically turning this into `application/json`?

connexion/apis/flask_api.py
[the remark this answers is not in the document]
dtkav: oh yeah - you're right it is backwards compatible - nice!

connexion/apis/abstract.py
```diff
@@ -238,23 +242,195 @@ def get_request(self, *args, **kwargs):
...
+    def _jsonify_data(cls, data, mimetype):
+        if not isinstance(data, bytes):
+            # assume mimetype is json if mimetype is None and it looks like json
+            if (mimetype is None and isinstance(data, (dict, list, tuple))) \
+                    or (isinstance(mimetype, str) and is_json_mimetype(mimetype)):
+                body = cls.jsonifier.dumps(data)
+            elif isinstance(data, str):
+                body = data
+            else:
+                body = str(data)
```
dtkav: What are some other datatypes where the string representation is better? I'm having trouble thinking of any.
For both `int` and `float`, I think I'd prefer the json representation.

connexion/apis/flask_api.py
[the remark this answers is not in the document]
dtkav: this is a tricky one! upon more reflection, if your intent is to allow a seamless switch between aiohttp and flask, then I think it makes sense to adopt the flask tuple behavior into abstract. One way to think of this is "if there's no content-type, then default to using json for serialization". I think that's a reasonable default, and it's much more useful and predictable than str(). If folks want to override the default, they can return a ConnexionResponse, framework response, or specify the content-type explicitly. What do you think?

connexion/apis/abstract.py
cognifloyd: ok. Now, when mimetype is None, anything that can be jsonified will be, catching TypeError to rescue with `str()`. Duck-typing for the win - if it quacks like JSON and the mimetype is None, then it is JSON.

connexion/apis/flask_api.py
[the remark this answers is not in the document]
cognifloyd: OK. I went ahead and got rid of `flask_api._jsonify_data()`. See the commit message about why I think this is a combination of new feature (`str()` instead of `TypeError`) and bugfix (do not jsonify data with non-json mimetypes). What do you think? Does that make sense?

ie - use the behavior from flask for the blue square, and use the behavior from aiohttp for the yellow squares:

Selection_033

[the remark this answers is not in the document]
dtkav: parallel conversation going on gitter, but IMO str() is anti-useful. I think it's important for APIs to return structured data by default, and if the data can't be represented in a structured way it's better to error (and require the developer to implement serialization) than call str() and return something unstructured.

connexion/apis/abstract.py
dtkav: I think this is getting really close.
Eventually we want to be able to register serializers for arbitrary mimetypes, so I think there's a default path (mimetype is none, try to do json), and then maybe some built-in overridable serializers (text/plain uses str() for example).
So if we can get backwards compatible behavior with flask (default serializer is json), but an escape hatch registering new mimetypes/serializers then I think we're golden.
One thing I don't think we should do is introduce anything opinionated, or anything too magical/surprising.

cognifloyd:
OK. So, harmonizing the serialization between aiohttp and flask turns out to be ugly.

Several of the flask tests specify a mimetype of `text/plain` and then expect the response to be jsonified. So, I had to do `if mimetype in [None, "text/plain"]`. Validators also allow for validating json in both `application/json` and `text/plain`.

From the tests it looks like you can currently `return data` with aiohttp, but you can't return a tuple. Data may be any type and then it would be serialized in `_cast_body` to some kind of string. So, we can change the aiohttp serialization if something is included in a tuple, but returning a dict with mimetype text/plain (or no mimetype) currently just serializes to a dict literal which is not valid JSON.

I also had to adjust one of those aiohttp tests because now that it is jsonified instead of just stringified, there's a newline at the end.

So, this is cleaner and more consistent, but harmonizing the jsonification might have to be a separate PR that goes in a major version bump. I can imagine this being part of a pluggable serializers feature. It wouldn't be too much more work to have a chain of serializers that vote based on the mediatype.

How far can we harmonize serialization in this PR?

connexion/apis/abstract.py
cognifloyd: It turns out that it is magical and surprising no matter how I look at it. From both aiohttp and flask sides, it feels a little off. :(

cognifloyd:
We could introduce the planned new behavior in `abstract` and then override it with the backwards compatible equivalent in both `aiohttp_api` and `flask_api`.

cognifloyd:
K. I removed the commits that harmonized serialization between flask and aiohttp. I will push a new PR where we can continue that work, but I expect that will have to wait until a major version bump so we can make backwards incompatible adjustments.

dtkav:
ok that makes sense. Happy to tackle it in another commit.
What do you think about adding a deprecation warning for folks using weird defaults?
I'm feeling more and more excited about just fixing all of this internal api mismatch, and pluggable serialization and deserialization on a branch and push for a major version bump.

cognifloyd:
I added some `FutureWarning`s about deprecation where behavior will need to change to harmonize serialization.
> warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

I picked `FutureWarning` because it seemed to fit better than `DeprecationWarning`
> warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).

cognifloyd:
@dtkav @hjacobs @jmcs I think is ready to merge.

hjacobs:
:+1:

PR #851, opened 2019-01-14
Datetime serialization

### LINKED ISSUE #846: datetime support for JSonifier
### Description
Connexion using AioHttp raise a 500 error when accessing specifications with date time example.
Note: when using ujson, datetime is cast to int without error, but it still invalid.

### Expected behaviour
Specifications are returned without error and properly serialized date time.

### Actual behaviour
500 error with json, int with ujson

### Steps to reproduce
openapi.yaml
```
openapi: "3.0.1"

info:
  title: Hello World
  version: "1.0"
servers:
  - url: http://localhost:8080/api

paths:
  /hello:
    get:
      summary: Generate data with date time
      operationId: hello.hello
      responses:
        200:
          description: greeting response
          content:
            application/json:
              schema:
                type: object
                properties:
                  date:
                    type: string
                    format: date-time
                example:
                  date: 2000-01-23T04:56:07.000+00:00
```

main.py
```
from aiohttp import web
import connexion

def hello() -> str:
    return web.json_response({"value": "2000-01-23T04:56:07.000+00:00")


if __name__ == '__main__':
    app = connexion.AioHttpApp(__name__, port=8080, specification_dir='openapi/')
    app.add_api('openapi.yaml', arguments={'title': 'Hello World Example'})
    app.run()
```

Making error with curl
```
curl localhost:8080/api/openapi.json
<html><head><title>500 Internal Server Error</title></head><body><h1>500 Internal Server Error</h1>Server got itself in trouble</body></html>



### Additional info:

Output of the commands:

- `python --version`
Python 3.6.7
Version: 2018.0.dev1

### PR DESCRIPTION
**Pull request by Jyhess**
Fixes #846 .

 - Add date, datetime and uuid serialization for AioHttp
 - Use JSonifier to serialize all data in AioHttp
 - Move Jsonifier in its own file for easier extension
 - Remove ujson dependency: Flask use json from itsdangerous, [ujson does not support custom types](#124) for AioHttp.

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -17,7 +17,7 @@ from connexion.handlers import AuthErrorHandler
 from connexion.jsonifier import JSONEncoder, Jsonifier
 from connexion.lifecycle import ConnexionRequest, ConnexionResponse
 from connexion.problem import problem
-from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
+from connexion.utils import is_json_mimetype, yamldumper
 from werkzeug.exceptions import HTTPException as werkzeug_HTTPException
 
 try:
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -11,16 +11,9 @@ from aiohttp.web_middlewares import normalize_path_middleware
 from connexion.apis.abstract import AbstractAPI
 from connexion.exceptions import OAuthProblem, OAuthScopeProblem
 from connexion.handlers import AuthErrorHandler
+from connexion.jsonifier import JSONEncoder, Jsonifier
 from connexion.lifecycle import ConnexionRequest, ConnexionResponse
-from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
-
-try:
-    import ujson as json
-    from functools import partial
-    json.dumps = partial(json.dumps, escape_forward_slashes=True)
-
-except ImportError:  # pragma: no cover
-    import json
+from connexion.utils import is_json_mimetype, yamldumper
 
 logger = logging.getLogger('connexion.apis.aiohttp_api')
 
@@ -33,7 +26,7 @@ def oauth_problem_middleware(request, handler):
     except (OAuthProblem, OAuthScopeProblem) as oauth_error:
         return web.Response(
             status=oauth_error.code,
-            body=json.dumps(oauth_error.description).encode(),
+            body=AioHttpApi.jsonifier.dumps(oauth_error.description).encode(),
             content_type='application/problem+json'
         )
     return response
@@ -302,7 +295,7 @@ class AioHttpApi(AbstractAPI):
     def _cast_body(cls, body, content_type=None):
         if not isinstance(body, bytes):
             if content_type and is_json_mimetype(content_type):
-                return json.dumps(body).encode()
+                return cls.jsonifier.dumps(body).encode()
 
             elif isinstance(body, str):
                 return body.encode()
@@ -314,7 +307,7 @@ class AioHttpApi(AbstractAPI):
 
     @classmethod
     def _set_jsonifier(cls):
-        cls.jsonifier = Jsonifier(json)
+        cls.jsonifier = Jsonifier(cls=JSONEncoder)
 
 
 class _HttpNotFoundError(HTTPNotFound):
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -20,13 +20,6 @@ from connexion.problem import problem
 from connexion.utils import is_json_mimetype, yamldumper
 from werkzeug.exceptions import HTTPException as werkzeug_HTTPException
 
-try:
-    import ujson as json
-    from functools import partial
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: connexion/apps/aiohttp_app.py [hunk 1]
```diff
@@ -20,7 +20,7 @@ def __init__(self, import_name, only_one_api=False, **kwargs):
         self._api_added = False
 
     def create_app(self):
-        return web.Application(debug=self.debug)
```
dtkav: hey @Jyhess , this doesn't seem related to json serialization.
Can you factor this out into a separate PR.

dtkav:
This is looking great. I had one minor comment about some unrelated changes in your diff, but otherwise nice!

connexion/jsonifier.py
```diff
@@ -0,0 +1,59 @@
+import datetime
+import json
+import uuid
+
+import six
+
+
+class JSONEncoder(json.JSONEncoder):
+    def default(self, o):
```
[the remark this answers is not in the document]
dtkav: If i'm reading the `codacy` output correctly, I think you might need to rename this method, and then set it as an attribute named `default`.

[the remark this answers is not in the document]
Jyhess: Is there a possibility to mark this a false positive for Codacity ?
Because it's the standard way to extend JSONEncoder. If we hack our code for Codacity, it can confuse other developers which would like to extend our own JSONEncoder.

connexion/apps/aiohttp_app.py
Jyhess: I will do it

connexion/jsonifier.py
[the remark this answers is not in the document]
dtkav: You are right. There's an example from the python docs that does it exactly as you did: https://docs.python.org/2/library/json.html

I think it's fine to ignore it then.

jmcs:
@Jyhess the changes to master broke your PR, sorry. Can you fix the conflicts. Thanks.

dtkav:
The problem is coming from the new version of Flake8. I'm trying to come up with a fix. (#867 #866 )

Jyhess:

hjacobs:

cognifloyd:
@Jyhess I'm looking forward to getting this and your other PRs merged. I figure this is the closest to getting in, so could you rebase this one first?

connexion/apis/aiohttp_api.py
```diff
@@ -33,7 +26,7 @@ def oauth_problem_middleware(request, handler):
     except (OAuthProblem, OAuthScopeProblem) as oauth_error:
         return web.Response(
             status=oauth_error.code,
-            body=json.dumps(oauth_error.description).encode(),
+            body=AioHttpApi.jsonifier.dumps(oauth_error.description).encode(),
```
cognifloyd: oauth_problem_middleware has been replaced in master. Now all of the Exception handling uses `connexion.problem.problem()` to generate the response. `problem()` returns a ConnexionResponse. That then goes through `AioHttpApi.get_response()` -> `AioHttpApi._get_aiohttp_response_from_connexion()` -> `AioHttpApi._cast_body()` where it gets converted to JSON if needed.

So, this change is no longer needed as it is covered by the changes below in this file.

dtkav:
I tried to resolve resolve the conflicts against master.
Sorry if the merge messes up your workflow.

dtkav:
Awesome work! My only nit would be to move the unrelated aiohttp example to another PR.

dtkav:
@hjacobs I've fixed the conflicts against master for this datetime serialization branch - It looks ready to me now!

connexion/apis/aiohttp_api.py
```diff
@@ -16,7 +16,17 @@
...
+from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
+from werkzeug.exceptions import HTTPException as werkzeug_HTTPException
+
+try:
+    import ujson as json
+    from functools import partial
+    json.dumps = partial(json.dumps, escape_forward_slashes=True)
+
+except ImportError:  # pragma: no cover
+    import json
```
cognifloyd: This `import ujson` / `import json` block should be removed.

With this PR, `json.dumps` calls are centralized in the Jsonifier so that datetimes can be handled consistently.
As such, this file actually does not use this json import anymore, so it should be removed.

dtkav: thanks @cognifloyd - great catch

dtkav:
@hjacobs can you please take a look at this feature (datetime serialization). It look ready to go.

connexion/apis/aiohttp_api.py
cognifloyd: This PR removed this import block.
The merge with master re-introduced it.
Please remove this import block again.

dtkav: My bad. Does it look right to you now?

I was forced to use the editor built into github to resolve the merge (since it isn't my PR), so I'm missing a bunch context I would normally have in my local environment.

cognifloyd: That resolves my only concern :) Thanks

hjacobs:
:+1:

PR #869, opened 2019-02-05
Async security check

### LINKED ISSUE #868: Security handlers should be async when using AioHttp
### Description
In security functions, I need to call other URL to verify token validity.
With AioHttp, this must be done in coroutine to not block main thread.
But if try to use a coroutine with `x-tokenInfoFunc` (or others), I have following error:
```
  [...]
  File "/home/sagnard/dev/connexion/connexion/decorators/security.py", line 298, in wrapper
    request.context['user'] = token_info.get('sub', token_info.get('uid'))
AttributeError: 'generator' object has no attribute 'get'
```
More than that, we should have access to aiohttp request object in those function, when it is defined with `pass_context_arg_name=request`.

### Expected behaviour
I can use coroutine for security handlers, and it can accept a request parameter.

### Actual behaviour
There is an error if I try.

### Steps to reproduce
Update tests/fakeapi/auth.py to add `@asyncio.coroutine` decorator to `fake_basic_auth`, and execute aiohttp tests.

### Additional info:

Output of the commands:

- `python --version` 3.6.7
- `pip show connexion | grep "^Version\:"` master

Jyhess:
Fixes #868 .

Allow to use coroutines with Aiohttp in security handlers.

 - Move all function from `connexion.operations.security` to a class in `connexion.security.SecurityHandlerFactory`
 - Specialize this class for AioHttp and Flask, and instantiate them in dedicated apis
 - AioHttp specialization now support coroutines
 - `pass_context_arg_name` option is supported for security handlers


Jyhess:

elsnoon:
Hi guys, when we can expect this to be merged? 

It would be great to have async security check. At the moment we have some workaround solution with middleware.

defol:
Is there any progress on this?

hjacobs:
Sorry for the delay. Could you rebase the PR and fix the conflicts? We will take a look afterwards.

cognifloyd:
I will rebase this one once #849 is merged. (Thanks @Jyhess for giving me access so I can move the PRs forward)

connexion/exceptions.py
```diff
@@ -83,8 +83,7 @@ class OAuthProblem(Unauthorized):
 
 
 class OAuthResponseProblem(OAuthProblem):
-    def __init__(self, token_response, **kwargs):
-        self.token_response = token_response
```
dtkav: what are the implications of removing this?

connexion/security/aiohttp_security_handlers_factory.py
```diff
@@ -0,0 +1,111 @@
...
+
+import aiohttp
+
+from ..exceptions import OAuthProblem, OAuthResponseProblem, OAuthScopeProblem
+from .security_handler_factory import SecurityHandlerFactory
+
+logger = logging.getLogger('connexion.api.security')
+
+
+class AioHttpSecurityHandlerFactory(SecurityHandlerFactory):
```
dtkav: are we doing to have to write one of these for every framework we support?
Is there a more generic version of this like `AsyncSecurityHandlerFactory`?

connexion/security/aiohttp_security_handlers_factory.py
```diff
@@ -0,0 +1,111 @@
...
+        :type token_info_url: str
+        :rtype: types.FunctionType
+        """
+        @asyncio.coroutine
+        def wrapper(token):
+            if not self.client_session:
+                # Must be created in a coroutine
+                self.client_session = aiohttp.ClientSession()
+            headers = {'Authorization': 'Bearer {}'.format(token)}
+            token_request = yield from self.client_session.get(token_info_url, headers=headers, timeout=5)
```
dtkav: 5 is a magic number here - it might make sense to add as a field on the base class in case someone wants to change it.

connexion/exceptions.py
cognifloyd: It looks like nothing uses it. Everywhere that `OAuthResponseProblem` is raised, `token_response` is set to `None`.

If we remove it, we should also remove it from `docs/exceptions.rst`:

> This exception is thrown when there is a validation issue from your OAuth 2 Server. It contains a
> ``token_response`` property which contains the full http response from the OAuth 2 Server

If we decide to retain it, then we'll probably need to actually include the response here. But this is actually raised in places that have nothing to do with an oauth response... That's fishy. :(

`get_tokeninfo_func` calls `get_token_info_remote` to actually request an oauth token, so I would expect this to only be raised in that context, but those do not raise anything. I think we need a new exception named appropriately for how it is actually used. Any recommendations on what to name it?

connexion/security/aiohttp_security_handlers_factory.py
```diff
@@ -0,0 +1,111 @@
...
+                kwargs[self.pass_context_arg_name] = request.context
+            if need_to_add_required_scopes:
+                kwargs[self.required_scopes_kw] = required_scopes
+            token_info = func(*args, **kwargs)
+            while asyncio.iscoroutine(token_info):
+                token_info = yield from token_info
+            if token_info is self.no_value:
+                return self.no_value
+            if token_info is None:
+                raise OAuthResponseProblem(description=exception_msg)
```
cognifloyd: There is no oauth server response here, and on further inspection this function is not just for oauth so raising `OAuthResponseProblem` is actually not correct. This is doing the same as the old security.py version, but that still does not make sense.

connexion/security/security_handler_factory.py
```diff
@@ -0,0 +1,328 @@
...
+                kwargs[self.pass_context_arg_name] = request.context
+            if need_to_add_required_scopes:
+                # required_scopes is passed in kw because python 2.7 does not support syntax:
+                # def wrapper(request, *args, required_scopes=None)
+                kwargs[self.required_scopes_kw] = kw.get(self.required_scopes_kw)
+            token_info = func(*args, **kwargs)
+            if token_info is self.no_value:
+                return self.no_value
+            if token_info is None:
+                raise OAuthResponseProblem(description=exception_msg)
```
cognifloyd: This is odd. Here we're using `OAuthResponseProblem` for bearer token, basic auth and api key, not just for oauth. look for all of the `_generic_check` usage to see what I'm talking about.

cognifloyd:
K. I'm not entirely confident about my fixes for generic auth issues.

connexion/exceptions.py
dtkav: Some history here - every security handler used to raise OAuthResponseProblem in case of an error, so including it (or inheriting from it) was for backwards compatibility.
Hopefully we can drop that requirement at some point.

cognifloyd: Oh. Well, I just got rid of it everywhere where it was in use, and added it where it made sense.

connexion/security/aiohttp_security_handlers_factory.py
cognifloyd: done

connexion/security/aiohttp_security_handlers_factory.py
cognifloyd: done

cognifloyd:
OK. That took some mind bending to get this one rebased with tests passing.
I addressed all of the things I and @dtkav have found so far. Please help me with reviews to see if I missed anything that needs to be adjusted.

Now the questions:
- Can we make these changes to use UnauthorizedProblem instead of OAuth*Problem?
- Is `auth_remote_token_timeout` a good name for the new option that exposes changing that timeout? I've got that in here as well.

So much has shifted around in here, I'm not sure how to think about backwards compatibility. I think the only safe way to make this much change is to save it for the next major version bump (either v3 with #1096 or whatever counts for a major CalVer release in #1100).

connexion/exceptions.py
cognifloyd: If we need to adjust the PR to split that out into another PR then fine. This PR wore me out, so if changes need to occur I'll wait a bit to make them.

chbndrhnns:
Hey, is there any progress planned on that PR?

cognifloyd:
Yes. I hope to get to this soonish. I've been blocked with several other work projects.

First, I need to rebase this on master.
Then, I plan on splitting the `UnauthorizedProblem` changes/fixes into a separate PR. So, even though the current interface doesn't make sense, this PR shouldn't change that. This will go back to `OAuthUnauthorizedProblem` with `token_response=None`.

After that, I need reviews to see if I missed anything.

cognifloyd:
K. I just rebased this on master, removing the `OAuthResponseProblem` changes. Those exception changes/cleanups are in my [unauthorized_problem](https://github.com/cognifloyd/connexion/tree/unauthorized_problem) branch.

Next steps:
- [ ] satisfy travis
- [ ] adjust to use async/await instead of coroutine decorator

cognifloyd:
@dtkav could you review when you get a chance?

cognifloyd:
@hjacobs this is ready for review

chbndrhnns:
@jmcs I saw you did some work recently on this project. This PR seems to be ready for quite some time. Any chance to have a look and approve/review?

I am in a position where I would need such functionality for performance reasons and I would prefer the changes in the official repo instead of forking my own version of connexion just to get async security handlers.

jmcs:
:+1:

PR #830, opened 2018-12-20
AioHttpApi oauth middleware should catch all OAuth problems

lynxknight:
Fixes #829

PR #876, opened 2019-02-15
OAuthResponseProblem -> 401 in AioHttpApp

### LINKED ISSUE #875: oauth_problem_middleware not catching OauthResponseProblem in aiohttpapp
### Description

AioHttpApp is shipped with a dedicated [middleware for wrapping oauth errors](https://github.com/zalando/connexion/blob/master/connexion/apis/aiohttp_api.py#L33) in order not to send default 500 responses.
Problem is, OAuthResponseProblem seems not to be included and is then raising an error 500 when the api key is not valid.
From my perspective, adding OAuthResponseProblem as an catched exception should solve this problem.

### Expected behaviour

Providing an api key that is not valid should raise an Unauthorized 401 error.

### Actual behaviour

Error code 500 without custom description is provided

### Steps to reproduce

Using api key authentication managed by a custom x-apikeyInfoFunc, any request without a valid api key will [raise an OauthResponseProblem](https://github.com/zalando/connexion/blob/master/connexion/decorators/security.py#L270) here and won't be transformed in _aiohttp code terminology_. 

### Additional info:

python 3.7.1
connexion 2.2.0
aiohttp 3.5.4

arastel:
Fixes #875  .

 - invalid api key is now described as 401 with appropriate description instead of generic error 500

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -9,7 +9,7 @@ import aiohttp_jinja2
 from aiohttp import web
 from aiohttp.web_exceptions import HTTPNotFound
 from connexion.apis.abstract import AbstractAPI
-from connexion.exceptions import OAuthProblem, OAuthScopeProblem
+from connexion.exceptions import OAuthProblem, OAuthScopeProblem, OAuthResponseProblem
 from connexion.handlers import AuthErrorHandler
 from connexion.lifecycle import ConnexionRequest, ConnexionResponse
 from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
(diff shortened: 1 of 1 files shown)


connexion/apis/aiohttp_api.py
```diff
@@ -30,7 +30,7 @@
 def oauth_problem_middleware(request, handler):
     try:
         response = yield from handler(request)
-    except (OAuthProblem, OAuthScopeProblem) as oauth_error:
+    except (OAuthProblem, OAuthScopeProblem, OAuthResponseProblem) as oauth_error:
```
dtkav: `OAuthResponseProblem` is a subclass of `OAuthProblem`.. I'm confused why this is needed.

arastel: Sorry, I forgot to close it after seeing that [PR 830](https://github.com/zalando/connexion/pull/830 ) was also solving that.

PR #877, opened 2019-02-15
Flask API: allow setting content type in response headers

### LINKED ISSUE #266: Validation issue with validate_responses=True and connexion.problem(...)
When using connexion.problem(...) to return an error response AND at the same time enabling validate_responses in the API, I get an exception.

With the example in attachment, when going to 
http://127.0.0.1:5009/api/dogs
I get an exception TypeError: <Response 109 bytes [403 FORBIDDEN]> is not JSON serializable.

If I set validate_response=False, it works as expected.
If I remove the responses/403 from the test.yaml, it works (ie no exception) but I would expect it to fail as the response is not valid.

[debug_connexion.zip](https://github.com/zalando/connexion/files/421981/debug_connexion.zip)

sandrinr:
Partially fixes #266 . If this PR is accepted then one way to go forward would be to change `connexion.problem.problem()` to return a `tuple` response instead of a `ConnexionResponse`.

 - Allow for changing the content type of a response via the header element when using a tuple response from a controller method.

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -165,6 +165,9 @@ class FlaskApi(AbstractAPI):
     @classmethod
     def _build_flask_response(cls, mimetype=None, content_type=None,
                               headers=None, status_code=None, data=None):
+        # use content type specified in headers, if set
+        content_type = headers.get("Content-Type", content_type) if headers else None
+        mimetype = content_type.rsplit(";", 1)[0] if content_type else mimetype
         kwargs = {
             'mimetype': mimetype,
             'content_type': content_type,


connexion/apis/flask_api.py
```diff
@@ -165,6 +165,9 @@ def _get_flask_response_from_connexion(cls, response, mimetype):
     @classmethod
     def _build_flask_response(cls, mimetype=None, content_type=None,
                               headers=None, status_code=None, data=None):
+        # use content type specified in headers, if set
+        content_type = headers.get("Content-Type", content_type) if headers else None
```
dtkav: I would expect that passing `content_type` explicitly would take precedent over headers.
why not:
```
if content_type is None:
    content_type = headers.get("content-type")
```

sandrinr: I think the headers should take precedence because it is the only way to pass such values in the tuple based controller return API. Also, the `content_type` parameter is defaulting to `mimetype` in some paths and this `mimetype` is set according to the `produces` spec parameter.

However, the `... else None` part should be `... else content_type`.

hjacobs:
I'm closing this PR for now (as it's pretty old, there are no recent updates, and no tests).

PR #881, opened 2019-02-18
fix strict_validation debug log variable

sherzberg:
The log message for `strict_validation` uses the incorrect variable in the log format. This PR fixes that.


jmcs:
👍

PR #885, opened 2019-02-24
Forward options to the AioHTTP app.

marcofavorito:

 - The `**options` arguments the [`AioHttpApp.run`](https://github.com/zalando/connexion/blob/master/connexion/apps/aiohttp_app.py#L71) method are not forwarded to the [`run_app`](https://github.com/zalando/connexion/blob/master/connexion/apps/aiohttp_app.py#L93) method of the `aiohttp` framework. It is analogous to what is done here: https://github.com/zalando/connexion/blob/master/connexion/apps/flask_app.py#L93, but for the `flask` framework.

### DIFF
--- a/connexion/apps/aiohttp_app.py
+++ b/connexion/apps/aiohttp_app.py
@@ -87,7 +87,7 @@ class AioHttpApp(AbstractApp):
 
             access_log = options.get('access_log')
 
-            if options.get('use_default_access_log'):
+            if options.pop('use_default_access_log'):
                 access_log = logger
 
             web.run_app(self.app, port=self.port, host=self.host, access_log=access_log, **options)
--- a/connexion/apps/aiohttp_app.py
+++ b/connexion/apps/aiohttp_app.py
@@ -87,7 +87,7 @@ class AioHttpApp(AbstractApp):
 
             access_log = options.get('access_log')
 
-            if options.pop('use_default_access_log'):
+            if options.pop('use_default_access_log', None):
                 access_log = logger
 
             web.run_app(self.app, port=self.port, host=self.host, access_log=access_log, **options)
--- a/connexion/apps/aiohttp_app.py
+++ b/connexion/apps/aiohttp_app.py
@@ -90,6 +90,6 @@ class AioHttpApp(AbstractApp):
             if options.get('use_default_access_log'):
                 access_log = logger
 
-            web.run_app(self.app, port=self.port, host=self.host, access_log=access_log)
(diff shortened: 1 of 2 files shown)


connexion/apps/aiohttp_app.py
```diff
@@ -90,6 +90,6 @@ def run(self, port=None, server=None, debug=None, host=None, **options):
             if options.get('use_default_access_log'):
```
dtkav: I'm not sure what the rationale was to manage these explicitly, but it seems like we should do one, or the other, not both.

cognifloyd: git history shows that this was part of the very first commit that added aiohttp: 5ce51450fe21927673727d1284a110342f499075

So, nothing enlightening to add there. It is a little odd...

tests/aiohttp/test_aiohttp_app.py
```diff
@@ -26,7 +26,7 @@ def test_app_run(web_run_app_mock, aiohttp_api_spec_dir):
     app.run(use_default_access_log=True)
     logger = logging.getLogger('connexion.aiohttp_app')
     assert web_run_app_mock.call_args_list == [
-        mock.call(app.app, port=5001, host='0.0.0.0', access_log=logger)
+        mock.call(app.app, port=5001, host='0.0.0.0', access_log=logger, use_default_access_log=True)
```
cognifloyd: Why is this change needed here? Isn't the **options sufficient in aiohttp_app? or do tests break without this?

connexion/apps/aiohttp_app.py
cognifloyd:
```suggestion
            if options.pop('use_default_access_log'):
```
This is a connexion-specific option, so we should pop it here and not pass it on the aiohttp.

tests/aiohttp/test_aiohttp_app.py
cognifloyd: I see the error. See my other comment for how I would prefer to make the tests pass.

marcofavorito:
@cognifloyd thanks for your comments. I updated my PR as discussed.

Do you have a list of contributors? Can I add my name? What is your policy on that?

cognifloyd:
I do not see a contributors doc. :) Now for someone at Zalando to hit the merge button...

@hjacobs? @jmcs? This one is a simple but helpful fix. Please merge.

hjacobs:
:+1:

PR #894, opened 2019-03-05
fixes basic auth example doc; adds apikey example

sherzberg:
Fixes a documentation link issue. The basic auth example link was under the api key example docs. This PR also implements an example for apiKey security scheme.

One question though (there are TODOs in the code I would like to update), the Connexion docs state that the `x-apikeyInfoFunc` method takes two parameters, `apikey` and `required scopes`. In testing, the required scopes are always None, and after looking at the openapi spec (https://swagger.io/docs/specification/authentication/#scopes) it appears that scopes are not used for the api key scheme.

Should the `x-apikeyInfoFunc` method not take in a `required_scopes` arg?

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -58,6 +58,8 @@ taken from there, otherwise it's None. This allows authorizing individual
 operations with oauth scope while using basic authentication for
 authentication.
 
+You can find a `minimal Basic Auth example application`_ in Connexion's "examples" folder.
+
 ApiKey Authentication
 ---------------------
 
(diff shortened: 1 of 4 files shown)


sherzberg:
I see in the Connexion code where the required scopes are gathered (https://github.com/zalando/connexion/blob/2.2.0/connexion/operations/secure.py#L93) and I assume this is to match the openapi spec (so only gather the required scopes if the scheme is oauth2). I will remove the TODOs.

dtkav:
Thanks for adding this example!

jmcs:
:+1:

PR #897, opened 2019-03-06
Upgrade insecure dependency PyYAML to 4.2b2+

iamyohann:

 - Bump PyYAML to 4.2b2+ (resolves CVE-2017-18342 https://nvd.nist.gov/vuln/detail/CVE-2017-18342)


dtkav:
Hey @iamyohann ,

Have a read through this thread: https://github.com/p1c2u/openapi-spec-validator/issues/60

TL;DR - We can't pin a beta release, as it will break dependency managers like pipenv. However, this CVE only relates to the use of yaml.load(), which we do not use (we use safe_load, or extend the safe loader).

iamyohann:
@dtkav I see, can I leave this PR open until a stable version of PyYAML is released?

dtkav:
Yeah, absolutely.

ThoreKr:
So with 5.1 it should be fine unless there are also side effects as in yaml/pyyaml#265?

dtkav:
superseded by #902

PR #900, opened 2019-03-10
Remove unused metavar

bbelyeu:

 - It appears the mockmode metavar is unused in code. Please correct me if I am wrong.
 - The problem it causes by being there is it overrides the default --help` string from `connexion run --help` so that it is difficult to understand what the possible arguments are for `--mock`. From reading the `click` docs & code, it seems that if we leave the metavar out, it will correctly populate the available options.

### DIFF
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -73,7 +73,7 @@ def main():
               help='Returns status code 501, and `Not Implemented Yet` payload, for '
               'the endpoints which handlers are not found.',
               is_flag=True, default=False)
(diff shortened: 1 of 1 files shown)


dtkav:
This metavar is not "unused" per se, but I think the help text is clearer without it.

The difference is shown below:
`connexion run --help`

with:
```
  --mock MOCKMODE                 Returns example data for all endpoints or
                                  for which handlers are not found.
```
without:
```
  --mock [all|notimplemented]     Returns example data for all endpoints or
                                  for which handlers are not found.
```

dtkav:
( ^ kicking Travis )

hjacobs:
:+1:

PR #903, opened 2019-03-15
bump and pin isort, apply all automatic fixes

dtkav:
Fixes the build.

 - apply all of the isort changes that resulted from the latest version
 - pin to the latest version (4.3.15) so this doesn't happen again, unless we bump the version on purpose


jmcs:
:+1:

PR #912, opened 2019-03-29
Fix return type docstring for get_response

terrycorley:

 - Fix return type docstring in `connexion.apis.flask_api.FlaskApi.get_response` so editors/IDEs don't raise warnings when type hinting


dtkav:
( ^^ kicking travis)

hjacobs:
:+1:

Issue #930, opened 2019-04-25
Use jsonschema for validation


patrickkwang:
A JSON version of the OpenAPI 3.0 specification is now available:
https://github.com/OAI/OpenAPI-Specification/commit/6d17b631fff35186c495b9e7d340222e19d60a71
One can validate against this using jsonschema instead of using openapi_spec_validator, here:
https://github.com/zalando/connexion/blob/master/connexion/spec.py#L227
In my experience, both work but jsonschema produces more useful error messages.

rafaelcaricio:
Hi @patrickkwang,

This seems like an interesting idea since we would possibly be able to remove one dependency. Would you be interested in baking a PR with this change?

PR #936, opened 2019-05-07
jsonschema validation

patrickkwang:
Addresses #930.

 - use jsonschema for validation instead of openapi_spec_validator

### DIFF
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -15,17 +15,11 @@ import pkg_resources
 import yaml
 from jsonschema import Draft4Validator
 from jsonschema.validators import extend as extend_validator
-from openapi_spec_validator.exceptions import OpenAPIValidationError
 
 from .exceptions import InvalidSpecification
 from .json_schema import resolve_refs
 from .operations import OpenAPIOperation, Swagger2Operation
 from .utils import deep_get
-
-try:
-    import collections.abc as collections_abc  # python 3.3+
-except ImportError:
-    import collections as collections_abc
 
 validate_properties = Draft4Validator.VALIDATORS["properties"]
 
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -12,6 +12,8 @@ import pkg_resources
 
 import jinja2
 import jsonschema
+from jsonschema import Draft4Validator
+from jsonschema.validators import extend as extend_validator
 import yaml
 from openapi_spec_validator.exceptions import OpenAPIValidationError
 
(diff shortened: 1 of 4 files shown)


connexion/spec.py
```diff
@@ -156,6 +158,9 @@ class Swagger2Specification(Specification):
     yaml_name = 'swagger.yaml'
     operation_cls = Swagger2Operation
 
+    response = requests.get('https://raw.githubusercontent.com/OAI/OpenAPI-Specification/6d17b631fff35186c495b9e7d340222e19d60a71/schemas/v2.0/schema.json')
```
bobh66: This will be a problem in deployments that do not have access to the public internet.  We deploy in locations that are isolated from the outside world, and the application will fail to start if this change is merged.  I'd suggest including the specs in the PR so they are packaged and delivered along with connexion.  That eliminates the need for network access to retrieve them at run time.

rafaelcaricio: This cannot be like that. We cannot rely on internet connection.

patrickkwang:
Many of the tests seem to fail because the OpenAPI specs used for testing are actually invalid.
For example: https://travis-ci.org/zalando/connexion/jobs/530330378#L787
If "in: path", then you must have "required: true":
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterRequired
Perhaps openapi-spec-validator was not handling this correctly.

dtkav:
Hey @patrickkwang - I'm excited for these changes - thanks for taking them on!

hjacobs:
Could you make the tests pass?

patrickkwang:
No, the tests will fail because many of them are based on the invalid `simple` schema: https://github.com/zalando/connexion/pull/936#issuecomment-490949868

The tests are corrected in #1010, so this PR will be dependent on that one.

dtkav:
It looks like the jsonschema version doesn't check the case where a parameter default is the wrong type, and this results in two failing tests.
This is something that we can check at runtime, but it is somewhat unfortunate. Overall, it's probably worth the simplicity of this change, but we'll have to remove those tests.
I'll see if folks upstream have any thoughts on this.

patrickkwang:
The JSON Schema actually doesn't enforce that default values validate against the associated schema:
https://json-schema.org/understanding-json-schema/reference/generic.html#annotations
The OpenAPI specification is stricter:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#schema-object
So jsonschema is doing what it ought to. Someone could fork it to enforce that extra OpenAPI requirement.

dtkav:
Here's the relevant section from the spec.
> Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if type is string, then default can be "foo" but cannot be 1.

This PR uses jsonschema to validate the openapi spec.
I think the jsonschema spec provided fails to validate that defaults are correct type properly.

I think this is a bug in their openapi spec validation as implemented with jsonschema.

(the language here is very confusing - ack!)

dtkav:
Also of note, this doesn't completely alleviate the need for `openapi-spec-validator` yet. We are still using it for remote '$ref' resolution for yaml files.

patrickkwang:
It turns out it's not too hard to extend jsonschema to enforce that all the default values validate against the associated (sub)schemas. The last two tests pass now! This `instance_validator` business feels janky, though; I intend to ask @Julian if there's a cleaner way to handle references when validating the defaults.

dtkav:
nice! I tried to ask in the swagger irc if there's a way to do it in jsonschema, but didn't hear back.

connexion/spec.py
```diff
@@ -207,17 +244,21 @@ def base_path(self, base_path):
 
     @classmethod
     def _validate_spec(cls, spec):
```
RobbeSneyders, line 246: This method is now the same in both classes, so we should be able to lift this to the base `Specification` class.

RobbeSneyders (commented):
Hi @patrickkwang, I hope you're still interested in getting this merged!
The changes look good, I left one comment.
We also switched to the `main` branch, so would be great if you can retarget and rebase / merge.

RobbeSneyders (approved):
Thanks! LGTM.

connexion/spec.py
```diff
@@ -164,6 +204,9 @@ class Swagger2Specification(Specification):
     yaml_name = 'swagger.yaml'
     operation_cls = Swagger2Operation
 
+    schema_string = pkg_resources.resource_string('connexion', 'resources/schemas/v2.0/schema.json')
+    openapi_schema = json.loads(schema_string.decode('utf-8'))
```
vmarkovtsev, line 208: How about `del schema_string` after loading `openapi_schema`, we waste quite some memory here...

PR #947, opened 2019-05-16
Mention pythonic_params in request docs

### LINKED ISSUE #929: Allow setting a different name for a URL parameter and its corresponding python variable
Suppose I want to implement an endpoint `/data?from=12&until=52` which accepts parameters `from` and `until` (specifying timestamps or something). The problem is that I can't do this because Connexion always uses the name of the query parameter as the name for the Python variable. It's a shame to couple these things together because there are restrictions on Python variable names that don't exist for query parameters - I can't have a Python variable called `from`.

What I would suggest is a solution like this:

```
/data:
    get:
      operationId: "controller.get_data"
      parameters:
      - name: from
        in: query
        x-parameter-name: start_time
        schema:
          type: integer
```

christiansiegel:
Fixes #929

There was no documentation about preventing naming collisions between parameter names and python built-ins with the `pythonic_params` option.

### DIFF
--- a/docs/request.rst
+++ b/docs/request.rst
@@ -54,6 +54,38 @@ of the endpoint parameter `message` to your view function.
 
 Connexion will also use default values if they are provided.
 
+If you want to use a parameter name that collides with a Python built-in,
+you can enable the `pythonic_params` option:
+
+.. code-block:: python
+    app = connexion.FlaskApp(__name__)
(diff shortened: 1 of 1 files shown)


kornicameister:
ping @hjacobs 

this looks as something useful that can be added to docs.

hjacobs:
Thanks!

hjacobs:
:+1:

PR #948, opened 2019-05-17
Add swagger_ui_config option to pass config to the Swagger UI

### LINKED ISSUE #841: Default to DISPLAY_OPERATION_ID=false?
### Description

I'm using connexion to expose the API specification to our customers. I've noticed that in the path, it exposes the codepath (operationId) associated with each route. We'd like to hide this. I've found that Swagger UI configures this using parameter displayOperationId, but I'm not sure how to set it within connexion?

### Expected behaviour

Display of operation Id configurable upon startup of the app.

### Actual behaviour

Operation ID shown.

I may just be overlooking the obvious here. I've tried setting "DISPLAY_OPERATION_ID" as an environment variable but that didn't seem to do the trick. 
https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/

Thanks for your reply!

christiansiegel:
Fixes #841 by add an option `swagger_ui_config` to pass config parameters to the Swagger UI . 

This is done by providing a `swagger-ui-config.json` endpoint with the contents of the `swagger_ui_config` option. This file is then referenced via the `configUrl` parameter in the Swagger UI `index.j2` template.

 - Add `swagger_ui_config` option  to pass config parameters to the Swagger UI

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -139,6 +139,13 @@ class AioHttpApi(AbstractAPI):
                 self._get_swagger_ui_home
             )
 
+        if self.options.openapi_console_ui_config is not None:
+            self.subapp.router.add_route(
+                'GET',
+                console_ui_path + '/swagger-ui-config.json',
+                self._get_swagger_ui_config
+            )
+
         # we have to add an explicit redirect instead of relying on the
         # normalize_path_middleware because we also serve static files
         # from this dir (below)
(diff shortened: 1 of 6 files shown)


dtkav:
This is a great idea! Nice work.
Can you please add a few tests that try to hit the new endpoint based on config existing / not existing?

dtkav:
@hjacobs can you take a look at this one when you get a chance? It will help clean up the interface with the swagger-ui-bundle package.

hjacobs:
:+1:

PR #951, opened 2019-05-22
Fix aiohttp upgrade

cognifloyd:
Fix test_run_with_aiohttp_not_installed with aiohttp 3.5.2+.

The error was prevented by requiring `aiohttp<3.5.2` in #844.
Somewhere between 3.5.1 and 3.5.2, this aiohttp reimport started succeeding.
I used pycharm's debugger to determine that it is this import that was successfully re-importing aiohttp.
https://github.com/zalando/connexion/blob/ff3adb393c67dea51a735ef3211794084db3623e/connexion/cli.py#L151-L155

It's not clear which change caused the issue, but I would guess one of these:

- aio-libs/aiohttp#3469 (Remove wildcard imports)
- aio-libs/aiohttp#3464 (Don't suppress gunicorn cleanup errors)
- aio-libs/aiohttp#3471 (Refactor workers)
- aio-libs/aiohttp#3500 (Ignore done tasks)

 - Revert #844 
 - Also set `sys.modules['aiohttp'] = None` in the test
 - Upgrade aiohttp

### DIFF
--- a/requirements-aiohttp.txt
+++ b/requirements-aiohttp.txt
@@ -1,4 +1,4 @@
-aiohttp>=2.2.5,<3.5.2
+aiohttp>=2.2.5
 aiohttp-swagger>=1.0.5
 ujson>=1.35
 aiohttp_jinja2==0.15.0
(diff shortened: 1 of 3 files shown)


cognifloyd:
@dtkav wrote #844 and @hjacobs approved it. What do you think of the fix here so that we revert #844?

kornicameister:
Yeah, they some things happening regarding imports and so on.

hjacobs:
:+1:

PR #952, opened 2019-05-22
Respond with problems by default in aiohttp.

cognifloyd:
AioHttp is not responding with problems yet. This uses an aiohttp middleware to handle exceptions and return a problem response.

Done:
- [x] Handle all exceptions handled by aiohttp to make sure they return problem json.
- [x] Fix the Unauthorized/OAuthProblem return not having problem json body
- [x] All werkzeug and aiohttp exceptions are converted to problem json responses preserving status/description/title.
- [x] All exceptions become problem json.
- [x] All current tests pass without modification.
- [x] Add tests for problem+json output when using aiohttp. Reuses the fixture and overall logic from `test_errors`.

Open questions:
 - Do any docs need to change?
   - This is just bringing AioHTTP exception handling into line with current Flask exception handling.
   - https://connexion.readthedocs.io/en/latest/exceptions.html is out of date: https://github.com/zalando/connexion/blob/master/docs/exceptions.rst
   - https://connexion.readthedocs.io/en/latest/response.html#error-handling also out of date (raise is better): https://github.com/zalando/connexion/blob/master/docs/response.rst
   - There are essentially NO docs on aiohttp except for a brief reference here: https://github.com/zalando/connexion/blob/6bf058f6b7c20d87537eaa8bb3f701f69b383765/docs/quickstart.rst#L86-L93

 - Return problems by default in aiohttp.
 - Send the debug logs and tracebacks like aiohttp does.
 - Drop `oauth_problem_middleware` in favor of handling all the problem output in `problems_middleware` because `oauth_problem_middleware` didn't actually return valid problem json, but this does.

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -3,8 +3,13 @@ import logging
 import re
 import traceback
 from contextlib import suppress
-from http import HTTPStatus
 from urllib.parse import parse_qs
+
+try:
+    from http import HTTPStatus
+except ImportError:
+    # httpstatus35 backport for python 3.4
+    from httpstatus import HTTPStatus
 
 import aiohttp_jinja2
 import jinja2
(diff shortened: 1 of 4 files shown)


cognifloyd:
This is ready for review, and hopefully to be merged :)

jmcs:
:+1:

Issue #935, opened 2019-05-07
multipart/form-data content validation


maximumG:
### Description

I'm trying to have some validation on a REST route accepting a multipart/form-data content. This multipart is composed of 2 contents:

* a ZIP file named `stack_zip`  (application/zip)
* a form-data named `variables` (application/json)

![multipart](https://user-images.githubusercontent.com/4469833/57282086-a2d40300-70ac-11e9-9911-61fa66f1fbbb.png)

### Expected behaviour

The `variables` content should be validated against a schema defined in openapi.yaml.

### Actual behaviour

The `variables` content is received by connexion as a raw string even if the `Content-Type` is set to `application/json`.

It also seems that Werkzeug is:

* keeping the headers when parsing the a File object  (application/zip)
* removing the headers when parsing any non-file object

### Steps to reproduce

OpenAPI schema.

```yaml

  /zones/{zoneId}/deploy_stack:
    post:
      tags:
      - Stack
      summary: Deploy a Docker Stack
      operationId: api.v1.view.deploy_stack
      responses:
        200:
          description: Successful response - Docker stack deployed.
        404:
          description: The specified `Zone` does not exist.
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                variables:
                  type: object
                  properties:
                    stack_name:
                      type: string
                stack_zip:
                  type: string
                  format: binary
            encoding:
              stack_zip:
                contentType: application/zip
              variables:
                contentType: application/json

```

Client command:

```bash

curl -i -X POST 'http://localhost:9090/v1/zones/bo/deploy_stack'           -H 'Content-type:multipart/form-data'           -F 'stack_zip=@artifacts.zip;type=application/zip'           -F 'variables={"stak_name": "bar"}; type=application/json'
```

### Additional info:

 I tested this feature against the connexion 2.0 main branch as well as the branch from https://github.com/zalando/connexion/pull/760

I guess it would be nice if the library could parse the encoding section in the YAML specs and try to convert the form in a JSON object in the content-type is set to application/json. I guess I could start by tryiing to modify the code in #760 and add support for encoding.

maximumG:
## Deserialize complex formdata

I investigated deeper this problem to find any workaround. Basically I modified the *uri_parsing.py* module with the following code in the *OpenAPIURIParser* class.

```python

   def resolve_form(self, form_data):
        if self._body_schema is None or self._body_schema.get('type') != 'object':
            return form_data
        for k in form_data:
[...]

PR #954, opened 2019-05-22
Fix for discarded form data

rsnyman:
Attempt to fix a situation where form data is thrown away when the requestBody has no name.

This PR fixes a problem I'm having where a POST requestBody without a name is being discarded. I'm not sure if this is the correct way to do it, but this solves my problem. Please let me know if you know of a better way of doing this.

See issue #935 for more information about this.

 - Return parameters when the content type is "multipart/form-data"

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -5,6 +5,7 @@ from connexion.operations.abstract import AbstractOperation
 
 from ..decorators.uri_parsing import OpenAPIURIParser
 from ..utils import deep_get, is_null, is_nullable, make_type
+from ..http_facts import FORM_CONTENT_TYPES
 
 logger = logging.getLogger("connexion.operations.openapi3")
 
@@ -273,6 +274,9 @@ class OpenAPIOperation(AbstractOperation):
 
         if x_body_name in arguments or has_kwargs:
             return {x_body_name: res}
+        # If the content type is "multipart/form-data" return the properties as-is
+        if self.consumes[0] in FORM_CONTENT_TYPES:
+            return res
         return {}
 
     def _sanitize_body_argument(self, body_arg, body_props, additional_props, sanitize):


connexion/operations/openapi.py
```diff
@@ -273,6 +273,9 @@ def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
 
         if x_body_name in arguments or has_kwargs:
             return {x_body_name: res}
+        # If the content type is "multipart/form-data" return the properties as-is
+        if self.consumes[0] == 'multipart/form-data':
+            return res
         return {}
```
kornicameister: I am pretty sure it might be sufficient to just return `res` in here. 
In method that calls the one you modified, `request.form.items` is put into `request_body`.

kornicameister: Also, that's is needed for #987 to work.

connexion/operations/openapi.py
```diff
@@ -273,6 +273,9 @@ def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
 
         if x_body_name in arguments or has_kwargs:
             return {x_body_name: res}
+        # If the content type is "multipart/form-data" return the properties as-is
+        if self.consumes[0] == 'multipart/form-data':
```
kornicameister: I believe it might be better to use this code. `connexion` maintains own list of what it understands as multipart.
```suggestion
        if self.consumes[0] in http_facts.FORM_CONTENT_TYPES:
```

connexion/operations/openapi.py
rsnyman: I just did it like this because I wasn't entirely sure if a non-empty response was always going to be the correct response. I can most certainly change it though.

connexion/operations/openapi.py
rsnyman: Ah, that's definitely better than mine :-)

connexion/operations/openapi.py
kornicameister: I pulled that version into #987 and there are some tests failing, so yours might be better TBH.
Rebase on top of master BTW, there are some changes that should make it better with tests and for you to see if direction is fine.

Tohnmeister:
Why was this pull request closed? It doesn't appear to merged into master/main ever, yet it was a fix for an issue I'm facing with multipart form data.

connexion/operations/openapi.py
RichardBruskiewich: > 
> 
> I believe it might be better to use this code. `connexion` maintains own list of what it understands as multipart.
> 
> PS. I am not sure if amounts of dots is OK, but there is for sure a module `http_facts` in root package.

Not sure about just checking the FORM_CONTENTS_TYPES if one specifically needs to check for 'multipart/form-data'?

PR #955, opened 2019-05-22
 Converting response to raise a ProblemException

badcure:
The goal of this to convert all possible error responses to be a raised ProblemException subclass instead. Also for flask, allow for the option to skip the error handling.

The goal is to allow the end application handle the exceptions, as well as customize the response  that aligns with the application's payload structure.

### DIFF
--- a/connexion/__init__.py
+++ b/connexion/__init__.py
@@ -8,7 +8,6 @@ from .decorators.produces import NoContent  # NOQA
 from .exceptions import ProblemException  # NOQA
 # add operation for backwards compatability
 from .operations import compat
-from .problem import problem  # NOQA
 from .resolver import Resolution, Resolver, RestyResolver  # NOQA
 
 full_name = '{}.operation'.format(__package__)
--- a/connexion/__init__.py
+++ b/connexion/__init__.py
@@ -8,6 +8,7 @@ from .decorators.produces import NoContent  # NOQA
 from .exceptions import ProblemException  # NOQA
 # add operation for backwards compatability
 from .operations import compat
+from .problem import problem  # NOQA
 from .resolver import Resolution, Resolver, RestyResolver  # NOQA
 
 full_name = '{}.operation'.format(__package__)
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -55,7 +55,8 @@ def problems_middleware(request, handler):
     try:
         response = yield from handler(request)
     except ProblemException as exc:
-        response = exc.to_problem()
+        response = problem(status=exc.status, detail=exc.detail, title=exc.title,
+                           type=exc.type, instance=exc.instance, headers=exc.headers, ext=exc.ext)
     except (werkzeug_HTTPException, _HttpNotFoundError) as exc:
         response = problem(status=exc.code, title=exc.name, detail=exc.description)
     except web.HTTPError as exc:
--- a/connexion/apps/abstract.py
+++ b/connexion/apps/abstract.py
@@ -14,7 +14,7 @@ logger = logging.getLogger('connexion.app')
 class AbstractApp(object):
     def __init__(self, import_name, api_cls, port=None, specification_dir='',
                  host=None, server=None, arguments=None, auth_all_paths=False, debug=False,
-                 resolver=None, options=None):
+                 resolver=None, options=None, skip_error_handlers=False):
         """
         :param import_name: the name of the application package
         :type import_name: str
@@ -63,8 +63,9 @@ class AbstractApp(object):
 
         logger.debug('Specification directory: %s', self.specification_dir)
 
-        logger.debug('Setting error handlers')
-        self.set_errors_handlers()
+        if not skip_error_handlers:
+            logger.debug('Setting error handlers')
+            self.set_errors_handlers()
 
     @abc.abstractmethod
     def create_app(self):
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -52,12 +52,13 @@ class InvalidSpecification(ConnexionException, ValidationError):
     pass
 
 
-class NonConformingResponse(ConnexionException):
+class NonConformingResponse(ProblemException):
     def __init__(self, reason='Unknown Reason', message=None):
         """
         :param reason: Reason why the response did not conform to the specification
         :type reason: str
         """
+        super(NonConformingResponse, self).__init__(status=500, title=reason, detail=message)
         self.reason = reason
         self.message = message
 
@@ -66,6 +67,18 @@ class NonConformingResponse(ConnexionException):
 
     def __repr__(self):  # pragma: no cover
         return '<NonConformingResponse: {}>'.format(self.reason)
+
+
+class BadRequestProblem(ProblemException):
+
+    def __init__(self, title='Bad Request', detail=None):
+        super(BadRequestProblem, self).__init__(status=400, title=title, detail=detail)
+
+
+class UnsupportedMediaTypeProblem(ProblemException):
+
+    def __init__(self, title="Unsupported Media Type", detail=None):
+        super(UnsupportedMediaTypeProblem, self).__init__(status=415, title=title, detail=detail)
 
 
 class NonConformingResponseBody(NonConformingResponse):
--- a/connexion/exceptions.py
+++ b/connexion/exceptions.py
@@ -1,8 +1,6 @@
 from jsonschema.exceptions import ValidationError
 from werkzeug.exceptions import Forbidden, Unauthorized
 
-from .problem import problem
-
 
 class ConnexionException(Exception):
     pass
@@ -22,12 +20,6 @@ class ProblemException(ConnexionException):
         self.instance = instance
         self.headers = headers
         self.ext = ext
-
-    def to_problem(self):
-        return problem(status=self.status, title=self.title, detail=self.detail,
-                       type=self.type, instance=self.instance, headers=self.headers,
-                       ext=self.ext)
-
 
 class ResolverError(LookupError):
     def __init__(self, reason='Unknown reason', exc_info=None):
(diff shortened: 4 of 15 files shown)


connexion/handlers.py
```diff
@@ -45,12 +45,11 @@ def handle(self, *args, **kwargs):
         """
         Actual handler for the execution after authentication.
         """
-        response = problem(
+        raise AuthenticationProblem(
```
cognifloyd: This is causing the error in `tests/aiohttp/test_aiohttp_api_secure.py::test_auth_all_paths[pyloop]` because if no middleware handles the raised exception, then aiohttp swallows the response converting it to a 500 error:
```
>       assert get_inexistent_endpoint.status == 404
E       assert 500 == 404
E         -500
E         +404
```
You might need something like #952 to add that middleware for aiohttp.

badcure: Ahh okay, thanks! I will work on that today. I was confused why it was happening last night, but it was getting late for me..

It sort of makes sense now that you pointed it out.

badcure:
Side note: In reviewing some of the errors "test_default_object_body", I plan on removing those tests.

While I have no idea how it was passing before, reviewing https://github.com/json-schema/json-schema/issues/5 seems to suggest that default should NOT be applied when something is required. Below is the exception I am seeing for this case:
```
  File "/connexion/connexion/decorators/validation.py", line 178, in validate_schema
    self.validator.validate(data)
  File "/connexion/venv/py27/lib/python2.7/site-packages/jsonschema/validators.py", line 123, in validate
    raise error
ValidationError: None is not of type 'object'

Failed validating 'type' in schema:
    {'components': {'examples': {'justAnExample': {'summary': 'a basic example.',
                                                   'value': 'Good evening, doctor.'}},
                    'parameters': {'OptionalName': {'description': 'Name of the person to greet.',
                                                    'in': 'path',
                                                    'name': 'name',
                                                    'required': False,
                                                    'schema': {'type': 'string'}}},
                    'requestBodies': {'fakeapi.hello.test_body_sanitization_body': {'content': {'application/json': {'schema': {'components': <Recursion on dict with id=140022751344632>,
                                                                                                                                'properties': {'body1': {'type': 'string'},
                                                                                                                                               'body2': {'type': 'string'}},
                                                                                                                                'type': 'object'}}},
                                                                                    'description': 'Just a testing parameter in the body',
                                                                                    'required': True}},
                    'schemas': {'new_stack': {'properties': {'image_version': {'description': 'Docker image version to deploy',
                                                                               'type': 'string'}},
                                              'required': ['image_version'],
                                              'type': 'object'}}},
     'default': {'image_version': 'default_image'},
     'properties': {'image_version': {'description': 'Docker image version to deploy',
                                      'type': 'string'}},
     'required': ['image_version'],
     'type': 'object',
     'x-body-name': 'stack',
     'x-scope': ['']}
```

connexion/apps/abstract.py
```diff
@@ -63,8 +63,9 @@ def __init__(self, import_name, api_cls, port=None, specification_dir='',
 
         logger.debug('Specification directory: %s', self.specification_dir)
 
-        logger.debug('Setting error handlers')
-        self.set_errors_handlers()
+        if not skip_error_handlers:
+            logger.debug('Setting error handlers')
+            self.set_errors_handlers()
```
cognifloyd: For aiohttp, the `set_error_handlers` does not actually register the exception handler. Instead the exception handling middleware is registered in `api`. So, the `skip_error_handlers` will have no effect for aiohttp. It would be nice to unify the API for Flask and Aiohttp to add the error handlers in the same way, but I don't think that fits in this PR.

connexion/apis/aiohttp_api.py
```diff
@@ -27,7 +29,7 @@
 
 @web.middleware
 @asyncio.coroutine
-def oauth_problem_middleware(request, handler):
+def generic_problem_middleware(request, handler):
```
cognifloyd: This will need to be rebased on master as #952 was just merged.
Now, aiohttp turns all exceptions into problems by default.

This will need to at least replace `exc.to_problem()` with `problem_from_object(exc)`.

connexion/__init__.py
```diff
@@ -8,7 +8,6 @@
 from .exceptions import ProblemException  # NOQA
 # add operation for backwards compatability
 from .operations import compat
-from .problem import problem  # NOQA
```
cognifloyd: Raising ProblemExceptions is nicer and much better for everything in core connexion, but removing the ability to use `problem()` by importing directly from `connexion` is a major API change.

connexion/apps/abstract.py
cognifloyd: Maybe this should be a separate PR as well - Use this PR to convert all uses of `problem()` to `raise ProblemException()` and then another PR to add `skip_error_handlers`.
And then at some point someone can harmonize how the Flask and Aiohttp clients setup (or skip setting up) error handling.

connexion/problem.py
```diff
@@ -40,3 +40,10 @@ def problem(status, title, detail, type=None, instance=None, headers=None, ext=N
     return ConnexionResponse(status, mimetype, content_type,
                              body=problem_response,
                              headers=headers)
+
+
+def problem_from_object(exception):
```
cognifloyd: I'm not sure about the naming here (though it's no worse than `to_problem` which is also odd).

What is the benefit of moving this factory method from `ProblemException.to_problem` into a separate factory function?

badcure: IMO, I have always viewed exceptions as simply exceptions. I think tracing a response through an exception(and back up to other code) can be very confusing at times. I have always found out easier to think of it as "Okay we now have this exception, let me use this other function to convert it to a response for me". 

Also, any suggestions for the name? I have no opinion on the names, I just picked something that looked similar. If you have a more preferred naming convention, I can change it.

phillipmadlan:
Any update on this PR and possibility of merging @cognifloyd? Thanks!

cognifloyd:
I do not work for Zalando, so I can't merge any PRs here.
But, I would love to see this included, so I've been reviewing trying to help move it along so that when the Zalando people have some time, they can just merge it.

To that end, this PR has a merge conflict since #952 was merged. I would recommend rebasing.

ConstantinoSchillebeeckxSimpleRose:
Any movement on this, I'd love to see it merged?

hjacobs:

ConstantinoSchillebeeckxSimpleRose:
[All set](https://github.com/zalando/connexion/pull/1054).

connexion/apis/aiohttp_api.py
cognifloyd: @badcure when you rebase, please also update _generic_problem which has been added to this file.

badcure:
@ConstantinoSchillebeeckxSimpleRose and @hjacobs  Sorry for the MIA-ness. I have updated the PR with the changes. Let me know what you think

connexion/__init__.py
hjacobs: I agree, we should not break the API without need

connexion/exceptions.py
```diff
@@ -23,11 +21,6 @@ def __init__(self, status=400, title=None, detail=None, type=None,
         self.headers = headers
         self.ext = ext
 
-    def to_problem(self):
-        return problem(status=self.status, title=self.title, detail=self.detail,
-                       type=self.type, instance=self.instance, headers=self.headers,
-                       ext=self.ext)
-
```
cognifloyd: I think we should leave this here with a deprecation warning. Removing this here is not backwards compatible.

@hjacobs What is the policy on connexion deprecations for connexion? ProblemException.to_problem was one of the primary interfaces for creating problems. It is no longer necessary with this change, but I would hate to break any code that depends on it without either a deprecation or a major version bump. Thoughts?

tests/test_validation.py
```diff
@@ -34,40 +36,54 @@ def orig_handler(*args, **kwargs):
...
+    assert exc.value.detail == "Wrong type, expected 'integer' for path parameter 'p1'"
     request = MagicMock(path_params={'p1': '1.2'}, **kwargs)
-    assert json.loads(handler(request).data.decode())['detail'] == "Wrong type, expected 'integer' for path parameter 'p1'"
+    with pytest.raises(BadRequestProblem) as exc:
+        handler(request)
+    assert exc.value.detail == "Wrong type, expected 'integer' for path parameter 'p1'"
 
-    request = MagicMock(path_params={'p1': 1}, query={'q1': '4'}, headers={}, cookies={})
-    assert json.loads(handler(request).data.decode())['detail'].startswith('4 is greater than the maximum of 3')
+    request = MagicMock(path_params={'p1': 1}, query={'q1': '4'}, headers={})
```
cognifloyd: What difference does `, cookies={}` make? Why did you remove it here and in the next several changed request = lines?

tests/test_validation.py
```diff
@@ -34,40 +36,54 @@ def orig_handler(*args, **kwargs):
...
     assert handler(request) == 'OK'
 
-    request = MagicMock(path_params={'p1': '123'}, query={}, headers={'h1': 'x'}, cookies={})
-    assert json.loads(handler(request).data.decode())['detail'].startswith("'x' is not one of ['a', 'b']")
-
-    request = MagicMock(path_params={'p1': '123'}, query={}, headers={}, cookies={'c1': 'b'})
-    assert handler(request) == 'OK'
-
-    request = MagicMock(path_params={'p1': '123'}, query={}, headers={}, cookies={'c1': 'x'})
-    assert json.loads(handler(request).data.decode())['detail'].startswith("'x' is not one of ['a', 'b']")
```
cognifloyd: Why did you drop these cookies tests?

connexion/exceptions.py
hjacobs: Yes, let's leave it here with deprecation note (and create a follow-up issue to clean up in a future release)

tests/test_validation.py
badcure: The issue is that request.cookies was `mock.Mock` object. This was causing an issue in the underling code.

tests/test_validation.py
badcure: My accident. Added them back. Let me know if there is anything else with the unit tests.

connexion/exceptions.py
badcure: Added it back with a deprecated warning. Let me know if the message should be changed.

connexion/__init__.py
badcure: This had been added back

connexion/handlers.py
badcure: Resolving this since everything seems to be working now. Let me know if something is off.

connexion/apps/abstract.py
badcure: Resolving since everything appears to be passing, but let me know if you feel something else should be changed.

hjacobs:
:+1:

Issue #546, opened 2017-11-16
What's the easiest way to implement acceptance/unit tests for generated Flask app?


acidjunk:
I'm currently in the process of writing acceptance tests for a connexion based API in Flask. As a Flask user i'm used to test stuff with py.test and the pip modules: [pytest-flask](https://pytest-flask.readthedocs.io/en/latest/tutorial.html#step-2-configure) and / or [Flask_Testing](https://flask-testing.readthedocs.io/en/latest/)

Both have helpers to provide some config and start a live server when needed. This doesn't seem to work with the Flask app created by Connexion. This also makes it nearly impossible to use Flask's config system and the url_for() methods.

Is there a way to use the testing plugins or to create a more Flask compliant app?

Possibly related to #489 and #364

acidjunk:
This is how I use it now:
I added a construction to the place where the app is inited, so that it can return a flask app when needed, something like:
```
def run(actually_run=True):
    app = connexion.App(__name__, specification_dir='./swagger/')
    api = app.add_api(swagger_file)
    if actually_run:
        app.run(host=HOST, port=PORT, ...)
    else:
        return app.app
```

Then a live server can be started from tests that need it:
```
import requests
from flask_testing import LiveServerTestCase
from main import run_application

class MyTest(LiveServerTestCase):

    def create_app(self):
        app = run(False)
        return app

    def test_server_is_up_and_running(self):
        response = requests.get(f"{self.get_server_url()}/an_endpoint_for_your_api")
        self.assertEqual(response.status_code, 200)
```

dirkschneemann:
I had a good experience using the `WebTest` library in the past (`pip install webtest`): https://docs.pylonsproject.org/projects/webtest/en/latest/
It will test your application on [WSGI level](https://en.wikipedia.org/wiki/Web_Server_Gateway_Interface) which has some benefits:
* It is independent of the underlying implementation framework - your tests will work now with `connexion/Flask` and keep working if you decide to go for the `gevent` or `tornado` or any other WSGI-compatible backend in the future.
* Everything works on the WSGI-API level, no need to start a "real" webserver in the background and shut it down again after your tests.
* The tests can be run with all usual Python unittest runners (the built-in one, py.test, nose etc.)

The documentation of `WebTest` is quite extensive and covers all major use cases. A quick example to get started:
```
from webtest import TestApp
import unittest

from main import app # the connexion Flask app object

class TestEndpoints(unittest.TestCase):
[...]

PR #957, opened 2019-05-23
Attempting to provide an example on how to set it up and unit tests.

### LINKED ISSUE #489: Add testing documentation?
It would be nice to have documentation / and / or examples of what are good workflows for testing a connexion app's API.

badcure:
Fixes #489 .

 - Adds its own tox test section.
 - Creates a simple flask API from an open API schema
 - Unit tests/Integration interacts with the simple API.

Note: Looks like a couple issues are showing up, had to skip a few tests for now.

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -1,0 +1,90 @@
+from __future__ import absolute_import
+
+import unittest
+import json
+import mock
+
+from tests.integration import flask_setup
+
+
+class BaseTest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        super(BaseTest, cls).setUpClass()
+        flask_setup.setupApp()
+
+    def setUp(self) -> None:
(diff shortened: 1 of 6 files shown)


tests/integration/base.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import absolute_import
```
dtkav: I think it would be more consistent across the repo to use pytest form, rather than unittest form.

badcure: This PR is more to address #546 than expand on the unit tests, which is to help set up guidance how others can set up integration tests. 

My preference is to keep the examples more generic than pytest specific, since not everything in python land uses pytest. With that said, I can change it to use pytest(i.e. consistent with everything else)...but that IMO implies that is the only supported way of testing.

hjacobs: Yes, please use pytest

RobbeSneyders:
Closing due to lack of contributor feedback.

PR #971, opened 2019-06-07
handle deepObject and explode: true

### LINKED ISSUE #970: objects with style: deepObject and explode: true are not supported
### Description

https://swagger.io/docs/specification/serialization/#query defines possible serialization styles. Connexion is not supporting at least one of them (style: deepObject).

might be related to https://github.com/zalando/connexion/issues/963

### Expected behaviour

`/users?id[role]=admin&id[firstName]=Alex` would result in an object/dict named id and the value `{"role": "admin", "firstName": "Alex"}`

### Actual behaviour

when issuing `/users?id[role]=admin&id[firstName]=Alex` the id parameter is not passed to the function linked to that endpoint.

### Steps to reproduce

create an endpoint with the following parameter:

```
- name: id
  required: true
  in: query
  style: deepObject
  explode: true
  schema:
    type: object
    properties:
      bla:
        type: string
```

try to query that endpoint with the query parameter `?id[role]=admin&id[firstName]=Alex`. There will be no data passed to the function.

### Additional info:

Output of the commands:

- `python --version` : Python 3.6.7
- `pip show connexion | grep "^Version\:"` : 2.2.0

drummerwolli:
Fixes #970 .

 - handle deepObject style and explode true params introduced in OAS3


drummerwolli:
question: this new feature is opnly available in OAS3. if i correctly checked the tests, then there is currently no scenario set up for this (different results when using OAS3 vs. OAS2), or? since i believe there will be more things like this in the future, how would you rather handle this?

drummerwolli:
@hjacobs @prawn-cake @jmcs this is the PR we talked in chat about ...

connexion/operations/abstract.py
```diff
@@ -184,10 +185,20 @@ def _get_val_from_param(self, value, query_defn):
         Convert input parameters into the correct type
         """
 
+    @abc.abstractmethod
+    def _get_val_from_object_param(self, value, query_defn):
+        """
+        Convert object input parameters into the correct type
+        """
+
     def _query_args_helper(self, query_defns, query_arguments,
```
prawn-cake: I might've overlooked it, but it would be nice to have a separate function which parses query string into a dictionary without the need to create an instance of a class of other dependencies and it's easier to test, e.g

```
def parse_object_str(query_str: str) -> dict
...

# The test would be
assert parse_object_str("id[role]=admin&id[firstName]=Alex") == {"role": "admin", "firstName": "Alex"}

```

drummerwolli: agreed, but this function takes also non-object style parameters. so the test would actually look like this:

```
def parse_object_str(query_str: str) -> dict
...

# The test would be
assert parse_object_str("foo=bar&id[role]=admin&id[firstName]=Alex") == {"foo": "bar", "id": {"role": "admin", "firstName": "Alex"}}
```

your example was filtered on one key only, which could also make sense, but that requires dissecting the query parameter extra again. i would have one function which takes query parameters, no matter what format, and returns a certain style of values back ...

connexion/operations/abstract.py
```diff
@@ -199,7 +210,12 @@ def _query_args_helper(self, query_defns, query_arguments,
                     logger.error("Function argument '{}' not defined in specification".format(key))
                 else:
                     logger.debug('%s is a %s', key, query_defn)
-                    res[key] = self._get_val_from_param(value, query_defn)
+                    if type(value) == dict:
```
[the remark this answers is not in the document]
drummerwolli: fixed, thx

connexion/decorators/validation.py
```diff
@@ -46,6 +67,8 @@ def coerce_type(param, value, parameter_type, parameter_name=None):
 
     def make_type(value, type_literal):
         type_func = TYPE_MAP.get(type_literal)
+        if type(value) == list and len(value) == 1 and type_literal != 'array':
```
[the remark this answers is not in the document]
drummerwolli: fixed, thx

dtkav:
It's been a while since I worked on this, but IIRC there was some part of this that needed to happen early in the request flow (maybe before schema validation?).

Have you seen this file: https://github.com/zalando/connexion/blob/master/connexion/decorators/uri_parsing.py#L149 ?

I think object-level and array-level decoding was happening in that file, and then basic type casting happens later on.
Hopefully that's more helpful than it is confusing - I'll have to sit down and re-read the code to get more insight.

drummerwolli:
@dtkav it was absolutely helping. i was wondering already why i had to implement param parsing twice. but this file was the key, now i just once disect the object :+1: 

converted this PR to a real one, to triger the tests

drummerwolli:
I consider this now ready4review ...

connexion/decorators/uri_parsing.py
```diff
@@ -96,8 +97,22 @@ def resolve_params(self, params, _in):
...
         for k, values in params.items():
+            groups = fullmatch(r'^(\w+)\[{1}(\w+)\]{1}$', k)
+            if groups:
+                possible_key = groups.group(1)
+                param_defn = self.param_defns.get(possible_key)
+                if param_defn \
+                        and param_defn.get('style', None) == 'deepObject' and param_defn.get('explode', False):
+                    value = {groups.group(2): values}
+                    key = possible_key
+                    if key not in resolved_param:
```
dtkav: consider using setdefault

tests/conftest.py
```diff
@@ -13,6 +13,7 @@
 FIXTURES_FOLDER = TEST_FOLDER / 'fixtures'
 SPEC_FOLDER = TEST_FOLDER / "fakeapi"
 SPECS = ["swagger.yaml", "openapi.yaml"]
+OPENAPI_SPEC_ONLY = ["openapi.yaml"]
```
dtkav: IIRC "openapi" now encompasses both "v2" (fka swagger) and "v3".
so in general, it's good to clarify that you mean v3.

tests/decorators/test_validation.py
```diff
@@ -43,6 +43,13 @@ def test_get_nullable_parameter():
     assert result is None
 
 
+def test_get_explodable_object_parameter():
```
dtkav: what does a non-exploded deepObject look like?
it's unclean from the spec, but do you support more than one level deep like foo[bar][baz]=5?

connexion/operations/openapi.py
```diff
@@ -324,3 +324,27 @@ def _get_val_from_param(self, value, query_defn):
             return [make_type(part, query_schema["items"]["type"]) for part in value]
         else:
             return make_type(value, query_schema["type"])
+
+    def _get_val_from_object_param(self, value, query_defn):
```
dtkav: why not just bake this into _get_val_from_param?

connexion/operations/abstract.py
```diff
@@ -199,7 +205,12 @@ def _query_args_helper(self, query_defns, query_arguments,
                     logger.error("Function argument '{}' not defined in specification".format(key))
                 else:
                     logger.debug('%s is a %s', key, query_defn)
-                    res[key] = self._get_val_from_param(value, query_defn)
+                    if isinstance(value, dict):
+                        if key not in res:
+                            res[key] = {}
+                        res[key].update(self._get_val_from_object_param(value, query_defn))
```
dtkav: if this is the only place where this is called, I would not add another function here, and instead let _get_val_from_param handle the type checking and returning the correct response.

connexion/operations/openapi.py
```diff
@@ -324,3 +324,27 @@ def _get_val_from_param(self, value, query_defn):
...
+        query_schema = query_defn["schema"]
+
+        if is_nullable(query_schema) and is_null(value):
+            return None
+
+        if query_schema["type"] == "object" and 'properties' in query_schema:
+            return_dict = {}
+            for prop_key in query_schema['properties'].keys():
+                for val_key in value.keys():
```
dtkav: does this need to be a for-loop?

connexion/utils.py
```diff
@@ -253,3 +254,8 @@ def ignore_aliases(self, *args):
     yaml.representer.SafeRepresenter.represent_scalar = my_represent_scalar
 
     return yaml.dump(openapi, allow_unicode=True, Dumper=NoAnchorDumper)
+
+
+def fullmatch(regex, string, flags=0):
+    """Emulate python-3.4 re.fullmatch()."""
```
dtkav: I was confused by this comment because python 3.4 is at End Of Life.
what about 
`backport of re.fullmatch() in python 3`

connexion/decorators/validation.py
```diff
@@ -63,6 +86,20 @@ def make_type(value, type_literal):
...
+    elif param_type == 'object':
+        converted_params = {}
+        if 'properties' in param_schema and param_schema['properties']:
+            for k, v in value.items():
+                try:
+                    converted_params[k] = make_type(v, param_schema['properties'][k]['type'])
+                except (ValueError, TypeError):
+                    converted_params[k] = v
+                except KeyError:
+                    if 'additionalProperties' not in param_schema or not param_schema['additionalProperties']:
```
dtkav:
```
if not params_schema.get("additionalProperties", True):
    ...
```

connexion/decorators/uri_parsing.py
```diff
@@ -96,8 +97,22 @@ def resolve_params(self, params, _in):
         """
         resolved_param = {}
         for k, values in params.items():
+            groups = fullmatch(r'^(\w+)\[{1}(\w+)\]{1}$', k)
+            if groups:
+                possible_key = groups.group(1)
+                param_defn = self.param_defns.get(possible_key)
+                if param_defn \
+                        and param_defn.get('style', None) == 'deepObject' and param_defn.get('explode', False):
```
dtkav: check out `_resolve_parameter_duplicates(...)` and `_split(...)`
these should get rid of the array unwrapping that you are doing in other places.

connexion/decorators/validation.py
```diff
@@ -46,6 +65,10 @@ def coerce_type(param, value, parameter_type, parameter_name=None):
 
     def make_type(value, type_literal):
         type_func = TYPE_MAP.get(type_literal)
+        if isinstance(value, list) and len(value) == 1 and type_literal != 'array':
```
dtkav: can this array unwrapping happen in uri_parsing ?

connexion/operations/openapi.py
```diff
@@ -324,3 +324,27 @@ def _get_val_from_param(self, value, query_defn):
...
+                    if prop_key == val_key:
+                        try:
+                            if type(value[val_key]) == list and len(value[val_key]) == 1:
```
dtkav: Can this array unwrapping be done in uri_parsing?

tests/decorators/test_validation.py
drummerwolli: > what does a non-exploded deepObject look like?

according to the [spec](https://swagger.io/docs/specification/serialization/#query) thats not possible and not a thing. not sure how to handle this.

>  but do you support more than one level deep like foo[bar][baz]=5?

will check on the possibility to include that ...

connexion/decorators/uri_parsing.py
drummerwolli: not sure what you mean with that. but i moved the array unnesting now to `uri_parsing`. would that be sufficient enough?

tests/decorators/test_validation.py
drummerwolli: > will check on the possibility to include that ...

added

drummerwolli:
@dtkav thx a lot for the thorough review. i pushed some changes regarding your comments.

please check again @hjacobs @jmcs @dtkav @prawn-cake

tests/fixtures/simple/openapi.yaml
```diff
@@ -150,6 +150,68 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  foo:
+                    type: object
+                    properties:
+                      foo2:
+                        type: string
```
dtkav: Can you test `int` or some other non-string type so that it catches type coercion errors?

connexion/utils.py
```diff
@@ -253,3 +254,17 @@ def ignore_aliases(self, *args):
...
 
     return yaml.dump(openapi, allow_unicode=True, Dumper=NoAnchorDumper)
+
+
+def fullmatch(regex, string, flags=0):
+    """Backport of re.fullmatch() for python version below 3.4."""
+    return re.match("(?:" + regex + r")\Z", string, flags=flags)
+
+
+def create_empty_dict_from_list(_list, _dict, _end_value):
```
dtkav: can you please comment what this function does?

connexion/utils.py
```diff
@@ -253,3 +254,17 @@ def ignore_aliases(self, *args):
...
+    current_key = _list[0]
+    _list.pop(0)
```
dtkav: can the two above lines be combined? `current_key = _list.pop(0)`

tests/fixtures/simple/openapi.yaml
dtkav: Can you also test a nested field with a default?

connexion/decorators/validation.py
```diff
@@ -19,10 +19,29 @@
 TYPE_MAP = {
     'integer': int,
     'number': float,
-    'boolean': boolean
+    'boolean': boolean,
+    'object': dict
 }
 
 
+class PropertyValidationError(Exception):
```
dtkav: I'm pretty sure jsonschema should handle nested objects and additionalProperties.
Is this really needed? What am I missing?

connexion/decorators/validation.py
```diff
@@ -63,6 +82,20 @@ def make_type(value, type_literal):
                 converted = v
             converted_params.append(converted)
         return converted_params
+    elif param_type == 'object':
+        converted_params = {}
+        if 'properties' in param_schema and param_schema['properties']:
+            for k, v in value.items():
+                try:
+                    converted_params[k] = make_type(v, param_schema['properties'][k]['type'])
```
dtkav: will this work for nested properties?

tests/fixtures/simple/openapi.yaml
```diff
@@ -150,6 +150,68 @@ paths:
...
+                  foo2:
+                    type: string
+              foofoo:
+                type: string
+      responses:
+        '200':
+          description: object response
+          content:
+            application/json:
+              schema:
```
dtkav: do you need to set `allowReserved: true`?
AFAICT your examples send non-percent-encoded `[ ]` characters

connexion/decorators/validation.py
drummerwolli: seems yes. the tests pass, and they use nested properties. or you see an additional test case missing?

tests/fixtures/simple/openapi.yaml
drummerwolli: shall we create a new issue to keep track of this? for us, this won't be needed for now.

connexion/decorators/validation.py
drummerwolli: forget it, the tests don't pass. will fix it.

drummerwolli:
@dtkav thx again for the feedback! i adopted your changes!

connexion/decorators/validation.py
```diff
@@ -63,6 +64,17 @@ def make_type(value, type_literal):
                 converted = v
             converted_params.append(converted)
         return converted_params
+    elif param_type == 'object':
+        converted_params = {}
+        if 'properties' in param_schema and param_schema['properties']:
```
dtkav: I tried implementing this, and I think this part needs to be recursive also.
I'll try to push up my branch for comparison.

connexion/decorators/uri_parsing.py
```diff
@@ -115,8 +133,21 @@ def resolve_params(self, params, _in):
             else:
                 resolved_param[k] = values[-1]
 
+        # set defaults if values have not been set yet
+        resolved_param = self.set_default_values(resolved_param, self.param_schemas)
```
dtkav: maybe this should go with the other defaults code?
( https://github.com/zalando/connexion/blob/master/connexion/operations/openapi.py#L308 )

dtkav:
Hey @drummerwolli - I played around with your branch a bit - have a look and see if any of my modifications look interesting?
https://github.com/zalando/connexion/compare/master...dtkav:fix_exploded_obejcts

connexion/decorators/validation.py
drummerwolli: i used your suggestion. very nice idea!

drummerwolli:
blocked by: https://github.com/zalando/connexion/issues/991

danie1k:
@drummerwolli could you check if your code will work with #1002 example specs?

drummerwolli:
@dtkav sry that it took me so long, but here again my changes. feel free to check again. @jmcs @hjacobs as well, if you want.

@danie1k i added your yaml changes into one test file quickly but got some new errors. i think this PR wont fix this issue :cry:

connexion/decorators/uri_parsing.py
```diff
@@ -96,8 +98,21 @@ def resolve_params(self, params, _in):
         """
         resolved_param = {}
         for k, values in params.items():
+            dict_keys = re.findall(r'\[(\w+)\]', k)
```
jmcs: Can you add some comments about what this does, including links to the specification?

drummerwolli: done, please check

drummerwolli:
@ancelotpinto @hjacobs @dtkav @jmcs @prawn-cake can we get this reviewed and merged somehow?

hjacobs:
:+1:

drummerwolli:
thx @hjacobs

dtkav:
The default setting code in the PR isn't scoped properly, so (for example) it will take query defaults and set the request.form to the query defaults.

I'll try to put up a minimal working example that shows this behavior and a fix.

dtkav:
#1092 shows the bug also

PR #973, opened 2019-06-13
Add sanitized_params option to allow disabling sanitization

### LINKED ISSUE #835: Sanitization removes sensible characters in API definition and json request body input
### Description

I am using Connexion v2.2.0 to build an API that handles json schemata.

Request bodies containing schemata with a field name containing characters other than [0-9a-zA-Z_] are removed while parsing the request and passed to the handlers sanitized. 
This leads to unexpected behavior. 

The sanitization done in method `parameter_to_arg` (in file `connexion/decorators/parameter.py`) removes sensible characters such as `$` that are used in both API definition and json body input such as in fields "$id" and "$schema". Those field names are common in json schematas following version 06, for example.

### Expected behaviour

For example, a valid json {"$id": "...", ...} is used in the call and it should be passed to the handlers as is.

### Actual behaviour

Following the above example, a valid json {"$id": "...", ...} used in the call is passed to the handlers as {"id": "...", ...}.

### Steps to reproduce

Assume the following API definition (`api.yaml`):
```yaml
openapi: "3.0.0"
info:
  title: "Some API"
  version: "0.0.1"
servers:
- url: /
paths:
  /test:
    post:
      operationId: app.test
      requestBody:
        required: true
        content:
          application/json:
            schema:
              x-body-name: body
              type: object
              properties:
                "$id":
                  type: string
      responses:
        200:
          description: Response.
```
that uses `test(body)`, in `app.py`, as its handler, which is defined as
```python
import connexion

def test(body):
    import json
    print(json.dumps(body, indent=2))
    return json.dumps(body, indent=2), 200

app = connexion.FlaskApp(
    __name__,
    specification_dir="./")
app.add_api("api.yaml", validate_responses=True)
app.run(port=8080)
```

Calling the above endpoint with `curl -X POST -H "Content-Type: application/json" localhost:5001/test -d '{"$id": "myid"}'`, the output of the handler `test` is 
```json
{
  "id": "myid"
}
```
instead of the expected 
```json
{
  "$id": "myid"
}
```

### Additional info:

- `python --version`:  
  Python 3.6.7 :: Anaconda, Inc.

- `pip show connexion | grep "^Version\:"`:  
  Version: 2.2.0

ptamilarasan:
Fixes #835 .

 - Ability to disable sanitization

This is useful when the keys in a JSON post body contains characters like `-` or `$`

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -35,7 +35,8 @@ class AbstractAPI(object):
     def __init__(self, specification, base_path=None, arguments=None,
                  validate_responses=False, strict_validation=False, resolver=None,
                  auth_all_paths=False, debug=False, resolver_error_handler=None,
-                 validator_map=None, pythonic_params=False, pass_context_arg_name=None, options=None):
+                 validator_map=None, pythonic_params=False, sanitized_params=True,
+                 pass_context_arg_name=None, options=None):
         """
         :type specification: pathlib.Path | dict
         :type base_path: str | None
@@ -53,6 +54,8 @@ class AbstractAPI(object):
         :param pythonic_params: When True CamelCase parameters are converted to snake_case and an underscore is appended
         to any shadowed built-ins
         :type pythonic_params: bool
+        :param sanitized_params: When True convert parameter names into sanitized Python safe variable names
+        :type sanitized_params: bool
         :param options: New style options dictionary.
         :type options: dict | None
         :param pass_context_arg_name: If not None URL request handling functions with an argument matching this name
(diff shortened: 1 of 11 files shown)


dtkav:
I'd suggest we start with #1008 , and then see if it's still a problem for folks still using openapi2.
Let me know what you think!

ptamilarasan:
I'm fine with #1008 instead, as its probably the more correct way to handle this.  When I get a chance I'll test the other PR and see if it addresses our issue.  You mentioned openapi2 but we were affected by this in openapi 3.

hjacobs:
Is this still needed after merging #1008?

dtkav:
I'm pretty sure #1008 resolves this, so I'll close it for now.
Please re-open if I missed something!

PR #990, opened 2019-07-07
Remove Python 3.4

### LINKED ISSUE #977: Retire EOL Python 3.4
Python 3.4 is EOL.
> Python 3.4 has now reached its end-of-life and has been retired. No more releases will be made.
https://www.python.org/dev/peps/pep-0429/

Also, the async/await keywords are nicer to work with in 3.5+ (used by the aiohttp client).
https://www.python.org/dev/peps/pep-0478/
https://www.python.org/dev/peps/pep-0492/

When can we deprecate / remove support for python 3.4?

kornicameister:
Fixes #977 .

 - remove Python 3.4 support
 - upgrade Python version in docs
 - remove backport import to `HttpStatus`


kornicameister:
@hjacobs question still remain, should we drop old 2.7 as well ?
Dropping older python and starting with 3.5 would open up possibility to enhance codebase with `mypy` and `typing` which I could do :)

kornicameister:
I believe https://github.com/pytest-dev/pytest/issues/5532 contains reason for failure.

kornicameister:
I've created an issue here: https://github.com/zalando/connexion/issues/991
I think that before any actions are taken, that ought be taken care of.

kornicameister:
@hjacobs perhaps, if you have time, you can take a look at this one here.

hjacobs:
Thanks!

hjacobs:
:+1:

kornicameister:
Ok, guess that means we stick with 2.7 for now :+1:

PR #995, opened 2019-07-08
Bump flask to be at least 1.0.4

### LINKED ISSUE #991: pytest>=5.0.0 incompatible with Flask<=1.0.4
### Description

Master builds are failing due to error:
```
import_name = 'conftest'
    def find_package(import_name):
        """Finds a package and returns the prefix (or None if the package is
        not installed) as well as the folder that contains the package or
        module as a tuple.  The package path returned is the module that would
        have to be added to the pythonpath in order to make it possible to
        import the module.  The prefix is the path below which a UNIX like
        folder structure exists (lib, share etc.).
        """
        root_mod_name = import_name.split('.')[0]
        loader = pkgutil.get_loader(root_mod_name)
        if loader is None or import_name == '__main__':
            # import name is not found, or interactive/main module
            package_path = os.getcwd()
        else:
            # For .egg, zipimporter does not have get_filename until Python 2.7.
            if hasattr(loader, 'get_filename'):
                filename = loader.get_filename(root_mod_name)
            elif hasattr(loader, 'archive'):
                # zipimporter's loader.archive points to the .egg or .zip
                # archive filename is dropped in call to dirname below.
                filename = loader.archive
            else:
                # At least one loader is missing both get_filename and archive:
                # Google App Engine's HardenedModulesHook
                #
                # Fall back to imports.
                __import__(import_name)
                filename = sys.modules[import_name].__file__
            package_path = os.path.abspath(os.path.dirname(filename))
            # package_path ends with __init__.py for a package
>           if loader.is_package(root_mod_name):
E           AttributeError: 'AssertionRewritingHook' object has no attribute 'is_package'
.tox/py37-min/lib/python3.7/site-packages/flask/helpers.py:682: AttributeError
```

This can be tracked to following issues:
- https://github.com/pytest-dev/pytest/issues/5532
- https://github.com/pytest-dev/pytest/issues/5542

Indeed (either one of those actions):
- restricting `flask` to be at least `1.0.4` makes tests pass.
- restricting `pytest` to be at most `4.9.x` makes test pass

Which brings us down to final question.
Should `Flask` be kept at low version (which is pretty old TBH) or should `pytest` be restricted.
Personally I favor running newer version instead of supporting older codebases.

### Expected behaviour

master branch builds correctly against `min` environments.

### Actual behaviour

master branch builds correctly against `min` fail with aforementioned error.

### Steps to reproduce

Just run `tox -e py3(X)-min` where `X` stands for `Python` minor version {4,5,6,7}.

kornicameister:
Fixes #991 .

- bump `flask` to be at least `1.0.4`

Reasons:
- `flask==0.10.1` was released in 2013, which makes it quite old thus unsupported.
- On Flask webpage you can find an information pointing to `1.x` branch as latest stable
- different tools, including these used in `connexion` progressed leaving Python versions, for which `flask<1.0` has been developed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ install_requires = [
 ]
 
 swagger_ui_require = 'swagger-ui-bundle>=0.0.2'
-flask_require = 'flask>=0.10.1'
+flask_require = 'flask>=1.0.4'
 aiohttp_require = [
     'aiohttp>=2.3.10,<3.5.2',
     'aiohttp-jinja2>=0.14.0'
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -64,14 +64,14 @@ def test_verify_oauth_scopes_remote(monkeypatch):
     session.get = get_tokeninfo_response
     monkeypatch.setattr('connexion.decorators.security.session', session)
 
-    with pytest.raises(OAuthScopeProblem, message="Provided token doesn't have the required scope"):
+    with pytest.raises(OAuthScopeProblem, match="Provided token doesn't have the required scope"):
         wrapped_func(request, ['admin'])
 
     tokeninfo["scope"] += " admin"
     assert wrapped_func(request, ['admin']) is not None
 
     tokeninfo["scope"] = ["foo", "bar"]
-    with pytest.raises(OAuthScopeProblem, message="Provided token doesn't have the required scope"):
+    with pytest.raises(OAuthScopeProblem, match="Provided token doesn't have the required scope"):
         wrapped_func(request, ['admin'])
 
     tokeninfo["scope"].append("admin")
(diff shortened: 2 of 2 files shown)


kornicameister:
cc @hjacobs

kornicameister:
Hmm...

```
E       pytest.PytestDeprecationWarning: The 'message' parameter is deprecated.
E       (did you mean to use `match='some regex'` to check the exception message?)
E       Please see:
E         https://docs.pytest.org/en/4.6-maintenance/deprecations.html#message-parameter-of-pytest-raises
```

but that's unrelated to `flask` change. I guess we gotta change some tests as well to use this `match`.
I will create another issue and fix'em in separate PR to make changes atomic.

hjacobs:
Does it work when switching to the `match` parameter?

kornicameister:
@hjacobs it passed for `2.7`, `3.4`, `3.5`, `3.6` and `3.7` versions. Should I just add the changes to this PR?

hjacobs:
@kornicameister yes

kornicameister:
@hjacobs it went perfectly ok :)

hjacobs:
@kornicameister thanks!

hjacobs:
:+1:

PR #996, opened 2019-07-10
Update jsonschema version to 3.0.1 & JSON Schema draft to 7

### LINKED ISSUE #983: Support jsonschema3
It looks like we can support jsonschema3 easily now.

setup.py forces 'jsonschema>=2.5.1,<3.0.0'.  It looks like this came in with openapi-spec-validator, which required jsonschema<3.0.0.  In the meantime, openapi-spec-validator has added support for jsonschema3 starting with 0.2.7.

As a quick test, I bumped jsonschema up to 3.0.1 and connexion's unittests still pass.

danie1k:
Fixes #983.

**Changes proposed in this pull request:**
- Update `jsonschema` requirement version to `>=3.0.1,4.0.0`
- Update default JSON Schema draft from `4` to `7`

**Related Pull Request(s):**
- https://github.com/p1c2u/openapi-spec-validator/pull/78

### DIFF
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -151,7 +151,7 @@ def schema_response_integer(valid):
     if valid == "valid":
         return 3
     else:
-        return 3.0
+        return 3.14
 
 
 def schema_response_number(valid):
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -20,7 +20,7 @@ class ResponseValidator(BaseDecorator):
         :type operation: Operation
         :type mimetype: str
         :param validator: Validator class that should be used to validate passed data
-                          against API schema. Default is jsonschema.Draft4Validator.
+                          against API schema. Default is jsonschema.Draft7Validator.
         :type validator: jsonschema.IValidator
         """
         self.operation = operation
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -5,12 +5,12 @@ import logging
 import sys
 
 import six
-from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
+from jsonschema import Draft7Validator, ValidationError, draft7_format_checker
 from werkzeug import FileStorage
 
 from ..exceptions import ExtraParameterProblem
 from ..http_facts import FORM_CONTENT_TYPES
-from ..json_schema import Draft4RequestValidator, Draft4ResponseValidator
+from ..json_schema import Draft7RequestValidator, Draft7ResponseValidator
 from ..problem import problem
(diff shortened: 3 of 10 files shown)


tests/fakeapi/hello.py
```diff
@@ -151,7 +151,7 @@ def schema_response_integer(valid):
     if valid == "valid":
         return 3
     else:
-        return 3.0
+        return 3.14
```
hjacobs: what's the background of this change?

danie1k: Seems that jsonschema draft 7 treats `3.0` as integer, because `3.0 == 3`.  
I havent search where/when such change has been made in jsonschema.

danie1k: Unit tests were failing without that update. And because it looks pretty resonable, I applied that.

dtkav:
the openapi spec is explicitly based on draft 4.

danie1k:
> the openapi spec is explicitly based on draft 4.

I didn't know that. What are the reasons for that?

dtkav:
It looks like my response wasn't quite right (sorry I was on mobile).
[OpenAPI 2.0 (fka swagger) is on draft-4](https://swagger.io/specification/v2/#data-types)
[OpenAPI 3.0 is on wright draft 00](https://swagger.io/specification/#data-types) (draft 5)

I don't have the details on _why_, but i do think it's important to follow the spec.

danie1k:
> I think swagger 2.0 needs to continue to use draft-4...

I assume that there might not be easy way of using two different jsonchema drafts simultaneously?

> but openapi 3.0 could be updated to draft-5.

[According to official JSON Schema website, Draft 5 should not be used](https://json-schema.org/draft-06/json-schema-release-notes.html#q-what-happened-to-draft-05), [so Draft 6 is natural choice thenx](https://json-schema.org/draft-06/json-schema-release-notes.html#q-what-are-the-changes-between-draft-04-and-draft-06), as a closest next version... But because [Draft 7 is is fully backwards-compatible with Draft 6](https://json-schema.org/draft-07/json-schema-release-notes.html), we should use it, IMO.

danie1k:
There might boe one more problem... The [p1c2u/openapi-spec-validator](https://github.com/p1c2u/openapi-spec-validator/) package, which is the dependency of Connexion. It looks a bit abandoned :thinking: and it still rely on Draft 4.

danie1k:
Such update is far more complex than it looks, more details in #983.

tests/fakeapi/hello.py
danie1k: @hjacobs I've found more detailed explanation:
* https://github.com/Julian/jsonschema/issues/575#issuecomment-513679876
* https://json-schema.org/latest/json-schema-core.html#rfc.section.4.2.3

PR #997, opened 2019-07-12
Add validation of cookie parameters.

fwkz:
Right now, based on following definition, parameters passed as cookies will not be validated.
```yml
    name: token
    in: cookie
    description: token to be passed as a cookie
    required: true
    schema:
      type: array
      items:
        type: string
```

Proposed change validate parameters stored in cookies in similar manner that we validate parameters passed in headers.


hjacobs:
:+1:

PR #1000, opened 2019-07-15
Fixes #510: Multiple file uploads via POST now work correctly.

### LINKED ISSUE #510: supporting multiple file upload
### Description

When I define a swagger path like that one:

```
paths:
  /importer:
    post:
      operationId: api.importer.startImport
      description: |
        Start import process.
      consumes:
        - multipart/form-data
      parameters:
        - name: assets
          in: formData
          type: array
          items:
            type: string
            format: binary
```

I am not able to get the "array" of files defined under the name assets. 
Only the first file which is send by the name assets...

`def startImport(xml,importMeta,assets = None) -> object:
    print(assets)`
Result is : `<FileStorage: '25d' ('application/octet-stream')>`

### Expected behaviour

Array of Files should be accessible in the function.

### Actual behaviour

Only one file is send to the function

### Steps to reproduce
Sending 2 File with the key "assets" 

### Additional info:

Output of the commands:

Python 3.6.2
Connexion Version: 1.1.14

simondrabble:

 Fix issue #510

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -176,7 +176,7 @@ class AbstractOperation(SecureOperation):
 
     @staticmethod
     def _get_file_arguments(files, arguments, has_kwargs=False):
-        return {k: v for k, v in files.items() if k in arguments or has_kwargs}
+        return {k: files.getlist(k) for k in files.keys() if k in arguments or has_kwargs}
 
     @abc.abstractmethod
     def _get_val_from_param(self, value, query_defn):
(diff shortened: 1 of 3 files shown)


hjacobs:
Thanks!

hjacobs:
:+1:

ddurham2:
I believe this change breaks the aiohttp api. aiohttp passes a dict and not a wz.datatypes.ImmutableMultiDict as flask does.

dtkav:
Hey @ddurham2 good catch. We have a lot of problems with aiohttp / flask compatibility since the API is not well defined and tests aren't shared between the two frameworks.
#1095 is underway to fix some of the inconsistency in responses, but it's likely we'll have to make some major fixes to ConnexionRequest consistency as well.

PR #1004, opened 2019-07-28
Fix: not handling file parameters properly

saleepeppe:
Connexion is not handling properly parameters of type "file" in multipart/form routes.

I attach a simple application to test the bug:
[testing_connexion.zip](https://github.com/zalando/connexion/files/3439265/testing_connexion.zip)

The bug is reproducible with the current master branch both on win and unix on a python 3.7 environment.

It was not easy to figure out what was the problem and I am not sure this is the best way of fixing it.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -144,7 +144,7 @@ class RequestBodyValidator(object):
                     return error
             elif self.consumes[0] in FORM_CONTENT_TYPES:
                 data = dict(request.form.items()) or (request.body if len(request.body) > 0 else {})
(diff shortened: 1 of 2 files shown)


TimConvect:
I think this is a duplicate of:
https://github.com/zalando/connexion/pull/1000

dtkav:
hi @saleepeppe - I'm missing some information in order to be able to help.
I can't tell what you think is not working properly.
If you could post a failing test case, that would be very helpful.

RobbeSneyders:
Closing due to lack of response

PR #1010, opened 2019-08-01
Test schema corrections

patrickkwang:
This includes assorted fixes to bring the "simple" test schema in line with OpenAPI 3.0.

These issues were identified by attempting to validate the schema against [the OpenAPI schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json) using [jsonschema](https://github.com/Julian/jsonschema).

### DIFF
--- a/tests/fixtures/simple/openapi.yaml
+++ b/tests/fixtures/simple/openapi.yaml
@@ -424,7 +424,7 @@ paths:
         - name: items
           in: query
           description: An comma separated array of items
-          style: csv
+          style: form
           explode: true
           schema:
             type: array
--- a/tests/fixtures/simple/openapi.yaml
+++ b/tests/fixtures/simple/openapi.yaml
@@ -406,7 +406,8 @@ paths:
         - name: items
           in: query
           description: An comma separated array of items
-          style: simple
+          style: form
+          explode: false
           schema:
             type: array
             default: ["squash", "banana"]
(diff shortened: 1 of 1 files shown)


tests/fixtures/simple/openapi.yaml
```diff
@@ -1021,10 +1039,10 @@ components:
       required:
         - image_version
   parameters:
-    OptionalName:
+    Name:
       name: name
       in: path
       description: Name of the person to greet.
-      required: false
```
dtkav: All of the other changes make sense to me (adding descriptions, clarifying style and explode, etc.
But this is a change in the actual content and intent of the endpoint. Can you say more about why you did this?

dtkav:
Thanks for doing this. I'm excited to be able to use the OpenAPI schema for validation.
I left a minor comment, otherwise looking good!

tests/fixtures/simple/openapi.yaml
patrickkwang: Per the OpenAPI specification, path parameters _must_ have `required: true`. The nominal purpose of this test is to "check if add_operation is called only on http methods field", which does not obviously require that the name parameter by optional, so it seemed reasonable to make the name required. But perhaps there is some deeper purpose that I'm missing.

hjacobs:
Could you rebase this PR?

dtkav:
Hey @patrickkwang - If you give me write permissions to the PR I can resolve the conflict.
I'd love to get this and #936 through.

tests/fixtures/simple/openapi.yaml
dtkav: ah yep - I missed that it was a path param. :+1:

hjacobs:
:+1:

PR #1017, opened 2019-08-13
Make schema validation throw an exception

### LINKED ISSUE #1016: Schema validation errors not caught by error handler
### Description

Schema validation errors are not caught by error handlers.
I want my application to have a "standard" JSON error format.
For this i register an error handler to format the error response.
Schema validation errors bypass the error handler and do not allow to format the response.

### Expected behaviour

```
HTTP 500
{ "error": 'My custom error format" }
```

### Actual behaviour

```
HTTP 400
{
  "detail": "1000 is greater than the maximum of 90",
  "status": 400,
  "title": "Validation error",
  "type": "about:blank"
}
```

### Steps to reproduce

application.py:
``` python
connexion_app = connexion.App(root_path, specification_dir='src/api/rest')

def handle_connexion_exception(exception):
        return '{ "error": 'My custom error format" }', 500

connexion_app.add_error_handler(ProblemException, handle_connexion_exception)
```

Cause a openAPI schema error (call an endpoint with invalid JSON).

### Additional info:

The culprit seems to be `validate_schema` https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L188 that does not throw an error and just returns the endpoint response.

NMFR:
This will allow users to intercept the error and format its response.
Also change the schema validation title to "Validation error" so the response is easier to read.

Fixes #1016

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -8,7 +8,7 @@ import six
 from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
 from werkzeug import FileStorage
 
-from ..exceptions import ExtraParameterProblem
+from ..exceptions import ExtraParameterProblem, ProblemException
 from ..http_facts import FORM_CONTENT_TYPES
 from ..json_schema import Draft4RequestValidator, Draft4ResponseValidator
 from ..problem import problem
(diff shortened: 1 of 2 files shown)


NMFR:
One of the tests is failing but i think its not the PRs fault, seems to fail to download some zip.
Can i get a confirmation on this please or did i break something?

dtkav:
There were significant changes to exception handling in #955 
Those changes do not rename 'BadRequest' -> 'ValidationError', but I think they fix some of the problems you were having.

NMFR:
Yeah, fixed elsewhere. closing.

PR #1019, opened 2019-08-14
Using string parameters when uploading a multipart/form-data file fails

sherbang:
So far I only have a failing unit test for this problem.

It appears that `connexion.operations.OpenAPIOperation._get_body_argument()` is dropping my string parameters, but I'm not sure if that's the correct place to fix this?

I'm not sure that I'll be able to any additional work into this myself, but wanted to at least shine a light on this issue before I move on.

### DIFF
--- a/tests/api/test_parameters.py
+++ b/tests/api/test_parameters.py
@@ -212,6 +212,21 @@ def test_formdata_file_upload(simple_app):
     assert response == {'filename.txt': 'file contents'}
 
 
+def test_formdata_file_upload_and_string(simple_app):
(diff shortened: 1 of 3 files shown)


sherbang:
Separate, but related issue.  File upload alone always fails when the api is created with `pythonic_params=True`.

It appears that the conversion from camelCase to snake_case isn't happening everywhere it needs to.

hjacobs:
Thanks for the test case!

dtkav:
(^ kicking travis)

PR #1029, opened 2019-08-30
Fix routing doc

callkalpa:


hjacobs:
I think the docs are correct as the resource is a collection.

PR #1046, opened 2019-09-25
OpenAPI 3.* has the body name 'requestBody' which Connexion resolves …

RichardBruskiewich:
…to the argument tag "request_body" so this line of code is not standards compliant, thus non-object JSON request bodies are not captured as arguments (on line 263) without the artificial use of an explicit  'x-body-name' with value 'request_body'

Fixes # .

 -
 -
 -

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -244,7 +244,7 @@ class OpenAPIOperation(AbstractOperation):
         return {}
 
     def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
-        x_body_name = self.body_schema.get('x-body-name', 'body')
+        x_body_name = self.body_schema.get('x-body-name', 'request_body')
         if is_nullable(self.body_schema) and is_null(body):
             return {x_body_name: None}
 


pbasista:
> this line of code is not standards compliant
Which standards do you refer to?

From what I understand, OpenAPI specification does not mandate which variable names can be used by the implementing frameworks.

RichardBruskiewich:
Hi @pbasista ,

See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#requestBodyObject

Note the use of the specification field tag "requestBody" for the POST operation.  The Connexion framework translates this into a "request_body" parameter to the generated handler. This is not recognized by the above code, thus, even though the POST body contents (e.g. JSON data value) makes its way to this level of the code, it is not captured and sent back as a parameter to the caller, thus, does not get propagated to the handler. When one imposes the artificial "x-body-name" == to "request_body", this "fixes" the problem; however, this begs the question: should the Connexion code not automatically default to this (as per the OpenAPI standard).

Try it yourself: create a small stub spec which has the requestBody POST body tag, generate the Python code then single step down to this level of the code where you see all of this happen!

pbasista:
> The Connexion framework translates this into a "request_body" parameter to the generated handler.

Which handler do you refer to? Connexion by default does not use the name `request_body` for that parameter. It uses the name `body`.

> should the Connexion code not automatically default to this (as per the OpenAPI standard).

The OpenAPI Specification, to the best of my knowledge, does not specify the names for the parameters used in web handlers.

In the last paragraph of you message, you refer to some kind of Python code generation. Can you explain what specifically you mean by that? Connexion itself does not generate any code. If you want to use it, you need to write the code yourself.

RichardBruskiewich:
Sorry.. yes... I fully understand your confusion now (and mine as well).  You are absolutely correct... Connexion doesn't do code generation. *mea culpa* for saying that. Thanks for continued kind patience with this issue.

Here's the actual scenario: perhaps you can help me figure out the point of intervention:

1) I am standardising on the OpenAPI 3.0 specification (published by https://www.openapis.org/). I am using a code generator related to the Swagger code generator (specifically, the independent spin-off of the SmartBear team, https://openapi-generator.tech/) to generate Python Flask code. It is actually the code generator which ties Connexion to Flask for this purpose, so when one follows the stack trace, it is Flask app.py calling the Connexion via the "EndOfRequestLifecycleDecorator in the Connexiondecorator.py, which ultimately hits the wrapper method of the parameter.py module, which attempts to process the request, including the reported capture of the *requestBody* contents.

All I know is that the request.body and request.json values come in the door intact at that point, but , the request_body gets set to the request_json, then when the kwargs.update() function is called with operations.get_arguments, then this latter method calls self._get_body_arguments. this final method fails to pic up the request body _unless_ the *x-body-name* is set to 'request_body'.

Quick question at this point: what is the *self.body_schema* assumed at this point? Is Connection assuming a Swagger 2.0 or OpenAPI 3.* or no particular web request encoding standard at this point? Which authority, in effect, has globally defined the *body_schema* attribute you are using at this point? Somehow, I don't imagine that it is the Connexion package developer(?). As a corollary, which authority dictates the existence and meaning of the "x-body-name" attribute, which defaults to value "body" in the Connexion code base?

Please do kindly let me know who I should contact to enforce the actual standard in play here. As far as I can tell, the OpenAPI code generators are just passing the OpenAPI specification through their code, translating "requestBody" into "request_body". Is Flask going to specifically be OpenAPI aware for this? I simply do not know.

dtkav:
Hi @RichardBruskiewich - the name "body" was chosen for connexion and is unrelated to the openapi specifciation. My guess is that your problems are related to code generation from a different project.

Even if 'request_body' were a better name to use, it would cause major breakage to change it now, so I'm going to close this PR.

Maybe follow up with openapi-generator?

PR #1050, opened 2019-10-09
Support jsonschema >= 3.0.0

### LINKED ISSUE #983: Support jsonschema3

neverpanic:
Fixes #983.

Other projects have started using jsonschema >= 3.0.0, and modern python packaging tools such as poetry now fail to install projects using both connexion and a different dependency if they depend on different versions of jsonschema.

Allow jsonschema versions >= 3.0.0 to avoid this problem. Add a version branch to avoid a warning that would otherwise be thrown by jsonschema when using the deprecated `types` parameter of the jsonschema.IValidator constructor.

 - setup.py: Support jsonschema >= 3.0.0
 - Avoid warning when using jsonschema >= 3.0

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -4,8 +4,10 @@ import functools
 import logging
 import sys
 
+import pkg_resources
 import six
 from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
+from jsonschema.validators import extend
 from werkzeug.datastructures import FileStorage
 
 from ..exceptions import ExtraParameterProblem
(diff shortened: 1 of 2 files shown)


hjacobs:
I'm not sure, but I remember in other PRs that there was potentially an issue with a new jsonschema version -- I wonder what prevented us from updating earlier?

vssteve:
FYI, I found the change to allow jsonschema>=3.0.0 prevented connexion 2.4.0 from running for me because I'm using Python 2.7 and this happened because jsonschema 3.x eventually brings in the more_itertools package which now requires on Python 3.5 or later. In my case the dependency order was connexion 2.4.0->jsonschema 3.1.1 ->zipp 0.6.0->more_itertools 7.2.0.

neverpanic:
@vssteve Nothing in this PR requires jsonschema > 3. You can just manually downgrade jsonschema to a version < 3 for the remaining two months that python 2.7 is still supported.

daisieh:
Is this update included in any releases of connexion? It looks like it's not included in the 2.7 release on PyPI...

PR #1052, opened 2019-10-11
add opentracing functionality

Heiss:
This request adds a tracing functionality to connexion. Currently, you can use it with jaeger-client-python, but it should work with every implementation of opentracing specification.

For simple usage, it adds an optional parameter _use_tracer_ for add_api method, which takes an object of an opentracing implementation (currently no type checking in code).

With this feature, every request creates a span with help of the given tracer and reports automatically to the corresponding tracing-plattform, which was specified by the user. If there is a span-id in the request-header, it will use it to continue a session over multiple microservices.

Todo:
- Requests should take the current span-id and add it to every request to another service. Maybe this have to make from the user itself?

Warning:
This fork has not been tested, if it runs under python 2.7.

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -11,6 +11,8 @@ from connexion.lifecycle import ConnexionRequest, ConnexionResponse
 from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
 from werkzeug.local import LocalProxy
 
+from ..decorators.TracerDecorator import TracerDecorator
+
 logger = logging.getLogger('connexion.apis.flask_api')
 
 
@@ -114,6 +116,7 @@ class FlaskApi(AbstractAPI):
         return self._internal_handlers
 
     @classmethod
+    @TracerDecorator
     def get_response(cls, response, mimetype=None, request=None):
         """Gets ConnexionResponse instance for the operation handler
         result. Status Code and Headers for response.  If only body
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -11,8 +11,6 @@ from connexion.lifecycle import ConnexionRequest, ConnexionResponse
 from connexion.utils import Jsonifier, is_json_mimetype, yamldumper
 from werkzeug.local import LocalProxy
 
-from ..decorators.TracerDecorator import TracerDecorator
-
 logger = logging.getLogger('connexion.apis.flask_api')
 
 
@@ -116,7 +114,6 @@ class FlaskApi(AbstractAPI):
         return self._internal_handlers
 
     @classmethod
-    @TracerDecorator
     def get_response(cls, response, mimetype=None, request=None):
         """Gets ConnexionResponse instance for the operation handler
         result. Status Code and Headers for response.  If only body
--- a/connexion/decorators/TracerDecorator.py
+++ b/connexion/decorators/TracerDecorator.py
@@ -5,7 +5,7 @@ logger = logging.getLogger('')
 
 
 def TracerDecorator(func):
-    def wrapper(cls, response, mimetype=None, request=None):
+    def wrapper(response, mimetype=None, request=None):
         tracer = get_tracer()
 
         # if tracer is configured, then start a span now
@@ -28,12 +28,12 @@ def TracerDecorator(func):
             else:
                 scope = tracer.start_span("TracerDecorator")
 
-            resp = func(cls, response, mimetype, request)
+            resp = func(response, mimetype, request)
 
             scope.log_kv({"response": response})
             scope.finish()
         else:
-            resp = func(cls, response, mimetype, request)
+            resp = func(response, mimetype, request)
 
         return resp
 
--- a/connexion/decorators/TracerDecorator.py
+++ b/connexion/decorators/TracerDecorator.py
@@ -8,17 +8,13 @@ def TracerDecorator(func):
     def wrapper(cls, response, mimetype=None, request=None):
         tracer = get_tracer()
         logger.debug("{}".format(tracer))
-        # if jaeger is configured, then start a span now
-        if not tracer:
-            resp = func(cls, response, mimetype, request)
-
-        else:
+        # if tracer is configured, then start a span now
+        if tracer:
             from opentracing.ext import tags
             from opentracing.propagation import Format
 
             # extract the context from request header to continue a session
             # taken from https://github.com/yurishkuro/opentracing-tutorial/tree/master/python/lesson03#extract-the-span-context-from-the-incoming-request-using-tracerextract
-
             if request:
                 span_ctx = tracer.extract(Format.HTTP_HEADERS, request.headers)
                 span_tags = {tags.SPAN_KIND: tags.SPAN_KIND_RPC_SERVER}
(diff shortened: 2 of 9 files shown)


hjacobs:
Thanks for the PR!

connexion/decorators/TracerDecorator.py
```diff
@@ -0,0 +1,40 @@
...
+                path = urlparse(request.url).path
+
+                scope = tracer.start_span(path + "_" + request.method, child_of=span_ctx, tags=span_tags)
+                scope.log_kv({"request": request})
+            else:
+                scope = tracer.start_span("TracerDecorator")
+
+            resp = func(cls, response, mimetype, request)
+
+            scope.log_kv({"response": response})
```
ancelotpinto: are we working under the assumption that these calls do not throw? as if they do we will fail a successful response? do we need a try-except-finally here?

connexion/apis/flask_api.py
```diff
@@ -114,6 +116,7 @@ def _handlers(self):
         return self._internal_handlers
 
     @classmethod
+    @TracerDecorator
```
ancelotpinto: is this specific to flask or will it work in general for other implementations too?

Heiss: It should work for other implementations aswell, because the tracer decorator doesn't take any flask specific parameter, but i didn't test it.

connexion/decorators/TracerDecorator.py
Heiss: The variable func holds the function: get_response from flask_api.py ([L#120](https://github.com/zalando/connexion/blob/13809da779dfddc474bcfe38541983c9a5aa2a5d/connexion/apis/flask_api.py#L120), see your other comment).
This function returns a ConnexionRequest-Object. So it doesn't throw an error in my understanding.

connexion/apis/flask_api.py
ancelotpinto: In that case, can we add it also to the other api implementations? aiohttp, etc?

Would be nice to know it works universally and not just for flask :)

kornicameister:
My 5 cents on it: is that it makes conexxion less usable for people that might not want to have opentracing. It is not also a rocket science to just access `Flask` or `aiohttp` application that lies beneath `connexion` application and hook up tracing there. 
It feels wrong, that's another thing. `connexion` job was, IMHO, always to bridge `OpenAPI` and Python REST world. `connexion` can't be a tool that bundles everything because it makes it less usable for people that just search for `API first framework` but whole bunch of things they consider non-essential.

To prove the point, ask yourself a question: how hard it is to implement OpenTracing over pure `Flask` application? I don't find it particularly hard and I worked with projects that had it done and `connexion` was in there.

connexion/apis/flask_api.py
Heiss: Done. :) It is now available for all implementations. I made a quick test for aiohttp and it works as expected.

Heiss:
> My 5 cents on it: is that it makes conexxion less usable for people that might not want to have opentracing. It is not also a rocket science to just access `Flask` or `aiohttp` application that lies beneath `connexion` application and hook up tracing there.
> It feels wrong, that's another thing. `connexion` job was, IMHO, always to bridge `OpenAPI` and Python REST world. `connexion` can't be a tool that bundles everything because it makes it less usable for people look for `API first framework` but find whole bunch of things they consider non-essential.
> 
> To prove the point, ask yourself a question: how hard it is to implement OpenTracing over pure `Flask` application? How many of people are using `RestyResolver` which is mentioned by the author as a approach that makes you super fast. So, if I would to use different resolution mechanism and `aiohttp` I can only mourn because `TraceDecorator` is not good enough there. I am sorry if that sounds harsh, I get it that it does, I get that writing frameworks is tough job. But I also get that sometimes packing a project with features just for it to have tons of them won't it make shine.
> 
> BTW. Last point is as personal as it could be. I find connexion very useful but I truly believe and negate adding features that are easy to implement on your own. Just think about it please and think if it is worth a price, adding a feature like that. Or perhaps, just perhaps, it makes more sense to invest in polishing up existing implementation.

Thank you for your comment. I got your point. I made this for my daily work, so that it reduce boilerplate code for the upcoming (micro-)services for my institution. The framework needs to be simple to use (simplicity beats performance). Connexion adds everything, i searched for (in except: Tracing-Support and Prometheus metrics, which will be the next thing).

You are right, it wasn't difficult to implement, but not all features have to be complicated to be useful. Sometimes the small things are the main reason to choose a framework over another. Everything is there, you only need to put it together.

In the current state, it will only be executed, if you configure a tracer at startup. It doesn't impact your work, if you don't. In my opinion, if you don't want to use something, you should not have to. But it is a possibility to make you feel better, because you know, that you can add a functionality with a simple change in your configuration.

I worked at this in my fork and i thought, it could be useful for other people. But if connexion wants to be a "lightweight" framework, it is also okay for me. Github was made for this use-case. :)

kornicameister:
@Heiss hey, I appreciate no hard feeling attitude ;-). Well, it's not my call as usual. I am just a next door guy who wanted to stir the discussion up. 

Anyway, as you noticed already ; it's more or less a philosophical question. I prefer things light and prefer to know how to extend things instead of just push features ;-). 

BTW. Why can't you implement your tracing as an internal library that you install next connexion? Wouldn't it work?

Heiss:
@kornicameister: Because i don't think about it. :) This should work, too. Thank you for your hint. Maybe this would be much easier then my approach to maintain my fork for every connexion-update.

Before i start to implement the internal library, i will figure out, how to include prometheus, so this could be a library for the use-case: "make connexion microservice-ready". Maybe this would be more flexible for everyone.

kornicameister:
Point is that microservices are just buzz words. They have their strength and can be used if needed. Sometimes, however, microservices just get in the way. Sometimes even tracing via opentracing might get in the way. You mentioned `opentracing`. That's fair enough, but what happens if I try to set up your idea on GCE? There's this `stackdrive` and I am trying, although lightly, to find an assoc between `strackdriver` and `opentracing` and am unable to. 

( FYI: I am not an expert, I might have messed up facts here :P so if I am wrong, please somebody, tell me :D )

All in all...tracing via connexion, to be trully universal ought to take into account not only the fact that there is `Flask` or `aiohttp` implementation here for `connexion`. It should also make it universal enough so that there's this `abstract` driver that `connexion` understands and know and then you implement tracing solutions on top of that via `plugins`. That has this important benefit to lift a pressure and responsibility from `connexion` developers ;-).

dtkav:
I agree with @kornicameister - while this would be cool functionality, it adds a real maintenance cost. We need to be careful about which "batteries" we include.

if it's easy to bolt on to the side, I'd say lets keep it separate.

Maybe the best way to share your contribution would be as an example, or writing a blog post or gist for how to use opentracing and connexion together.

Heiss:
Thank you for your suggestions. I made a small module for my use case (like @kornicameister suggested), where i extend the functionality of connexion my own way. For someone who wants to take a look at it: https://pypi.org/project/connexion-plus/

I understand, that this "battery" should be optional.

kornicameister:
@Heiss I suppose that an effort to make connexion pluggable wouldn't be in vain but it takes @hjacobs, one core maintainer I know of, to make such decisions. This is pretty high level stuff worth a minor or even major release IMHO. 

I'd not mind seeing that in connexion TBH ;-)

Issue #998, opened 2019-07-15
Introduce typings


kornicameister:
### Description

Long story shorts. I would like to see connexion having typings.
Reason is simple...the codebase is large and quite complicated which makes having a typed code a necessity needed to unveil the functionality behind lines of code.
Obviously speaking we would need to drop support for `2.7` which I voted for in #990 which is not that bad of an idea given that `2.7` is reaching EOL and it's been, I guess, about 10 years since `3.x` has been introduced to us.

### Expected behaviour

connexion has typed codebase

### Actual behaviour

connexion does not have typed codebase

kornicameister:
@hjacobs what do you say. I can help with making a typings for connexion if you want to.
But let's see if can drop `2.7` first. If so, I will do just that along with creating a plan to proceed with introducing typings.

rafaelcaricio:
We don't need to drop support for python 2.7 to start introducing typing in the Connexion codebase. https://mypy.readthedocs.io/en/latest/python2.html

kornicameister:
@rafaelcaricio yeah, it is not needed but `2.7` goes out of the picture in few months. No reason to stick around with it.

hjacobs:
I vote for dropping Python 2.7 support!

rafaelcaricio:
I vote for drop 2.7 support as well.

kornicameister:
@hjacobs @rafaelcaricio 

- https://github.com/zalando/connexion/pull/1058
- https://github.com/zalando/connexion/pull/1057

cclauss:
The five syntax errors mentioned in #1392 might be a useful place to start the gradual adoption of type hints.

RobbeSneyders:
Closing in favor of #1111

PR #1057, opened 2019-10-22
Clean up dropping Python 3.{4,5}

kornicameister:
This is a follow up for c8d8973c7e390a5e2071c7b494dd7bd52bd55ff1
(dropping Python 3.5). There were some leftovers in requirements for
Pythonv versions `connexion` no longer supports.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,8 +27,6 @@ install_requires = [
     'requests>=2.9.1',
     'six>=1.9',
     'inflection>=0.3.1',
-    'pathlib>=1.0.1; python_version < "3.4"',
-    'typing>=3.6.1; python_version < "3.6"',
     'openapi-spec-validator>=0.2.4',
 ]
 


kornicameister:
cc @hjacobs

jmcs:
:+1:

setup.py
```diff
@@ -27,8 +27,6 @@ def read_version(package):
     'requests>=2.9.1',
     'six>=1.9',
     'inflection>=0.3.1',
-    'pathlib>=1.0.1; python_version < "3.4"',
```
jmcs: This breaks python 2.7, so it will need a couple more months before it can go.

kornicameister: There's #1058 already in which is a partial fix for #998

hjacobs:
Waiting for #1058 to be ready.

hjacobs:
@kornicameister can you update this PR as we now merged the removal of 2.7? Thanks!

hjacobs:
:+1:

PR #1059, opened 2019-10-22
Import from collections.abc to support Python 3.8

sochi:

 - import from `collections.abc` to support Python 3.8
 - fallback to import from `collections` if `collections.abc` is not available


hjacobs:
Can you add a unit test? Python 3.8 is running on Travis and no test failed, so we should add one to capture this :smirk:

hjacobs:
:+1:

PR #1071, opened 2019-11-12
Fix unit tests on Windows

drewdogg:
These five unit tests were failing under Windows due to differences in OS behavior.
The three tests in tests/test_api.py were failing since the NamedTemporaryFile object cannot be opened more than one time on Windows.  
The two tests in tests/aiohttp/test_aiohttp_app.py were failing due to hard-coded '/' path separator.


hjacobs:
:+1:

PR #1076, opened 2019-11-21
+ADD: name parameter for add_api

alexpdr:
- Added the required 'name' parameter for add_api in the flask app to allow dynamic api adding. (Otherwise you would get an error as blueprints get initialised with an empty string as name.)


dtkav:
Can you say more about your use case? What do you mean by "dynamic api adding"?

RobbeSneyders:
Closing due to lack of response.

PR #1077, opened 2019-11-22
Pop apiKey from query parameters in security decorator to fix #470

### LINKED ISSUE #470: Security GET parameters are rejected by strict validation
### Description

If I have the following definition, connexion will reject all `/ping` requests with `Extra query parameter(s) api_key not in spec`.

```yaml
securityDefinitions:
  ApiKeyAuth:
    type: apiKey
    in: query
    name: api_key
paths:
  /ping:
    security:
      - ApiKeyAuth: []
    get:
      summary: Return a simple JSON payload for health check purposes
      responses:
        # …
```

### Expected behaviour

connexion should know that `api_key` is also a supported parameter name.

### Actual behaviour

connexion thinks `api_key` has no place to be there

### Steps to reproduce

1. Launch an app with the above Swagger file
2. Send a request to `/ping?api_key=1`

### Additional info:

Output of the commands:

- `python --version`: Python 2.7.13
- `pip show connexion | grep "^Version\:"`: Version: 1.1.9

svetlana-v:
Fixes:

Fixes Bad Request error when ApiKey authentication is used with the apiKey query parameter, and strict validation is enabled.
The corresponding open issue is https://github.com/zalando/connexion/issues/470.

- Pop apiKey from the query parameters before validation.

### DIFF
--- a/connexion/decorators/security.py
+++ b/connexion/decorators/security.py
@@ -251,8 +251,25 @@ def get_cookie_value(cookies, name):
 def verify_apikey(apikey_info_func, loc, name):
 
     def wrapper(request, required_scopes):
+
+        def _immutable_pop(_dict, key):
+            """
+            Pops the key from an immutable dict and returns the value that was popped,
+            and a new immutable dict without the popped key.
+            """
+            cls = type(_dict)
+            try:
+                _dict = _dict.to_dict(flat=False)
+                return _dict.pop(key)[0], cls(_dict)
+            except AttributeError:
+                _dict = dict(_dict.items())
+                return _dict.pop(key), cls(_dict)
+
         if loc == 'query':
-            apikey = request.query.get(name)
+            try:
+                apikey, request.query = _immutable_pop(request.query, name)
+            except KeyError:
+                apikey = None
         elif loc == 'header':
             apikey = request.headers.get(name)
         elif loc == 'cookie':
--- a/tests/fakeapi/hello.py
+++ b/tests/fakeapi/hello.py
@@ -218,6 +218,10 @@ def test_required_query_param():
     return ''
 
 
+def test_apikey_query_parameter_validation():
+    return ''
+
+
 def test_array_csv_query_param(items):
     return items
 
--- a/tests/fixtures/simple/openapi.yaml
+++ b/tests/fixtures/simple/openapi.yaml
@@ -340,6 +340,19 @@ paths:
       responses:
         '200':
           description: OK
+  /test_apikey_query_parameter_validation:
+    get:
+      operationId: fakeapi.hello.test_apikey_query_parameter_validation
+      parameters:
+        - name: name
+          in: query
+          schema:
+            type: string
+      security:
+        - api_key: []
+      responses:
+        '200':
+          description: OK
   /test_required_query_param:
     get:
       operationId: fakeapi.hello.test_required_query_param
@@ -1140,3 +1153,9 @@ components:
       required: false
       schema:
         type: string
+  securitySchemes:
+    api_key:
+      type: apiKey
+      name: apikey
+      in: query
+      x-apikeyInfoFunc: fakeapi.hello.apikey_info
--- a/tests/fixtures/simple/swagger.yaml
+++ b/tests/fixtures/simple/swagger.yaml
@@ -245,6 +245,19 @@ paths:
         200:
           description: OK
 
+  /test_apikey_query_parameter_validation:
+    get:
+      operationId: fakeapi.hello.test_apikey_query_parameter_validation
+      parameters:
+        - name: name
+          in: query
+          type: string
+      security:
+        - api_key: []
+      responses:
+        200:
+          description: OK
+
   /test_required_query_param:
     get:
       operationId: fakeapi.hello.test_required_query_param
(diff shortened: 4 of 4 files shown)


connexion/operations/abstract.py
```diff
@@ -445,3 +446,10 @@ def json_loads(self, data):
         :type data: bytes
         """
         return self.api.json_loads(data)
+
+    def _get_apikey_parameter(self):
+        security_methods = set().union(*self._security)
+        if 'ApiKeyAuth' in security_methods:
+            api_key_auth = deepcopy(self._security_schemes['ApiKeyAuth'])
```
dtkav: I don't think this key should be hardcoded.
AFAICT 'ApiKeyAuth'  is a user-supplied name for a securityScheme of type `apiKey`.
See this page: https://swagger.io/docs/specification/authentication/api-keys/
Where it calls 'ApiKeyAuth' an "arbitrary name for the security scheme" in the example.

svetlana-v: Thanks for the quick response. I've updated the code with the apiKey type check.

The implementation does not imply more than one apiKey in query security method for a path. The first found query apiKey will be added to the spec parameters. I also don't see a use case in using more than one query api keys for a path.

dtkav:
Your approach breaks through some major abstractions in connexion.

I wonder if there's a better way.

One approach would be to pop the apikey from the query params once it passes security (before validation). Do you think the apiKey is likely to be needed in the business logic of the application?

It might also just make sense for users to add the apiKey query param to the spec for their route.

This seems like a real ambiguity in the openapi3 spec - it might also be worth checking upstream to see what they had in mind.

svetlana-v:
No, indeed. I agree that apiKey is not needed there. I've updated the implementation to pop the api key from the query parameters.

I don't think that users should add the apiKey query parameter to the spec for the routes. Then this parameter would also be exposed in the swagger ui but basically would never be used. I think it is also logical to imply that the apiKey parameter will be in query when it is defined as a query param.

connexion/decorators/security.py
```diff
@@ -252,7 +253,12 @@ def verify_apikey(apikey_info_func, loc, name):
 
     def wrapper(request, required_scopes):
         if loc == 'query':
-            apikey = request.query.get(name)
+            if isinstance(request.query, ImmutableMultiDict):
```
dtkav: I wonder if there's a way to no have to be coupled to werkzeug here.
maybe something like
```
def _immutable_pop(d, kpop):
    """
    takes an immutable dict and the key to pop
    returns a new immutable dict without the popped key, and the value that was popped
    """
    popped = request.query.get(kpop)
    cls = type(d)
    ret = [(k, v) for k, v in d.items() if k != kpop]
    return cls(ret), popped
    
request.query, apikey = _immutable_pop(request.query, name)
```


# copy for immutable_
new = [(k, v) for k, v in request.query.items() if k != name]
request.query = ne
```

**Comment by dtkav** [comment]
Thanks for the update!
I'd love to get some more opinions from other users / contributors on this idea.

**Comment by cognifloyd** [comment]
A fascinating issue.

Yes. I think that removing any security-defined query parameters before validation is the right way to approach this.

File: connexion/decorators/security.py (hunk 2 above)
Comment (svetlana-v) [reply connexion/decorators/security.py]: I'm now thinking that this won't work with MultiDict objects as it will squash the same keys, e.g. ImmutableMultiDict([('a', '1'), ('a', 2), ('a', 3)]) -> ImmutableMultiDict([('a', '1')]). I have updated it with the same try / except handling as used in the uri parsing decorator.

File: connexion/utils.py [hunk 3]
```diff
@@ -245,3 +245,17 @@ def create_empty_dict_from_list(_list, _dict, _end_value):
         return {current_key: create_empty_dict_from_list(_list, _dict, _end_value)}
     else:
         return {current_key: _end_value}
+
+
+def _immutable_pop(_dict, key):
```
dtkav: This is only used once - I don't think it should be in utils.

connexion/decorators/security.py
```diff
@@ -252,7 +252,7 @@ def verify_apikey(apikey_info_func, loc, name):
 
     def wrapper(request, required_scopes):
```
dtkav: I would nest `_immutable_pop()` within `def wrapper` because it's only used here.
The reason to keep it as a function (rather than inlining) is just to clearly isolate and define what we're trying to do, and give the opportunity to include the docstring.

connexion/utils.py
dtkav: also FYI - the leading underscore is a convention that means that the function is private (not meant to be used outside of a class / module), so moving it to utils and keeping the underscore is confusing.

dtkav:
nice work - it's looking a lot tighter and cleaner.
please remove the function from utils (it's only used in one place) and then I think we're ready to go!

connexion/decorators/security.py
```diff
@@ -252,7 +252,7 @@ def verify_apikey(apikey_info_func, loc, name):
...
         if loc == 'query':
-            apikey = request.query.get(name)
+            apikey, request.query = _immutable_pop(request.query, name)
```
dtkav: I think you'll also need to handle the `KeyError` if `name` is not in `request.query`.

svetlana-v: Yes, indeed. I've added try / except here.

svetlana-v:
Thanks for the comments! The helper function is moved to the wrapper.

hjacobs:
:+1:

PR #1079, opened 2019-11-23
Keyerror fix for #1078

### LINKED ISSUE #1078: example response not working due to KeyError
### Description
I am using `operation.example_response()` in my custom resolver. I am getting `None` as my example, even though the example exists for given status code and content type. 

### Expected behaviour
`operation.example_response()` should return valid example when exists for status code and content type.

### Actual behaviour
`operation.example_response()` is returning None when exists for status code and content type.

### Steps to reproduce
Using [hello world yaml](https://github.com/zalando/connexion/blob/master/examples/openapi3/helloworld/openapi/helloworld-api.yaml) 

```python
import connexion
from connexion.mock import MockResolver

class CustomResolver(MockResolver):
    def mock_operation(self, operation, *args, **kwargs):
        response, code = operation.example_response() # Returns None, 200

app = connexion.FlaskApp(__name__, specification_dir='swagger/')
app.add_api('helloworld-api.yaml', resolver=CustomResolver(True))
app.run(port=8080)
```
When I do `http post  :8080/getting/dave` I get 
```sh
HTTP/1.0 200 OK
Content-Length: 0
Content-Type: text/plain; charset=utf-8
Date: Sat, 23 Nov 2019 09:44:16 GMT
Server: Werkzeug/0.16.0 Python/3.7.4

```

### Additional info:

Output of the commands:

- `python --version`
Python 3.7.4

Version: 2.4.0

codeasashu:
Added unittests

Fixes #1078 .

 - Status code can be in string

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -71,7 +71,7 @@ def deep_get(obj, keys):
         return obj
     try:
         return deep_get(obj[int(keys[0])], keys[1:])
-    except ValueError:
+    except (KeyError, ValueError):
         return deep_get(obj[keys[0]], keys[1:])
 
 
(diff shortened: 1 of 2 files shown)


dtkav:
I think this is fixed by #1088

codeasashu:
Yes, it does. Closing this

PR #1081, opened 2019-11-23
Added default status code

### LINKED ISSUE #1080: operation response schema and response example behaves differently
### Description
I am depening on `response_schema` and `example_response` extensively in my custom resolver. I am expecting them to give out schema and example for same operation in same fashion, but they are not.

Consider [hello world yaml](https://github.com/zalando/connexion/blob/master/examples/openapi3/helloworld/openapi/helloworld-api.yaml)

```python
from connexion.mock import MockResolver

class CustomResolver(MockResolver):
    def mock_operation(self, operation, *args, **kwargs):
        response, code = operation.example_response() # Returns "hello dave!", 200
        schema = operation.response_schema() #returns {} 
```

I can understand that we need to specify `status_code` and `content_type` for the schema, however https://github.com/zalando/connexion/blob/master/connexion/operations/abstract.py#L283 seems to be using default content type in case one fails to specify. Why then, are we not using same for `status_code`? Something like

```python
status_code = status_code or sorted(self._responses.keys())[0]
```

### Expected behaviour

`response_schema` should give out default appropriate schema found, even when `status_code` is not mentioned (increasing status code can be used to determine first matching schema).

### Actual behaviour
`response_schema` returns nothing when `status_code` is not mentioned. However, `example_response` does.

### Additional info:

Output of the commands:

- `python --version`
Python 3.7.4
Version: 2.4.0

codeasashu:
Added  content type to make it look consistent with other methods
Fixes #1080

 -
 -
 -

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -160,6 +160,8 @@ class OpenAPIOperation(AbstractOperation):
         return schema
 
     def response_schema(self, status_code=None, content_type=None):
+        content_type = content_type or self.get_mimetype()
+        status_code = status_code or sorted(self._responses.keys())[0]
         response_definition = self.response_definition(
             status_code, content_type
         )


hjacobs:
I'm not sure what the PR tries to do, but it definitely would need a test case.

codeasashu:
@hjacobs This fixes issue #1080 . Basically, I am depending on `response_schema`, which doesn't defaults to `status_code` (among the status codes in response). However, it does defaults to the `content-type` which feels non-uniform to me. 

This fix, allows `response_schema` to use the pool of status codes from response, defaulting to the lowest, unless passed specifically by user. 

I will try to add testcases this weekend.

PR #1084, opened 2019-11-29
Allow users to enable debug mode via environment variables

### LINKED ISSUE #1083: Can't use environment variables to enable flask debug mode
### Description

With Flask we can enable debug mode by using flask environment variables like explained here : https://flask.palletsprojects.com/en/1.1.x/config/#environment-and-debug-features

It seems that `AbstractApp` enforces `debug=False` (in its constructor params) so if we call `app.run()` without the `debug` param, `FlaskApp` is calling app.run with `debug=False`)

A simple way to fix this would be to use `debug=None` instead of `debug=False` but I don't know what is the impact for other app types.

### Expected behaviour

Using `FLASK_ENV=development` or `FLASK_DEBUG=True` should enable debug mode on Flask applications

### Actual behaviour

Using `FLASK_ENV=development`runs flask in development environment bug debug mode is off :

```
 * Serving Flask app "myapp" (lazy loading)
 * Environment: development
 * Debug mode: off
```

### Steps to reproduce

Don't use the `debug` parameter in `app.run()` call and set `FLASK_ENV` variable to `development`.

### Additional info:

Python's version: 3.7.1
Connexion's version: 2.4.0

cedricmessiant:
Fixes #1083 

 - use `None` as default value for `debug` parameter in `AbstractApp` to allow users to use flask environment variables to activate debug mode

### DIFF
--- a/connexion/apps/abstract.py
+++ b/connexion/apps/abstract.py
@@ -13,7 +13,7 @@ logger = logging.getLogger('connexion.app')
 @six.add_metaclass(abc.ABCMeta)
 class AbstractApp(object):
     def __init__(self, import_name, api_cls, port=None, specification_dir='',
-                 host=None, server=None, arguments=None, auth_all_paths=False, debug=False,
+                 host=None, server=None, arguments=None, auth_all_paths=False, debug=None,
                  resolver=None, options=None, skip_error_handlers=False):
         """
         :param import_name: the name of the application package
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -1,0 +1,21 @@
+# coding: utf-8
+from flask import Flask
+
+from connexion.apps.flask_app import FlaskApp
+from connexion.apis.flask_api import FlaskApi
+
+
+def test_flask_app_default_params():
+    app = FlaskApp('MyApp')
+    assert app.import_name == 'MyApp'
+    assert app.server == 'flask'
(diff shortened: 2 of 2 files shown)


dtkav:
Does this change the default behavior?
When you pass `None` does flask default to `True` or `False`?

cedricmessiant:
Do you mean connexion's default behavior or flask's?
By default in flask, debug is set to `None` in `flask/app.py`. Then, it uses the configuration variables (`FLASK_ENV`  or `FLASK_DEBUG`) or the parameter that `run()` method takes (it defaults to `None` also)

hjacobs:
I think this change should have some comment as it's easy to forget why it is now `None` and not `False` --- also a unit test would be nice.

dtkav:
yeah @cedricmessiant , does it change the default connexion behavior? We don't want to accidentally cause everyone to run in debugg mode when they upgrade.

cedricmessiant:
@dtkav yeah, no problem when I run the `helloworld` example (no option passed to `run()`, it runs with debug mode off). 

@hjacobs I added a test (and put a comment about debug in the test, maybe we could add a link to this MR/issue but I don't think it's necessary)

hjacobs:
:+1:

cedricmessiant:
Thanks a lot !

PR #1055, opened 2019-10-16
Fixes util.deep_get when obj is a list

ornj:
The example below will seem strange. It's the result of distilling a complex
API spec down to the root cause of the issue.

```yaml
---
openapi: 3.0.0
info:
  version: v0
  title: Repo an error
  description: Extremely contrived, but it illustrates the problem
paths:
  /:
    get:
      operationId: app.get
      responses:
        '200':
          description: Just for giggles
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/SomeWeirdResource"
components:
  schemas:
    Resource:
      allOf:
        - type: object
          properties:
            id:
              type: string
    SomeWeirdResource:
      allOf:
        - type: object
          properties:
            id:
              $ref: '#/components/schemas/Resource/allOf/0/properties/id'
```

While convoluted, this is a valid OpenAPI 3 spec.

```sh
$ yarn swagger-cli validate openapi/my_api.yaml
yarn run v1.17.3
$ /[redacted]/node_modules/.bin/swagger-cli validate openapi/my_api.yaml
openapi/my_api.yaml is valid
✨  Done in 0.23s.
```

`utils.deep_get` assumes the obj to be a dictionary when it can also be a list.

Fixes #870 

Changes proposed in this pull request:
 - Attempt to cast key as int
 - Catch `ValueError` and try original behavior

PR #1088, opened 2019-11-29
Fixed mocking function not working with examples under numeric status…

### LINKED ISSUE #1087: Mocks not working with numeric status codes
### Description
Mock server doesn't return examples when numeric status codes are used in the yaml file, resulting always in "no example response was defined".

### Expected behaviour
Having defined an example response, I expect it to be returned when mocking an API.

### Actual behaviour
"No example response was defined"

### Steps to reproduce
1. Start connexion using attached file
```
connexion run --mock all esempio.yml
```
[esempio.yml.txt](https://github.com/zalando/connexion/files/3905870/esempio.yml.txt)
2. Try to get some `/pages`

### Additional info:
The reason lies in the code working only in presence of "default" key: there is a `int()` casting which is supposed to explode when applied on 'default' string, jumping the control in the `except` block where there is the correct string. 
But this breaks when there is no "default", but only numeric codes.

Output of the commands:

- `python --version`: Python 3.7.4
- `pip show connexion | grep "^Version\:"`: Version: 2.4.0

aijanai:
… codes (e.g., 200) , due to status code string being casted as integer and not raising a ValueError.

Included tests that break without the patch (it was testing for 'default' only, but not for numeric codes like '200')

Fixes #1087 .

 - Adding check on `examples` structure type: if it's a dictionary, don't try to cast to integer or it will always result in empty returned value.
 - Added tests that, without the patch, break when there is no "default" string.

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -69,9 +69,10 @@ def deep_get(obj, keys):
     """
     if not keys:
         return obj
-    try:
+
+    if isinstance(obj, list):
         return deep_get(obj[int(keys[0])], keys[1:])
-    except ValueError:
+    else:
         return deep_get(obj[keys[0]], keys[1:])
 
 
--- a/tests/test_mock.py
+++ b/tests/test_mock.py
@@ -2,11 +2,44 @@ from connexion.mock import MockResolver
 from connexion.operations import Swagger2Operation
 
 
-def test_mock_resolver():
+def test_mock_resolver_default():
     resolver = MockResolver(mock_all=True)
 
     responses = {
         'default': {
+            'examples': {
+                'application/json': {
+                    'foo': 'bar'
+                }
+            }
+        }
+    }
+
+    operation = Swagger2Operation(api=None,
+                                  method='GET',
+                                  path='endpoint',
+                                  path_parameters=[],
+                                  operation={
+                                      'responses': responses
+                                  },
+                                  app_produces=['application/json'],
+                                  app_consumes=['application/json'],
+                                  app_security=[],
+                                  security_definitions={},
+                                  definitions={},
+                                  parameter_definitions={},
+                                  resolver=resolver)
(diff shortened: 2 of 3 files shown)


connexion/utils.py
```diff
@@ -70,7 +70,11 @@ def deep_get(obj, keys):
     if not keys:
         return obj
     try:
-        return deep_get(obj[int(keys[0])], keys[1:])
+        if type(obj) == dict:
```
hjacobs: you can use `isinstance(obj, dict)`

aijanai: Is it more efficient?
I'm willing to understand

aijanai: Is it more efficient?
I'm willing to understand

dtkav:
This was introduced by #1055

connexion/utils.py
dtkav: https://stereochro.me/ideas/type-vs-isinstance/

connexion/utils.py
```diff
@@ -70,7 +70,11 @@ def deep_get(obj, keys):
...
+            # when this is a dict, int() casting will always result in an empty value
+            return deep_get(obj[keys[0]], keys[1:])
+        else:
+            return deep_get(obj[int(keys[0])], keys[1:])
     except ValueError:
         return deep_get(obj[keys[0]], keys[1:])
```
dtkav: This should no longer be needed.

connexion/utils.py
dtkav: I think it would be clearer to check for the unexpected case (list) and default to the expected case (dict).

aijanai: fixed

aijanai:
Push forced new changes (hope it is ok with the procedure) to acknowledge the remarks

hjacobs:
:+1:

dtkav:
nice work! thanks for the PR.

PR #1095, opened 2019-12-07
WIP Harmonize serialization between aiohttp and flask

cognifloyd:
Was part of #849. Please refer to the long chain of discussion there for background behind this PR.
Please note that many of the commits here come from #849. Once that is merged the remaining commits will be easier to review.

@dtkav and I discussed quite a bit about what the right interface would be for users so that it is not surprising when things are or are not serialized as json. But, harmonizing them ends up breaking tests for aiohttp, flask, or both depending on which set of assumptions we prefer.

So, we would like for the interface to be more sane and jsonify data for both text/plain and application/json or if mimetype is not defined. But we'd also like to prevent jsonification for mimetypes where that does not make sense. Eventually, `_serialize_data()` should become pluggable, possibly where a chain of serializers can vote on which mimetypes they can handle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,6 @@ htmlcov/
 *.swp
 .tox/
 .idea/
+.vscode/
 venv/
+src/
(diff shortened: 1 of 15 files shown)


RobbeSneyders:
Some of these changes have been merged as part of other PRs and others are no longer relevant after https://github.com/spec-first/connexion/pull/1491.

PR #1101, opened 2019-12-13
Revert "Fixes #510: Multiple file uploads via POST now work correctly."

### LINKED ISSUE #510: supporting multiple file upload

dtkav:
Reverts zalando/connexion#1000

This break aiohttp because getlist() is specific to a werkzeug datastructure.

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -176,7 +176,7 @@ class AbstractOperation(SecureOperation):
 
     @staticmethod
     def _get_file_arguments(files, arguments, has_kwargs=False):
-        return {k: files.getlist(k) for k in files.keys() if k in arguments or has_kwargs}
+        return {k: v for k, v in files.items() if k in arguments or has_kwargs}
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


hjacobs:
:+1:

PR #1102, opened 2019-12-13
Remove six (Python 2 support)

hjacobs:
Remove `six` usage, i.e. remove Python 2 support.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -29,7 +29,7 @@ class AbstractAPIMeta(abc.ABCMeta):
         cls._set_jsonifier()
 
 
-class AbstractAPI(object, metaclass=AbstractAPIMeta):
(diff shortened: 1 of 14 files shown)


hjacobs:
Still TODO: remove `six.reraise`.

connexion/apis/abstract.py
```diff
@@ -31,8 +29,7 @@ def __init__(cls, name, bases, attrs):
         cls._set_jsonifier()
 
 
-@six.add_metaclass(AbstractAPIMeta)
-class AbstractAPI(object):
+class AbstractAPI(object, metaclass=AbstractAPIMeta):
```
cognifloyd:
```suggestion
class AbstractAPI(metaclass=AbstractAPIMeta):
```
In python 3 all classes are newstyle classes, so we don't need `(object)` anywhere.

hjacobs:
:+1:

Issue #1108, opened 2019-12-18
pip install in py2.7 pulls in unsupported 2.5.0


spacether:
### Description

pip installing connection in a py2.7 env pulls in unsupported 2.5.0

### Expected behaviour

One of the older releases should be installed instead

### Actual behaviour

The newest release 2.5.0 is pulled in

### Steps to reproduce

```
virtualenv venv -p /usr/local/bin/python
source venv/bin/activate
pip install connexion
```
Inspect the terminal to see what version was installed and we see:
```
Collecting connexion
  Using cached https://files.pythonhosted.org/packages/fa/ce/e3f760c75176282ddb8c5600ed9849aa09873fbaf7f9503832ee8041c855/connexion-2.5.0-py2.py3-none-any.whl
```

### Additional info:

Output of the commands:

- `python --version` Python 2.7.15
- `pip show connexion | grep "^Version\:"` Version: 2.5.0

spacether:
This was resolved by the above merge

hjacobs:
Released: https://github.com/zalando/connexion/releases/tag/2.5.1

PR #1109, opened 2019-12-18
Don't allow connexion installation in py27

spacether:
Pip installing connection in a py2.7 env pulls in unsupported 2.5.0
This PR updates connextion to only allow installation on python >= 3.6

If merged, this PR should fix:
https://github.com/zalando/connexion/issues/1108


hjacobs:
Thanks!

hjacobs:
:+1:

PR #1110, opened 2019-12-19
scope deepObject defaults code to query params only 

### LINKED ISSUE #1082: Default values not dependent on _in type (query, form, path)
### Description

Strict validation is on.
Swagger yaml contains default value for a parameter with "in:  query". Validation always fails because the validation includes the default value in the form data as an extra unexpected parameter.

### Expected behaviour
Should work

### Actual behaviour
Does not work

### Steps to reproduce
query parameter with default value and strict validation enabled (this will check the input form data)

### Additional info:
```
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -133,14 +133,14 @@ class AbstractURIParser(BaseDecorator):
 
         # set defaults if values have not been set yet
-        resolved_param = self.set_default_values(resolved_param, self.param_schemas)
+        resolved_param = self.set_default_values(resolved_param, self.param_schemas, _in)
 
         return resolved_param
 
-    def set_default_values(self, _dict, _properties):
+    def set_default_values(self, _dict, _properties, _in):
         """set recursively default values in objects/dicts"""
         for p_id, property in _properties.items():
-            if 'default' in property and p_id not in _dict:
+            if 'default' in property and property['in'] == _in and p_id not in _dict:
                 _dict[p_id] = property['default']
             elif property.get('type', False) == 'object' and 'properties' in property:
                 _dict.setdefault(p_id, {})
```

### LINKED ISSUE #1092: Regression in 2.4.0: spurious ERRORs logged saying "Function argument not defined in specification"
### Description

Connexion produces spurious ERRORs that function arguments are not defined in schema, *even though they are.* These errors appear in connexion version `2.4.0` and do not in `2.3.0`. The errors appear for every request.

I find the error appears in cases when a formData argument coexists with query arguments that have a default value.

For example:

```
2019-12-05 10:47:00,123 connexion.operations.swagger2 ERROR    Function argument 'queryArg1' not defined in specification
2019-12-05 10:47:00,123 connexion.operations.swagger2 ERROR    Function argument 'queryArg2' not defined in specification
2019-12-05 10:47:00,123 connexion.operations.swagger2 ERROR    Function argument 'queryArg3' not defined in specification
```

Seems like it might be related to an old bug #522.

### Expected behaviour

No errors should be produced when arguments are defined in spec.

### Actual behaviour

Errors are produced.

### Steps to reproduce

```yaml
  ...
  /orgs/{org_id}/resource:

    post:
      operationId: "controllers.create_resource"
      consumes:
      - "multipart/form-data"
      produces:
      - "application/json"
      parameters:
      - name: "org_id"
        in: "path"
        description: ""
        required: true
        type: "integer"
        format: "int64"
      - name: "queryArg1"
        in: "query"
        description: ""
        enum: [all, one, two, three]
        default: "all"  # NOTE has default value
        type: "string"
      - name: "queryArg2"
        in: "query"
        description: ""
        default: false  # NOTE has default value
        type: "boolean"
      - in: "query"
        name: "queryArg3"
        type: "boolean"
        default: false  # NOTE has default value
        description: ""
      - in: "formData"  # NOTE in formData
        name: "upload_file"
        required: true
        type: "file"
        description: "The file to upload."
      responses:
        200:
          description: "successful operation"
        404:
          description: "Not found"

```

In python, we have something like this:

```python
def create_resource(org_id, queryArg1='all', queryArg2=False, queryArg3=False):
    return {"ok": True}
```

### Additional info:

Output of the commands:

- `python --version`

Python 3.7.5

Version: 2.4.0

dtkav:
Fixes spread of defaults into other parameters

Redo #971 
Fixes #1092 
Fixes #1082 

 - scope deepObject code to openapi3-only code blocks (defaults do not belong in uri_parsing, nor does deepObject logic make sense in an AbstractURIParser)
 - fix set_default_values(...) which put defaults from any parameter type into all other parameters.

Future:
- modify testing fixtures to catch setting request.form, request.files, request.body etc unintentionally

### DIFF
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -6,7 +6,7 @@ import re
 import json
 from .. import utils
 
-from ..utils import create_empty_dict_from_list
+from ..utils import deep_merge
 from .decorator import BaseDecorator
 
 logger = logging.getLogger('connexion.decorators.uri_parsing')
@@ -97,20 +97,6 @@ class AbstractURIParser(BaseDecorator, metaclass=abc.ABCMeta):
         """
         resolved_param = {}
         for k, values in params.items():
-            # extract the dict keys if specified with style: deepObject and explode: true
-            # according to https://swagger.io/docs/specification/serialization/#query
-            dict_keys = re.findall(r'\[(\w+)\]', k)
-            if dict_keys:
-                k = k.split("[", 1)[0]
-                param_defn = self.param_defns.get(k)
-                if param_defn and param_defn.get('style', None) == 'deepObject' and param_defn.get('explode', False):
-                    param_schema = self.param_schemas.get(k)
-                    if isinstance(values, list) and len(values) == 1 and param_schema['type'] != 'array':
-                        values = values[0]
-                    resolved_param.setdefault(k, {})
-                    resolved_param[k].update(create_empty_dict_from_list(dict_keys, {}, values))
-                    continue
-
             param_defn = self.param_defns.get(k)
             param_schema = self.param_schemas.get(k)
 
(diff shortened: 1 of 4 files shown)


dtkav:
blocked by #1125

dtkav:
@hjacobs this is a pretty nasty bug in the latest release. It would be great if you can have a look (after #1125)

dtkav:
@hjacobs I made the default param test use `strict_app` which catches the bug on master.

hjacobs:
:+1:

PR #1113, opened 2019-12-24
Fix: aiohttp access_log duplicated if passed as argument

pando85:
- Pop access_log option from aiohttp options
  - Add test to access_log options in aiohttp run


hjacobs:
:+1:

PR #1122, opened 2020-01-14
x-nullable $ref fix

hmajid2301:
Fixes # 1070.

 - Allow x-nullable to follow $ref (resolve) for v1

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -89,10 +89,10 @@ def extend_with_nullable_support(validator_class):
     def nullable_support(validator, properties, instance, schema):
         null_properties = {}
         for property_, subschema in six.iteritems(properties):
-            if "$ref" in subschema:
-                ref_schema = validator.resolver.resolve(subschema["$ref"])
-                if len(ref_schema) == 2:
-                    subschema = ref_schema[1]
+            try:
+                _, subschema = validator.resolver.resolve(subschema["$ref"])
+            except ValueError:
+                pass
 
             if isinstance(instance, collections.Iterable) and \
                     property_ in instance and \
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -89,6 +89,11 @@ def extend_with_nullable_support(validator_class):
     def nullable_support(validator, properties, instance, schema):
         null_properties = {}
         for property_, subschema in six.iteritems(properties):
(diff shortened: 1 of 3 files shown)


connexion/decorators/validation.py
```diff
@@ -89,6 +89,11 @@ def extend_with_nullable_support(validator_class):
     def nullable_support(validator, properties, instance, schema):
         null_properties = {}
         for property_, subschema in six.iteritems(properties):
+            if "$ref" in subschema:
+                ref_schema = validator.resolver.resolve(subschema["$ref"])
+                if len(ref_schema) == 2:
```
dtkav: what does this do?
can you remove the magic number / add a comment?

dtkav:
@hjacobs - what did you have in mind for the v1_unsupported branch?
Should it still pass CI? Same standards for code review?

hjacobs:
@dtkav I don't want to support v1, but at least we can have a branch for it so bugfixes have a "home" (and others might be able to look at them). Haven't decided on releases yet.

connexion/decorators/validation.py
hmajid2301: `ref_schema` returns a tuple like so like the $ref and the actual schema definition ("definitions/foo", {}). Also yeh I can do that if you want.

dtkav: I had done the review before realizing this was against v1_unsupported.
I don't have a strong preference.

Another way:
```
try:
    _, subschema = validator.resolver.resolve(subschema["$ref"])
except ValueError:
    pass
```

dtkav:
Ok, that makes sense. Then I suppose if this is a sort of "user contrib" branch, then we probably don't need to have the same standards of code review and passing CI.

It seems unlikely that we'll be able to get this to pass CI without a lot of work anyway, because it's probably still testing deprecated python versions, and package versions aren't pinned to known working versions.

Are you happy to merge this then?

connexion/decorators/validation.py
hmajid2301: Yes that is much cleaner

hmajid2301:
What's the best to install the latest v1 version then with these changes ?

hjacobs:
@hmajid2301 you can install directly from the git branch:

```
pip install git+https://github.com/zalando/connexion.git@v1_unsupported
```

PR #1124, opened 2020-01-14
Fixes #1020, OAS3 false positive for extra form param

### LINKED ISSUE #1020: Connexion claims extra formData parameters that are in the spec
### Description

I have a simple service that expects a POST to a certain endpoint.  There's 1 parameter in the requestBody of the POST ("customerId"), and also a header parameter ("Authorization").

If I turn on `strict_validation=True` and I send a valid request to the endpoint, I get: `Extra formData parameter(s) customerId not in spec`.

### Expected behaviour

The `customerId` parameter _is_ in the spec, so Connexion should accept the post and call the endpoint's business logic, rather than returning a 400.

### Actual behaviour

Returns:
```
{
    "detail": "Extra formData parameter(s) customerId not in spec",
    "status": 400,
    "title": null,
    "type": "about:blank"
}
```

### Steps to reproduce

I'm using this spec.yaml:
```
openapi: "3.0.0"

info:
  title: Hello World
  version: "2"
servers:
  - url: /v1.0

paths:
  /test:
    post:
      operationId: hello.test
      summary: A test endpoint.
      parameters:
        # If this header param is here and `strict_validation` is enabled, Connexion gives a 400 error,
        # even if you DO send `customerId`.
        - name: Authorization
          in: header
          required: true
          schema:
            type: string
      requestBody:
        content:
            schema:
              type: object
              properties:
                customerId:
                  type: string
              required:
                - customerId
      responses:
        "200":
          description: OK.
        "401":
          description: Unauthorized.
```

And this app.py:

```
import connexion


def create_app():
    cxn = connexion.FlaskApp(__name__.split('.')[0], specification_dir='.')
    cxn.add_api('spec.yaml', strict_validation=True, arguments={'title': 'Hello World Example'})
    return cxn.app
```

And hello.py is just:

```
def test(body) -> str:
    return f"customerId is {body['customerId']}"
```

Then, send a (valid) request to the endpoint.  I'm using the `httpie` utility, although you can exercise the bug with the SwaggerUI too.

`http -v -f POST http://localhost:8000/v1.0/test "Authorization: Bearer abcd" "customerId=4"`

You will see the 400 response mentioned above.

### Additional info:

It seems like having the header in the spec is necessary to make the bug happen.  If you take out the `parameters:` block, it works.

Output of the commands:

- `python --version`: 
Python 3.6.6

Version: 2.3.0

dtkav:
blocked by:
#1125 
#1110 

When using an OAS3 spec with formdata, the validation logic looks
for the key 'formData' in the spec parameters list. This keys is
specific to OAS2, and will never be present, causing any form data to
throw an ExtraParameterProblem.

Fixes #1020 

 - Don't fail strict validation if `formData` key is missing from the spec parameters (this is expected in OAS3)

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -302,7 +302,11 @@ class ParameterValidator(object):
 
     def validate_formdata_parameter_list(self, request):
         request_params = request.form.keys()
-        spec_params = [x['name'] for x in self.parameters.get('formData', [])]
+        try:
+            spec_params = [x['name'] for x in self.parameters['formData']]
+        except KeyError:
+            # OAS 3
+            return set()
         return validate_parameter_list(request_params, spec_params)
 
     def validate_query_parameter(self, param, request):
--- a/tests/api/test_parameters.py
+++ b/tests/api/test_parameters.py
@@ -117,6 +117,16 @@ def test_strict_extra_query_param(strict_app):
     assert response['detail'] == "Extra query parameter(s) extra_parameter not in spec"
 
 
+def test_strict_formdata_param(strict_app):
+    app_client = strict_app.app.test_client()
+    headers = {'Content-type': 'application/x-www-form-urlencoded'}
+    url = '/v1.0/test_array_csv_form_param'
+    resp = app_client.post(url, headers=headers, data={"items":"mango"})
+    response = json.loads(resp.data.decode('utf-8', 'replace'))
+    assert response == ['mango']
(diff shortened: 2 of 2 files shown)


hjacobs:
:+1:

standag:
This cannot work with `defaultdict` here https://github.com/zalando/connexion/pull/1124/files#diff-e07625137dcac1e41acb635a9d80f538R243

PR #1125, opened 2020-01-15
Fixes Build: Allow new datetime behavior in PyYAML 5.3

dtkav:
Fixes Master, which is broken due to PyYAML change in 5.3
https://github.com/yaml/pyyaml/pull/163

 - update test to not expect `Zulu` datetime ending, as PyYAML 5.3 parses the date as TZ aware


dtkav:
@hjacobs master is broken due to a change in PyYAML ( https://github.com/yaml/pyyaml/pull/163 ).
This change seems reasonable to me (loading a spec with a datetime+tz will make a tz-aware datetime).
Alternatively we could pin 5.2, or 5.3 and choose one behavior. This PR accepts both.

hjacobs:
:+1:

PR #1127, opened 2020-01-15
fix pythonic path param types in aiohttp

### LINKED ISSUE #1116: pythonic_params do not convert the parameter type
### Description

The parameter type is not converted if the parameter is renamed to become pythonic.

I have `pythonic_params=True` and the following spec:

```yaml
/reposet/{id}:
    delete:
      operationId: delete_reposet
      parameters:
      - explode: false
        in: path
        name: id
        required: true
        schema:
          type: integer
        style: simple
```

This spec defines a single `integer` parameter in the path. 
There is the following controller:

```python
async def delete_reposet(request: web.Request, id_: int) -> web.Response:
    print(type(id_))
    return web.Response(status=200)
```

### Expected behavior

The controller prints `int` for each request.

### Actual behavior

The controller prints `str` for each request. In other words, the type is not converted. If I rename `id` to `id_` everywhere in the spec, the type is converted and the controller prints `int` as expected. If I remove `pythonic_params` and rename `id_` to `id` in the controller, it prints `int`, too.

### Steps to reproduce

(already covered by the description)

### Additional info:

Output of the commands:

- `python --version` `3.7.5`
- `pip show connexion | grep "^Version\:"` `2.4.0`

dtkav:
Fixes #1116

### DIFF
--- a/tests/aiohttp/test_aiohttp_datetime.py
+++ b/tests/aiohttp/test_aiohttp_datetime.py
@@ -27,7 +27,10 @@ def test_swagger_json(aiohttp_api_spec_dir, aiohttp_client):
         return data
 
     example = get_value(spec_data, 'paths./datetime.get.responses.200.content.application/json.schema.example.value')
-    assert example == '2000-01-23T04:56:07.000008Z'
+    assert example in [
+        '2000-01-23T04:56:07.000008+00:00',  # PyYAML 5.3
+        '2000-01-23T04:56:07.000008Z'
+    ]
     example = get_value(spec_data, 'paths./date.get.responses.200.content.application/json.schema.example.value')
     assert example == '2000-01-23'
     example = get_value(spec_data, 'paths./uuid.get.responses.200.content.application/json.schema.example.value')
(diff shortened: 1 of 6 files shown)


dtkav:
note, this is already rebased ontop of #1125 
#1125 should be merged first.

tests/aiohttp/test_aiohttp_datetime.py
```diff
@@ -27,7 +27,10 @@ def get_value(data, path):
         return data
 
     example = get_value(spec_data, 'paths./datetime.get.responses.200.content.application/json.schema.example.value')
-    assert example == '2000-01-23T04:56:07.000008Z'
+    assert example in [
```
dtkav: from #1125

tests/test_flask_encoder.py
```diff
@@ -63,7 +63,10 @@ def get_value(data, path):
         return data
 
     example = get_value(spec_data, 'paths./datetime.get.{}.example.value'.format(response_path))
-    assert example == '2000-01-23T04:56:07.000008Z'
+    assert example in [
```
dtkav: from #1125

hjacobs:
:+1:

PR #1130, opened 2020-01-21
match all but whitespaces inside object keys

### LINKED ISSUE #1129: Request parameters with `object` schema is not parsed with special characters
### Description

While parsing a request with `object` schema and sending its parameters with special characters, it doesn't get parsed and returns None type. 

### Expected behaviour
If such query is requested: 
[GET] "http://localhost:80/path/1/queries?sort[so$me&sort.order]=desc" 
we expect to get parsed argument as: `sort=so$me&sort.order`

### Actual behaviour

"http://localhost:80/path/1/queries?sort[so$me&sort.order]=desc" - we get `sort=None`

### Steps to reproduce
Swagger template has the reference of sort object as follows:
```
paths:
  /query/{id}:
    get:
      operationId: "app.controller.query.function"
      - name: sort
        required: false
        in: query
        style: deepObject
        explode: true
        schema:
          type: object
          additionalProperties:
            type: string
      responses:
        200:
          description: "successful"
          content:
            application/json:
              schema:
                type: object
                properties:
                  query_id:
                    type: string
```

### Additional info:

Output of the commands:

- `python --version` = Python 3.7.4
- `pip show connexion | grep "^Version\:"` =  2.5.1

drummerwolli:
Fixes #1129  .

 - inside of a exploded `deepObject`, the keys can be any character. Especially dots need to be allowed. sanitation does not apply, because it will result in an object key, not a python var name.

### DIFF
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -188,7 +188,7 @@ class OpenAPIURIParser(AbstractURIParser):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\S[^\[\]]+)\]', k)
+        key_path = re.findall(r'\[(\S[^\[\]]*)\]', k)
         root = prev = node = {}
         for k in key_path:
             node[k] = {}
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -188,7 +188,7 @@ class OpenAPIURIParser(AbstractURIParser):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(.[^\[\]]*)\]', k)
+        key_path = re.findall(r'\[([^\[\]]*)\]', k)
         root = prev = node = {}
         for k in key_path:
             node[k] = {}
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -188,7 +188,7 @@ class OpenAPIURIParser(AbstractURIParser):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\S+)\]', k)
+        key_path = re.findall(r'\[(\S[^\[\]]+)\]', k)
         root = prev = node = {}
         for k in key_path:
             node[k] = {}
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -188,7 +188,7 @@ class OpenAPIURIParser(AbstractURIParser):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\S[^\[\]]*)\]', k)
+        key_path = re.findall(r'\[(.[^\[\]]*)\]', k)
         root = prev = node = {}
         for k in key_path:
             node[k] = {}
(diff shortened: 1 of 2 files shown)


dneuhaeuser-zalando:
:+1:

nolar:
👍

connexion/decorators/uri_parsing.py
```diff
@@ -188,7 +188,7 @@ def _make_deep_object(k, v):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\w+)\]', k)
+        key_path = re.findall(r'\[(\S[^\[\]]+)\]', k)
```
nolar: Will it work with 1-char values? It looks here like 1 non-spaces, followed by 1+ other chars, giving minimum 2 in total. It was min 1 before the change.

nolar:
```
In [3]: re.match(r'\w+', 'hello')                                                                                     
Out[3]: <re.Match object; span=(0, 5), match='hello'>

In [4]: re.match(r'\w+', 'h')                                                                                         
Out[4]: <re.Match object; span=(0, 1), match='h'>

In [8]: re.match(r'(\S[^\[\]]+)', 'he')                                                                               
Out[8]: <re.Match object; span=(0, 2), match='he'>

In [9]: re.match(r'(\S[^\[\]]+)', 'h')                                                                                

In [10]:                                                   
```

drummerwolli: let me check

drummerwolli: adjusted

connexion/decorators/uri_parsing.py
```diff
@@ -188,7 +188,7 @@ def _make_deep_object(k, v):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\w+)\]', k)
+        key_path = re.findall(r'\[(\S[^\[\]]*)\]', k)
```
nolar: Also, this changes the logic in some cases. For example:

```
In [14]: re.match(r'\[(\w+)\]', '[[x]]')                                                                              

In [15]: re.match(r'\[(\S[^\[\]]*)\]', '[[x]]')                                                                       
Out[15]: <re.Match object; span=(0, 4), match='[[x]'>

In [16]: re.match(r'\[(\S[^\[\]]*)\]', '[[x]]').group(1)                                                              
Out[16]: '[x'
```

I assume that either `"x"` or `"[x]"` is intended to be a key in this case, not `"[x"`.

nolar:
👍

connexion/decorators/uri_parsing.py
nolar: Similar with `[     ]` — a space, many spaces, or any space-starting string cannot be a key — is it intended?

drummerwolli: spaces are fixed. I agree that the other case is not intended, but i'm lacking time/interest to invest the time to make the regex bullet proof for that now.

connexion/decorators/uri_parsing.py
```diff
@@ -188,7 +188,7 @@ def _make_deep_object(k, v):
         root_key = k.split("[", 1)[0]
         if k == root_key:
             return (k, v, False)
-        key_path = re.findall(r'\[(\w+)\]', k)
+        key_path = re.findall(r'\[(.[^\[\]]*)\]', k)
```
hjacobs: Why do you keep a first dot (".") in there (any character)?

nolar: So, if it was "1+ word-letters-only" before, I'd suggest to follow the same rule as "1+ of anything that is not `[]`".

 ```suggestion
        key_path = re.findall(r'\[([^\[\]]+)\]', k)
```

Nested `[]` can be sacrificed for now — it is quite non-trivial to implement opening-closing bracket matching in regexp without an extra parser.

hjacobs:
:+1:

PR #1133, opened 2020-01-22
Pluggable converter for input parameters based on property format

eyalkaspi:
The open API spec states that

> Tools can use the format to validate the input or to map the value to a specific type in the chosen programming language. Tools that do not support a specific format may default back to the type alone, as if the format is not specified.

In this pull review, we are adding a pluggable mechanism to convert input values for query parameters or request body (openapi API, not implemented for swagger) to be converted to custom types based on the format property.

The motivating example is the ability to convert inputs with format "date" to python's datetime.date and input with format "date-time" to python's "datetime.datetime". This PR does not include any custom converters in itself and is thus backwards compatible as it doesn't affect the behavior of applications which don't define custom converters.

A future change might add converters for widely used format specifiers (date, date-time, uri, ...), but given the complexity around date/date-time validation (jsonschema soft dependency on GPL libraries), simply providing the pluggable mechanism solves 99% of the challenge for our use case (without this pull request, we have to monkey patch connexion or use decorators/before_request to intercept the input to the API handler).

At this point, the review contains the code change and unit tests. If the maintainers accept the idea of this pull request, I'll include documentation update.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -37,7 +37,8 @@ class AbstractAPI(metaclass=AbstractAPIMeta):
     def __init__(self, specification, base_path=None, arguments=None,
                  validate_responses=False, strict_validation=False, resolver=None,
                  auth_all_paths=False, debug=False, resolver_error_handler=None,
-                 validator_map=None, pythonic_params=False, pass_context_arg_name=None, options=None):
+                 validator_map=None, format_converters=None,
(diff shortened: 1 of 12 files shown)


anis-campos:
Hi, can we have this reviewed ? This is a no-brainier, without it, we end-up having to parse each and every date fields, a real bug-prone nightmare !

RobbeSneyders:
Closing as this is based on outdated code. Happy to review an updated PR on this.

PR #1138, opened 2020-01-24
fix nested additionalProperties

### LINKED ISSUE #1137: v2.4+ silently removes additionalProperties from nested objects
### Description
This change introduced new behavior that silently removes nested (i.e. not top-level) properties from request bodies when they aren't explicitly declared in the `properties` block, even if `additionalProperties` is `true` (implicitly or explicitly): https://github.com/zalando/connexion/commit/54e50f22787c814caf52f45521991da02ff5f84f#diff-476ae7e600ebfcb9bae2b8199fdd669bR325

### Expected behaviour
All object properties are included in the sanitized object unless its schema explicitly sets `additionalProperties: false`

### Actual behaviour
Nested object properties are removed from the request body if a `properties` block is present in the object's schema unless they are explicitly declared in the object's `properties` block.

### Steps to reproduce
yaml:
```
components:
  schemas:
    FooBar:
      type: object
      additionalProperties: true
      properties:
        foo:
          type: string

paths:
  /foo:
    post:
      x-openapi-router-controller: app.controllers.foo
      operationId: foo
      requestBody:
        content:
          application/json:
            schema:
              type: object
              additionalProperties: true
              properties:
                bar: $ref: '#/components/schemas/FooBar'
```
call:
```
curl -XPOST -d'{
  "bar": {"foo": "foo", "qux": "qux"},
  "baz": "whatever"
}' /foo
```
controller:
```
def foo(body: dict):
    assert 'bar' in body         # no problem
    assert 'baz' in body         # additionalProperty also fine at top level
    assert 'foo' in body['bar']  # chill chill chill, everything as expected
    assert 'qux' in body['bar']  # NOPE! WHERE MY DATA AT THO?!
```

### Additional info:

Output of the commands:

- `python --version`
```
Python 3.7.5+
```
```
Version: 2.4.0
```

dtkav:
Fixes #1137 
Redo #971

 - Don't drop nested additionalProperties
 - Remove unnecessary filtering of objects in OAS3 (if it passed validation, it can be used as-is)

### DIFF
--- a/tests/api/test_responses.py
+++ b/tests/api/test_responses.py
@@ -142,7 +142,7 @@ def test_empty(simple_app):
 def test_exploded_deep_object_param_endpoint_openapi_simple(simple_openapi_app):
     app_client = simple_openapi_app.app.test_client()
 
-    response = app_client.get('/v1.0/exploded-deep-object-param?id[foo]=bar&id[foofoo]=barbar')  # type: flask.Response
+    response = app_client.get('/v1.0/exploded-deep-object-param?id[foo]=bar')  # type: flask.Response
     assert response.status_code == 200
     response_data = json.loads(response.data.decode('utf-8', 'replace'))
     assert response_data == {'foo': 'bar', 'foo4': 'blubb'}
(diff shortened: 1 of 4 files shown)


tests/api/test_responses.py
```diff
@@ -181,7 +188,7 @@ def test_nested_exploded_deep_object_param_endpoint_openapi(simple_openapi_app):
     response = app_client.get('/v1.0/nested-exploded-deep-object-param?id[foo][foo2]=bar&id[foofoo]=barbar')  # type: flask.Response
     assert response.status_code == 200
     response_data = json.loads(response.data.decode('utf-8', 'replace'))
-    assert response_data == {'foo': {'foo2': 'bar', 'foo3': 'blubb'}, 'foofoo': 'barbar'}
+    assert response_data == {'foo': {'foo2': 'bar', 'foo3': 'blubb'}, 'foo4': 'blubb', 'foofoo': 'barbar'}
```
dtkav: hey @drummerwolli - note the change here that defaults in the spec are included.
I think this is the right behavior (and consistent with how we do this elsewhere).

hjacobs:
Thanks @dtkav!

hjacobs:
:+1:

PR #1140, opened 2020-01-25
WIP: Parse dont validate

dtkav:
 - move type coercion into one place (with validation)
 - move sanitization to just before populating function arguments to avoid over-matching (like "$abc" and "abc$")
 - refactor + simplify notorious `_get_body_argument()` function.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -74,6 +74,8 @@ def coerce_type(param, value, parameter_type, parameter_name=None):
                 for k, v in d.items():
                     if k in schema['properties']:
                         d[k] = cast_leaves(v, schema['properties'][k])
+                    elif schema.get('additionalProperites'):
+                        d[k] = cast_leaves(v, schema['additionalProperties'])
                 return d
             return cast_leaves(value, param_schema)
         return value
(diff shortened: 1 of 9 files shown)


connexion/decorators/validation.py
```diff
@@ -71,15 +70,13 @@ def make_type(value, type_literal):
...
                 if type(d) is not dict:
-                    try:
-                        return make_type(d, schema['type'])
-                    except (ValueError, TypeError):
-                        return d
+                    return coerce_type(schema, d, parameter_type)
                 for k, v in d.items():
                     if k in schema['properties']:
                         d[k] = cast_leaves(v, schema['properties'][k])
+                    elif schema.get('additionalProperites'):
```
cognifloyd: how does this handle an explicit `additionalProperties: true`?
It looks like it calls `cast_leaves` -> `coerce_type` so `param` ends up as a True bool, so `param.get("schema", param)` (line 54) should fail.
If `additionalProperties` is not defined or None (ie the default value of True applies), then this skips handling it.
```suggestion
                    elif schema.get('additionalProperites') and not isinstance(schema.get('additionalProperties'), bool):
```
But that doesn't handle cases where `additionalProperties` are just allowed, and no schema for them is defined.

ThomasBendaSEL: also spelling - 'additionalPrope**rites**' --> 'additionalPrope**rties**'

RobbeSneyders:
This split is now done as part of https://github.com/spec-first/connexion/issues/1489.

PR #1150, opened 2020-02-10
Use `body` for DELETEs

### LINKED ISSUE #517: Attempting to run a DELETE request with an empty, non-required body returns a 400
### Description
When defining a DELETE endpoint, I have an optional body parameter of type array to provide a list of keys to delete.  If I provide that list, the endpoint will delete just those keys (this works fine).

What I would like to do is to make the body parameter on the delete request optional, such that if I don't provide it, I expect the endpoint to delete all its elements.

### Expected behaviour
When the DELETE request is called with no body, Connexion passes the body parameter as None to the endpoint handler function.

### Actual behaviour
Connexion returns a 400 error:
```
{
  "detail": "None is not of type 'array'",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}
```

### Steps to reproduce
Define a DELETE endpoint such as:
```yaml
paths:
  /items/:
    delete:
      summary: Delete a list of items or all of them
      description: |
        Given a list of IDs the end point will delete them. If however provided
        an empty list, the end point will delete all items
      operationId: delete_items
      produces:
         - application/json
         - application/problem+json
      parameters:
        - $ref: "#/parameters/items_keys"
      responses:
        200:
          description: Items deleted.
          examples:
            application/json:
              "\"Items Deleted\""

parameters:
  items_keys:
    name: forDelete
    description: IDs for deletion
    in: body
    required: false
    schema:
      type: array
      items:
        $ref: "#/definitions/key"

definitions:
  key:
    additionalProperties: false
    type: string
```

And an endpoint:
```python
def delete_items(forDelete):
    # Do some stuff here using either a list or a None type for forDelete    
    pass
```

### Additional info:

Output of the commands:

- `python --version`
```
Python 3.5.2
```
```
Version: 1.1.9
```

rail:
According to
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE, the
`DELETE` method may use a request with a body.

Fixes #517

 - Do not ignore the "DELETE" method while converting the request body into function args/kwargs.

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -267,7 +267,7 @@ class AbstractOperation(SecureOperation, metaclass=abc.ABCMeta):
         ret.update(self._get_query_arguments(query_params, arguments,
                                              has_kwargs, sanitize))
 
-        if self.method.upper() in ["PATCH", "POST", "PUT"]:
+        if self.method.upper() in ["DELETE", "PATCH", "POST", "PUT"]:
             ret.update(self._get_body_argument(body, arguments,
                                                has_kwargs, sanitize))
             ret.update(self._get_file_arguments(files, arguments, has_kwargs))


dtkav:
Unfortunately this violates the openapi specification. See #517 for more details.

rail:
Oh, I see. Thanks for the pointer.

PR #1158, opened 2020-02-17
Fix typo in security.rst from of to an or

qasimalbaqali:
Changed `of` to an `or`, I guessed a Dutch person wrote this and noticed it's one of the biggest gotchas to see if a Dutchie is the author.

Fixes # .
There is a type in the document where an or should replace the of.

 - Change "For example, an `x-tokenInfoFunc` of `auth.verifyToken` would pass..." to "For example, an `x-tokenInfoFunc` or `auth.verifyToken` would pass..."

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -11,7 +11,7 @@ With Connexion, the API security definition **must** include a
 ``x-tokenInfoFunc`` must contain a reference to a function
 used to obtain the token info. This reference should be a string using
 the same syntax that is used to connect an ``operationId`` to a Python
-function when routing. For example, an ``x-tokenInfoFunc`` or 
+function when routing. For example, an ``x-tokenInfoFunc`` with a value of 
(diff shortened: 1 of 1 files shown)


hjacobs:
I think "or" makes no sense here as the meaning is that "auth.verifyToken" is the property value of "x-tokenInfoFunc".

qasimalbaqali:
My guess was wrong then, my apologies.

hjacobs:
@qasimalbaqali we could change "of" to "with a value of" to make it more clear. WDYT?

qasimalbaqali:
I agree this makes more sense reading it like that.

hjacobs:
@qasimalbaqali can you update your PR accordingly? Thanks!

qasimalbaqali:
Changed accordingly

hjacobs:
:+1:

PR #1163, opened 2020-02-20
Enforcedefaults aiohttp

### LINKED ISSUE #1162: Body validator cannot change body itself when using AioHttpApp
### Description

[This](https://github.com/zalando/connexion/tree/master/examples/swagger2/enforcedefaults) doesn't work for the request body when using aiohttp and openapi3. The `body` argument of a handler is not affected by the validator. When using with OpenAPI 3.0, there is AFAIK no concept like `in: body` parameters, so it is quite a problem because there is no way to force the default values inside the body JSON.

### Expected behaviour

See https://github.com/p4l1ly/connexion/tree/enforcedefaults_aiohttp/examples/openapi3/enforcedefaults_aiohttp

Expected response is `{"foo": "foo"}`

### Actual behaviour

When the [example](https://github.com/p4l1ly/connexion/tree/enforcedefaults_aiohttp/examples/openapi3/enforcedefaults_aiohttp) is run using the current `zalando/master` branch, actual response is `{}`

### Steps to reproduce

Install connexion with `swagger-ui` support, run the [example](https://github.com/p4l1ly/connexion/tree/enforcedefaults_aiohttp/examples/openapi3/enforcedefaults_aiohttp)

Issue `curl -X POST localhost:8080/api/echo -H 'Content-Type: application/json' -d '{}'`

### Additional info:

Output of the commands:

- `python --version` `Python 3.7.6`
- `pip show connexion | grep "^Version\:"` `Version: 2.6.0`

p4l1ly:
Fixes #1162 .

 - Lazy load the `ConnexionRequest.json` property.
 - Example added.


connexion/lifecycle.py
```diff
@@ -24,7 +23,9 @@ def __init__(self,
 
     @property
     def json(self):
-        return self.json_getter()
+        if not hasattr(self, '_json'):
```
hjacobs: I think this change is unrelated to your example, why did you add it?

p4l1ly: It is related. Actually it fixes the issue #1162. As opposed to FlaskApp, [this code](https://github.com/zalando/connexion/blob/master/connexion/apis/aiohttp_api.py#L318)  says that the body is recomputed each time we access `ConnexionRequest.json`. So if a validator changes the json, it is not reflected in the `body` argument because [here](https://github.com/zalando/connexion/blob/master/connexion/decorators/parameter.py#L90) it is recomputed again.

p4l1ly: If you don't believe, please try the example without this change and see that it behaves differently (the `foo` attribute is not added with the default value).

RobbeSneyders (approved):
Thx @p4l1ly!

PR #1170, opened 2020-03-04
openapi: remove JSON body second validation and type casting

### LINKED ISSUE #691: `oneOf` does not work on `requestBody` when `oneOf` is nested under `#/components/schemas`

p4l1ly:
JSON body is already validated using jsonschema. There was also some type
casting but it was wrong: e.g. not recurring deeply into dicts and lists,
relying on existence of "type" in schema (which is not true e.g. if
oneOf is used). Anyway, the only reason why types should be casted is
converting integer values to float if the type is number. But this is in
most cases irrelevant.

Added an example, which did not work before this commit (echoed `{}`)
e.g. for
```
curl localhost:8080/api/foo -H 'content-type: application/json' -d '{"foo": 1}'
```
but now the example works (echoes `{"foo": 1}`).

Fixes #691 

From  `OpenAPIOperation._get_body_argument`:
- remove additional validation which is already done via jsonschema (including `additionalProperties` handling)
- remove type casting which didn't work either and it is irrelevant as the [specification](https://swagger.io/docs/specification/data-models/data-types/#numbers) says that the `number` type is for both integers and floats so we don't need to forcibly convert int to float (and there's really nothing else to convert in the json-parsed body).


p4l1ly:
I forgot to mention that there was also some not-fully-working default value handling (which worked only for top-most entity). I've removed that too. Currently, the only working default value handling in body is AFAIK using [this approach](https://github.com/zalando/connexion/tree/96bdcb010a4e5180c2bce18295e7b576c9c1ef0f/examples/swagger2/enforcedefaults) and if using aiohttp, also [this fix](https://github.com/zalando/connexion/pull/1163) (which has unfortunately not yet been merged) is needed.

p4l1ly:
ok, removing the default value handling breaks tests. I was not thinking about form bodies etc.

p4l1ly:
To increase coverage, I have removed some edge case checking, for the edge cases that I could not achieve in tests. Please check it.

dtkav:
You might find #1140 interesting. I was doing similar work, but haven't had any free time to finish it up.

p4l1ly:
I've found #1140 interesting, but it covers quite a big part of the connexion's code and I'm afraid I will not find time to dive into it in a foreseeable future. I'm not sure if #1140 handles the JSON body correctly, maybe I could transform the example in this pull request into test so that it could be added into #1140?

RobbeSneyders:
Hi @p4l1ly 
Happy to review if you can rebase on and retarget to the `v2` branch. We can later cherry pick it to `main` for v3.

p4l1ly:
Hi @RobbeSneyders , thanks for your response. I've rebased, retargeted and slightly refactored the pull request.

connexion/operations/openapi.py
```diff
@@ -302,25 +318,11 @@ def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
...
         body_arg = deepcopy(default_body)
         body_arg.update(body or {})
 
-        res = {}
         if body_props or additional_props:
-            res = self._get_typed_body_values(body_arg, body_props, additional_props)
-
-        if x_body_name in arguments or has_kwargs:
-            return {x_body_name: res}
+            return self._get_typed_body_values(body_arg, body_props, additional_props)
```
RobbeSneyders, line 325: Why do we need this for form types and not json?

RobbeSneyders, line 325: Why do we need this for form types and not json?

RobbeSneyders (changes requested):
Thanks @p4l1ly 

Could you convert the example into a test?

connexion/operations/openapi.py
p4l1ly: Because for json, strings are already converted to appropriate python types via `json.loads`. But there's no such conversion for forms, [only string splitting of arrays](https://github.com/spec-first/connexion/blob/84e33e5897a6a43f2a06ee5193ea958386252ce2/connexion/decorators/uri_parsing.py#L181). So I have left forms to be processed the old way and focused on fixing json, which is simpler to fix and more important (`oneOf` is rare in form data).

In my opinion, proper solution would perform type conversions somehow via jsonschema validators, at one place, close to the start of the request handling pipeline. Currently, types are converted inconsistently [here](https://github.com/spec-first/connexion/blob/84e33e5897a6a43f2a06ee5193ea958386252ce2/connexion/decorators/validation.py#L59) and [there](https://github.com/spec-first/connexion/blob/84e33e5897a6a43f2a06ee5193ea958386252ce2/connexion/decorators/uri_parsing.py#L179) and [there](https://github.com/spec-first/connexion/blob/84e33e5897a6a43f2a06ee5193ea958386252ce2/connexion/operations/openapi.py#L326) and [there](https://github.com/spec-first/connexion/blob/84e33e5897a6a43f2a06ee5193ea958386252ce2/connexion/operations/swagger2.py#L309). But please, this was intended to be a quick simple bugfix MR, let's not dive into big refactorings here.

p4l1ly:
@RobbeSneyders, I've just converted the example into a test.

RobbeSneyders (approved):
Thanks @p4l1ly.

@Ruwann I suggest to release this as 2.14.0 since we no longer cast numbers to float.

Ruwann:
Thanks for the efforts!

Indeed, let's release as 2.14.0 :)

PR #1171, opened 2020-03-04
Setting default values in JSON bodies is the default behaviour

p4l1ly:

 - JSON bodies' defaults could have been enforced using [this approach](https://github.com/zalando/connexion/tree/96bdcb010a4e5180c2bce18295e7b576c9c1ef0f/examples/swagger2/enforcedefaults). But why won't we make it implicit? Without any external "monkey patching"?

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -1,4 +1,5 @@
 from copy import deepcopy
+import six
 
 from jsonschema import Draft4Validator, RefResolver, _utils
 from jsonschema.exceptions import RefResolutionError, ValidationError  # noqa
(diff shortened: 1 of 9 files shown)


carloscbl:
Hey, i just did manually this, and it is passing default values into PATCH request, the ones where it should be disabled, or i'm wrong?
you get body param filled with defaults instead only the flask request.data

p4l1ly:
@carloscbl thank you for reviewing. Why do we want to ignore filling defaults in PATCH? As for your second note, I was focusing primarily on openapi3 and aiohttp, so there may be something wrong for flask. I've however added some tests and they're passing. Could you please specify in more detail how did you tested it and what was wrong?

p4l1ly:
Yes, I agree that it could be useful to see both the original and the modified body. But isn't that a more general problem (for a new issue)? For example, if you have query parameters with defaults, you get them in the handler with the defaults already filled in. AFAIK (correct me if I'm wrong) there's no mechanism to check, which of the parameters were provided from default values and which of them were provided by the client. Then, at this point of development, why should the behavior for bodies be different? If one doesn't want to automatically fill the defaults, he will just not provide them in the schema.

A mechanism for getting the original body however still exists (at least for `aiohttp`, I'm not very familiar with flask):
```
#!/usr/bin/env python3

import connexion


async def echo(body, request):
    return {'with_defaults': body, 'original': await request.json()}


if __name__ == '__main__':
    app = connexion.AioHttpApp(
        __name__, port=8080, specification_dir='.')
    app.add_api(
        'enforcedefaults-api.yaml',
        pass_context_arg_name="request",
    )
    app.run()
```

carloscbl:
I agree with what you say, and thanks i did not know about "pass_context_arg_name"
that will really help.

In other more important topic, today i found a breaking problem with ReadOnly spec

This was solved changing
DefaultsEnforcingDraft4Validator = extend_with_set_default(Draft4RequestValidator)
to
DefaultsEnforcingDraftValidator = extend_with_set_default(jsonschema.Draft7Validator)

and 
*args, validator=DefaultsEnforcingDraftValidator, **kwargs)

p4l1ly:
Hmm. I'm OK with advancing to Draft 7. However, [here](https://swagger.io/docs/specification/data-models/keywords/) I can see that OpenAPI 3 specification refers to JSON schema Draft 5 (which is named Draft4 in the jsonschema library, as Draft5 is just a cleanup of Draft4). Anyway, I've tried to change every occurence of Draft4 to Draft7 in the code and the tests fail. But this issue is about something else.

RobbeSneyders:
We've opted not to make this the default behavior, as this has some downsides, but we now offer a `DefaultsJsonRequestBodyValidator` which makes it easier to enable this. https://github.com/spec-first/connexion/pull/1616

PR #1179, opened 2020-03-13
expose swagger-ui index template variables to app configuration

### LINKED ISSUE #1178: expose swagger-ui index.j2 template arguments
### Context

Currently, connexion allows users to customize the swagger ui application by setting the `swagger_path` option, which should point to a bundle of files which will be mounted as the ui application.
In the absence of an explicitly set value, https://github.com/dtkav/swagger_ui_bundle is used.
Connexion expects to find an `index.j2` template file in this bundle which it renders internally with jinja2:
- [flask application](https://github.com/zalando/connexion/blob/c08111b6db136f0e9d24449fc0e3576ebf863c55/connexion/apis/flask_api.py#L285)
- [aiohttp application](https://github.com/zalando/connexion/blob/c08111b6db136f0e9d24449fc0e3576ebf863c55/connexion/apis/aiohttp_api.py#L234)

### Issue description

There is no clean way to use additional custom variables in the template file, because the rendering itself happens within connexion code.
It's also not flask/aiohhtp agnostic, so it's difficult to override with dependency injection.

#### Example use case

We have two deployments of the same application setup - one for production use, one as developmental sandbox. We would like to customize the visuals on the sandbox swagger-ui so it's distinct from the prod version, so we avoid damaging prod data by mistake when testing.

Right now the simplest solution is to have our own swagger ui bundle for sandbox applications which has a customized stylesheet, and to decide whether to use normal bundle or customized bundle during application start, based on which environment it's running in.

This is not entirely unreasonable solution, but it doesn't offer a lot of flexibility since the stylesheet is hardcoded in the bundle.

### Solution proposal

Give the users of connexion a way to pass additional options to the `index.j2` template.

#### Effect on our use case

We'd have a single customized swagger ui bundle with some additional styling options in `index.j2`, which we could distribute among our services (both aiohttp and flask).
The developers of each service could decide what to pass into the template during the application launch.

Please let me know what you think about this proposal, or if there is some better way to do this kind of thing.

msladecek:
Fixes #1178  .

- add an app option for variables which will be passed to swagger-ui index.j2 template

### DIFF
--- a/connexion/apis/aiohttp_api.py
+++ b/connexion/apis/aiohttp_api.py
@@ -237,7 +237,8 @@ class AioHttpApi(AbstractAPI):
     async def _get_swagger_ui_home(self, req):
         base_path = self._base_path_for_prefix(req)
         template_variables = {
-            'openapi_spec_url': (base_path + self.options.openapi_spec_path)
+            'openapi_spec_url': (base_path + self.options.openapi_spec_path),
+            **self.options.openapi_console_ui_index_template_variables,
         }
         if self.options.openapi_console_ui_config is not None:
             template_variables['configUrl'] = 'swagger-ui-config.json'
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -286,7 +286,8 @@ class InternalHandlers:
             prefix=escaped
         )
         template_variables = {
-            'openapi_spec_url': flask.url_for(openapi_json_route_name)
+            'openapi_spec_url': flask.url_for(openapi_json_route_name),
+            **self.options.openapi_console_ui_index_template_variables,
         }
         if self.options.openapi_console_ui_config is not None:
             template_variables['configUrl'] = 'swagger-ui-config.json'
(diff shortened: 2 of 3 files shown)


RobbeSneyders:
Thx for the PR @msladecek. Hopefully you're still interested in getting this merged :)

The changes look good to go. Before merging, would be great to just:
- add some info on this to the documentation
- rebase and retarget the PR to `main`, which is the new default branch

zhang-hongjie:
Hello, I have question, is it required that only the original contributor can continue the work for this merge request, please ? During 16+ months, maybe the original contributor don't want to continue work on it.  But this feauture is very interesting, it's better if we can integrate it.

RobbeSneyders:
Hi @zhang-hongjie the work in this PR was complete, so I rebased and retargeted it myself. I opened #1390 which should expose it in the documentation. Feel free to take up any open issues though.

RobbeSneyders (approved):

zhang-hongjie:
@RobbeSneyders  Thanks, very cool.

PR #1191, opened 2020-03-25
pass correct content_type to is_json_schema_compatible

### LINKED ISSUE #655: How to validate data if swagger-spec has consumer other than 'json' type
### Description
Not an issue more of a question of which i could not find a solution. 
Here's my swagger spec:

```json
"/endpoint1/": {
 "post": {
        "consumes": [
          "application/json",
          "multipart/form-data"
        ],
```

When i give my request for the following endpoint i can give input in 2 formats. For application/json type input i depend on connexion.decorator.validation to do the validation and for the 'multipart/form-data', i am currently doing the manual check which although same type of data but in a file.

### Expected behaviour
I expect that connexion recognises what type of request is it of the two. And if it is application/json it should validate rather than skipping it.

Is there a way to get the checks done?

### Actual behaviour
Skipping the validation for data type being application/json.

### Steps to reproduce
Just create a simple endpoint which consumes `application/json` and `multipart/form-data`.
Now pass the application/json data and you see that request body is not validated to the defined schema.

https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L101, For me this line is the problem, where multipart mimetype is rejected for validation.

### Additional info:

Output of the commands:

- `python --version` `Python 3.6.5`
- `pip show connexion | grep "^Version\:"` - `Version: 1.1.15`

### LINKED ISSUE #860: Content negotiation and response validation assumes incorrect schema for validation
### Description
In an API written in OAS v3 (which supports content negotiation), when an operation has two different response mimetypes _and_ one of them is not json-compatible, the validation assumes `application/json`. This entails some errors on the validation.

### Expected behaviour
The validation should honor the API specification and use the schema of the associated content type. One should be able to have an operation that can respond `application/json` and `application/octet-stream` according to the `Accept` header.

### Actual behaviour
When the operation initializes its mimetype and finds that the operation can produce more than one type, it assumes that it is `application/json`. However, if the implementation's response is not json or does, the validation will fail.

### Steps to reproduce
1. Create an API with an operation that can respond two different content types where one of the types is not json-compatible, like `application/json` and `application/content-type`. For example:
```
---
openapi: 3.0.2
info:
  title: example
  description: example
  version: 0.0.1
paths:
  /api/files/{id}:
    parameters:
      - name: id
        in: path
        description: File identifier
        required: true
        schema:
          type: string
    get:
      summary: Fetch file metadata or contents
      description: ...
      operationId: app.api.files.get
      responses:
        '200':
          description: File contents or metadata
          content:
            application/json:
              schema:
                type: object
            application/octet-stream:
              schema:
                type: string
                format: binary
```

2. Implement the operation while honoring the Accept header. For example:
```
import os

from flask import request, send_file


def get(*, id):
    # Content negotiation
    best = request.accept_mimetypes.best_match(['application/json',
                                                'application/octet-stream'],
                                               default=None)
    if best == 'application/json':
        return {'id': id, 'metadata': {'key': 'value'}}, 200

    elif best == 'application/octet-stream':
        tmp_file = tempfile.NamedTemporaryFile(mode='w+b')
        tmp_file.write(os.urandom(32)) # 32 random bytes
        tmp_file.flush()
        tmp_file.seek(0)

        response = send_file(tmp_file, mimetype='application/octet-stream')
        response.direct_passthrough = False
        return response, 200
```

3. Run connexion with response validation enabled.

4. Request the operation from a client using the Accept header to request
   `application/octet-stream`. In the example above, it will fail with a
   `UnicodeDecodeError: 'utf-8' codec can't decode byte` while trying to
   interpret the binary contents as json.

### Additional info:

After some analysis, I have identified some potential sources that produce this behavior.

First, it seems that
`connexion.operations.abstract.AbstrsactOperation.get_mimetype` incorrectly
assumes that the mimetype is `application/json` when the operation does not
produce json for all its responses (according to `connexion.utils.all_json`.
This makes the `ResponseValidator` have an incorrect mimetype field. Then,
`ResponseValidator.is_json_Schema_compatible` is confused by this and eventually
the `ResponseValidator.validate_response` calls `json.loads`, which fails and it
is interpreted as a validation error.

Output of the commands:

- `python --version`
Python 3.7.1

Version: 2.2.0

m-nikhil:
Fixes #655 
Fixes #860 
Related to pull request #760

 - Use correct content_type instead of self.mimetype in [is_json_schema_compatible](https://github.com/zalando/connexion/blob/96bdcb010a4e5180c2bce18295e7b576c9c1ef0f/connexion/decorators/response.py#L65) function in decorators/reponse.py

Problem Statement:
   - In the case of multiple content_types in spec, response validation assumes incorrect content_type and tries to json_load.

Expected Scenario:
- In the case of multiple content_types in spec, the content_type will be set in the response by the user; else the default application/json will be used. 

Solution:
-  In [is_json_schema_compatible](https://github.com/zalando/connexion/blob/96bdcb010a4e5180c2bce18295e7b576c9c1ef0f/connexion/decorators/response.py#L65) function, the logic rely on self.mimetype which defaults to application/json always. 
-  The above causes the function to always return True. 
- The is_json_schema_compatible must depend on the [content_type](https://github.com/zalando/connexion/blob/96bdcb010a4e5180c2bce18295e7b576c9c1ef0f/connexion/decorators/response.py#L39) set by the user. 
-  The changes made pass content_type to the function and operates the logic on content_type instead of self.mimetype, so returns the correct boolean.

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -46,7 +46,7 @@ class ResponseValidator(BaseDecorator):
         response_definition = self.operation.response_definition(str(status_code), content_type)
         response_schema = self.operation.response_schema(str(status_code), content_type)
 
-        if self.is_json_schema_compatible(response_schema):
+        if self.is_json_schema_compatible(response_schema,content_type):
             v = ResponseBodyValidator(response_schema, validator=self.validator)
             try:
                 data = self.operation.json_loads(data)
@@ -66,7 +66,7 @@ class ResponseValidator(BaseDecorator):
                 raise NonConformingResponseHeaders(message=msg)
         return True
 
-    def is_json_schema_compatible(self, response_schema: dict) -> bool:
+    def is_json_schema_compatible(self, response_schema: dict, content_type: str) -> bool:
         """
         Verify if the specified operation responses are JSON schema
         compatible.
@@ -77,7 +77,7 @@ class ResponseValidator(BaseDecorator):
         """
         if not response_schema:
             return False
-        return all_json([self.mimetype]) or self.mimetype == 'text/plain'
+        return all_json([content_type]) or content_type == 'text/plain'
 
     def __call__(self, function):
         """


m-nikhil:
@hjacobs @dtkav Can you guys please take a look.

uniqueg:
Would be very interested to have this merged, or have the issue of dealing with different content types addressed anyhow

vmarkovtsev:
Ported as https://github.com/athenianco/especifico/pull/10

m-nikhil:
Bumping it up!
Adding current maintainer,  @RobbeSneyders , @Ruwann

RobbeSneyders:
Thanks @m-nikhil.
Could you rebase on and target this to the `v2` branch? We'll cherrypick it to main for v3 later.

m-nikhil:
> Thanks @m-nikhil. Could you rebase on and target this to the `v2` branch? We'll cherrypick it to main for v3 later.
@RobbeSneyders , do let me know if it looks good.

RobbeSneyders:
Thanks @m-nikhil 

Two remarks:
- This PR doesn't actually fix #655 since that one is about incoming content-types, not response types. It should fix #1500 though. Could you update the description to reflect this?
- Could you add a test for the fixed behavior?

RobbeSneyders:
Outdated

Issue #1202, opened 2020-04-01
Inconsistent behaviour when a charset is defined in Content-Type header value with an invalid and valid json 


scoulomb:
### Description

Given following OpenApi definition

````
openapi: 3.0.1
info:
  title: Minimal project with connexion
  description: Sample project
  version: 1.0.0
servers:
  - url: /api/v1/test
paths:
  /entry:
    post:
      operationId: sample_code.endpoints.endpoint.post
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                kind:
                  type: string
                  description: 'Request current time or date'
                  enum:
                    - time
                    - date
              required:
                - kind
      responses:
        '201':
          description: Created
          content:
            application/json:
              schema:
                type: object
                properties:
                  theResponse:
                    type: string


````

Where I accept a `Content-Type: application/json`, with no charset
We found an inconsistent behaviour when a charset is defined in Content-Type header value with an invalid and valid json.

- (case 1) We can notice for valid json there is a tolerance to allow 'Content-Type: application/json; charset=utf-8'  even if not defined in the OpenApi file (is this expected?. Should we raise unsupported media type error since not defined?)
- (case 2) For invalid JSON 'Content-Type: application/json' is still accepted (as expected),
- (case 3) But for invalid json when the charset is present: an unsupported media type error is raised (not expected)

### Expected behaviour

 - I would expect case 3 to not raise unsupported media type error to be consistent with a valid json (case 1) when a charset is present
 - Alternatively we could remove the tolerance in case 1, if no charset is defined in OpenApi file, we could raise unsupported media type error  if Content-Type header contains a charset. In OpenApi it seems possible to explicitly declare a charset as shown here:  https://github.com/apiaryio/dredd/issues/1391

### Steps to reproduce

````
printf '(case 1) Tolerance to allow Content-Type: application/json; charset=utf-8 [EVEN IF NOT DEFINED IN OPENAPI?]\n-----\n'
curl --request POST --header 'Content-Type: application/json; charset=utf-8' http://server:8080/api/v1/test/entry --data '{"kind" : "time"}'

printf '(case 3) Invalid JSON Content-Type: application/json; charset=utf-8 is not accepted [NOT CONSISTENT WITH VALID JSON]\n-----\n'
curl --request POST --header 'Content-Type: application/json; charset=utf-8' http://server:8080/api/v1/test/entry --data '{"kind" "time"}'
````
Case 1 will output normal response
[...]

PR #1204, opened 2020-04-02
Fix inconsistent behaviour when a charset is defined in Content-Type header value with an invalid and valid json #1202

scoulomb:
# Fixes 
Issue https://github.com/zalando/connexion/issues/1202

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -131,8 +131,17 @@ def is_json_mimetype(mimetype):
     :type mimetype: str
     :rtype: bool
     """
-    maintype, subtype = mimetype.split('/')  # type: str, str
-    return maintype == 'application' and (subtype == 'json' or subtype.endswith('+json'))
+    simple_mime_type, _, optional_parameter = mimetype.partition(";")
+    maintype, _, subtype = simple_mime_type.partition("/")
+
+    is_valid_main_type = (maintype == 'application')
+    is_valid_subtype = (subtype == 'json' or subtype.endswith('+json'))
+    is_valid_optional_parameter = (
+            not optional_parameter or
+            optional_parameter.strip() == "charset=utf-8"
+    )
+
+    return is_valid_main_type and is_valid_subtype and is_valid_optional_parameter
 
 
 def all_json(mimetypes):
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -138,11 +138,7 @@ class RequestBodyValidator(object):
 
                 empty_body = not(request.body or request.form or request.files)
                 if data is None and not empty_body and not self.is_null_value_valid:
-                    try:
-                        ctype_is_json = is_json_mimetype(request.headers.get("Content-Type", ""))
-                    except ValueError:
-                        ctype_is_json = False
-
+                    ctype_is_json = is_json_mimetype(request.headers.get("Content-Type", ""))
                     if ctype_is_json:
                         # Content-Type is json but actual body was not parsed
                         raise BadRequestProblem(detail="Request body is not valid JSON")
--- a/tests/test_json_validation.py
+++ b/tests/test_json_validation.py
@@ -77,6 +77,38 @@ def test_writeonly(json_validation_spec_dir, spec):
     assert json.loads(res.data.decode())['title'] == 'Response body does not conform to specification'
 
 
+@pytest.mark.parametrize("spec", SPECS)
+def test_invalid_json(json_validation_spec_dir, spec):
+    app = build_app_from_fixture(json_validation_spec_dir, spec, validate_responses=True)
+    app_client = app.app.test_client()
+
+    # JSON Content-type, BadRequestProblem is raised
+    res = app_client.post('/v1.0/user', data="invalid json", content_type='application/json') # type: flask.Response
+    assert res.status_code == 400
+    assert json.loads(res.data.decode())['title'] == 'Bad Request'
+
+    res = app_client.post('/v1.0/user', data="invalid json", content_type='application/json; charset=utf-8') # type: flask.Response
+    assert res.status_code == 400
+    assert json.loads(res.data.decode())['title'] == 'Bad Request'
+
+    res = app_client.post('/v1.0/user', data="invalid json", content_type='application/anything+json; charset=utf-8') # type: flask.Response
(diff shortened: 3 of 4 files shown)


scoulomb:
I tested it locally here:
https://github.com/scoulomb/zalando_connexion_sample  (+buildLocalLib branch)

I did not see specific unit test to update, what would you advise?

connexion/utils.py
```diff
@@ -131,7 +131,9 @@ def is_json_mimetype(mimetype):
     :rtype: bool
     """
     maintype, subtype = mimetype.split('/')  # type: str, str
-    return maintype == 'application' and (subtype == 'json' or subtype.endswith('+json'))
+    return maintype == 'application' and (subtype == 'json'
+                                          or subtype == 'json; charset=utf-8'
+                                          or subtype.endswith('+json'))
```
Jiehong, line 136: Would using `subtype.containes('json')` be enough to better cover this?

scoulomb: If doing this we would accept invalid content-type? like `Content-type: application/json; charset=utf-83w9fmsefew0`, 

Also from https://stackoverflow.com/questions/9254891/what-does-content-type-application-json-charset-utf-8-really-mean and http://www.ietf.org/rfc/rfc4627.txt. 
We can see `Content-type: application/json; charset=utf-8 ` <=> `Content-type: application/json`, otherwise we would have ti support more encoding.

Jiehong: For the RFC4627, you didn't even highlight a specific passage, but I still checked and saw 2 things:

     JSON text SHALL be encoded in Unicode.  The default encoding is
     UTF-8.

and 


      JSON may be represented using UTF-8, UTF-16, or UTF-32.  When JSON
      is written in UTF-8, JSON is 8bit compatible.  When JSON is
      written in UTF-16 or UTF-32, the binary content-transfer-encoding
      must be used.

So it is indeed possible to not only send UTF-8.

But in our case, I agree it's fine that utf-8 is enough, as it's the default for now.

How would the code deal with `application/json;charset=utf-8`?

scoulomb: Based on our discussion, what do you think of this implementation?
```
# Reference is https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
# Where we can see MIME type can have optional parameter
# We can have a space or not in charset
def is_json_mimetype(mimetype):
    """
    :type mimetype: str
    :rtype: bool
    """
    simple_mime_type, optional_parameter = mimetype.split(";") if \
        ";" in mimetype else \
        (mimetype, None)
    maintype, subtype = simple_mime_type.split('/')  # type: str, str

    return maintype == 'application' and \
           (
                   subtype == 'json' or
                   subtype.endswith('+json')
           ) and \
           (
                   optional_parameter is None or
                   optional_parameter.strip() == "charset=utf-8"
           )


```

where

```
import unittest

from mimetype import is_json_mimetype


class TestUtils(unittest.TestCase):

    def test_is_json_mimetype_nominal(self):

        print(is_json_mimetype("application/json"))
        self.assertTrue(is_json_mimetype("application/json"))
        self.assertTrue(is_json_mimetype("application/json; charset=utf-8"))
        self.assertTrue(is_json_mimetype("application/json;charset=utf-8"))
        self.assertTrue(is_json_mimetype("application/anything+json;charset=utf-8"))
        self.assertTrue(is_json_mimetype("application/anything+json"))

        self.assertFalse(is_json_mimetype("application/xml"))
        self.assertFalse(is_json_mimetype("application/json;charset=utf-dfdfd"))

````



`

Comment (Jiehong) [reply connexion/utils.py:136]: I think this is better. (Just remove the print in the test, and maybe simplify the return value of the function, as it's a 9-lines return value).

Go ahead!

Comment (scoulomb) [reply connexion/utils.py]: Thanks for your suggestion I have updated accordingly.
I made some test with a sample server to check behavior here:
https://github.com/scoulomb/connexion/commit/fc315835c1de37707835f9d1f8043588e255d35e

**Comment by scoulomb** [comment]
Fix loss of coverage here: https://github.com/zalando/connexion/blob/c99dda68f4cad9e36815267eb68898220499ee5c/connexion/decorators/validation.py#L142

**Comment by scoulomb** [comment]
I added unit test in  test_json_validation.py 
However loss of coverage was due to the fact that in the former version of the code we were doing tuple unpacking after the split which could lead to a ValueError exception

>>> mimetype.split('/')
['application-json']
>>> a,b = mimetype.split('/')
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 1)

IMO it is a workaround to test a side effect of the function outsside of it:
As we now test separator presence it is useless to catch this exception so I will remove the try/except block

**Comment by scoulomb** [comment]
Hello @hjacobs , I fixed https://github.com/zalando/connexion/issues/1202 in this PR. Would you mind please have a look? Thanks a lot.

File: connexion/utils.py [hunk 2]
```diff
@@ -131,8 +131,21 @@ def is_json_mimetype(mimetype):
     :type mimetype: str
     :rtype: bool
     """
-    maintype, subtype = mimetype.split('/')  # type: str, str
-    return maintype == 'application' and (subtype == 'json' or subtype.endswith('+json'))
+    simple_mime_type, optional_parameter = mimetype.split(";") if \
```
hjacobs, line 134: you can simplify this code by using `mimetype.partition(";")` instead

connexion/utils.py
```diff
@@ -131,8 +131,21 @@ def is_json_mimetype(mimetype):
...
+        ";" in mimetype else \
+        (mimetype, None)
+    if "/" not in simple_mime_type:
+        return False
+    maintype, subtype = simple_mime_type.split('/')  # type: str, str
```
hjacobs, line 139: same here: `partition(..)` will always return 3 values, so we can get rid of the `if`

scoulomb: @hjacobs Very elegant. Thanks. I updated.

scoulomb:
@hjacobs do you think this PR is mergeable, or should I rework on it?

RobbeSneyders:
This seems like a good change @scoulomb. If you're still interested in getting it merged, please rebase on and target this PR to the `v2` branch.

PR #1208, opened 2020-04-07
fix: add parameter check to endpoint without parameters

### LINKED ISSUE #1207: Endpoints without parameters return 200 when passed a parameter
### Description
Endpoints that dont have any parameters specified (for example a health check) return a 200 code even when passed a parameter, it should return a 4XX code like endpoints with parameters when strict_validation is set to True

### Expected behaviour
return 4XX code when wrong parameter is passed

### Actual behaviour
returns a 200

### Steps to reproduce
create endpoint without any parameters and then call it with parameters while strict_validation is True

### Additional info:

Output of the commands:

- `python --version` 3.8.1
- `pip show connexion | grep "^Version\:"` 2.6.0

jeremyschiemann:
Fixes #1207 

 - remove condition to only check parameters if there are any

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -417,10 +417,9 @@ class AbstractOperation(SecureOperation, metaclass=abc.ABCMeta):
         """
         ParameterValidator = self.validator_map['parameter']
         RequestBodyValidator = self.validator_map['body']
-        if self.parameters:
-            yield ParameterValidator(self.parameters,
-                                     self.api,
-                                     strict_validation=self.strict_validation)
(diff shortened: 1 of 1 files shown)


hjacobs:
I don't exactly understand this change. Also the Travis CI build is now failing.

Ruwann:
Are you still interested in moving this PR forward? If so, please rebase onto main as we have moved from `master` to `main`.

It might make more sense to only do the check if strict_validation is used? That's what I get from the linked issue.

RobbeSneyders:
Closing due to lack of response. Feel free to reopen against `main`.

PR #1212, opened 2020-04-14
Document token_info argument

### LINKED ISSUE #1188: How to access tokeninfo response in handler function?
Is it somehow possible to access/pass the full oauth tokeninfo response to the handler function?

We have added some non-standard fields to our tokeninfo response and it'd be nice to be able to pass them to the handler function.

Right now it seems only the `sub` field from the response can be passed to the handler function as `user` argument. Would it be possible to add similar support for the whole tokeninfo response?

mnylen:
Fixes #1188

 - Add note that protected handler functions accepting `token_info` will be passed the full token info response in that argument.


hjacobs:
Thanks!

hjacobs:
:+1:

PR #1217, opened 2020-04-16
Fix _framework_to_connexion_response for StreamResponse

ClemDoum:
Returning a `aiohttp` `StreamResponse` (and probably `FileResponse`) was failing the `AioHttpApi._framework_to_connexion_response` made the assumption that the `response` object had a body, which is not the case for `StreamResponse` and `FileResponse`

 - We now check if the `response` given to `_framework_to_connexion_response` has a `body` attribute and default to a `None` body otherwise
 - I prefered to use: 
    ```python
    if hasattr(response, "body"):
        body = response.body
    ```
    rather than:
    ```python
    if isinstance(response, web.Reponse):
        body = response.body
    ```
     in prevision of the unlike case where `aiohttp` would add another `web.StreamResponse` sublcass without body


hjacobs:
:+1:

PR #1218, opened 2020-04-17
Switched from mock to unittest.mock

jlaba:
Since you dropped python 2.7 support, a switch from mock to unittest.mock is possible. The switch is done in this pull request.

By the way:
I made a request for packaging connexion-2.6.0 for debian unstable and already submitted a package. If you are interested in it, have a look at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958042

### DIFF
--- a/tests/decorators/test_parameter.py
+++ b/tests/decorators/test_parameter.py
@@ -1,5 +1,5 @@
 # we are using "mock" module here for Py 2.7 support
-from mock import MagicMock
+from unittest.mock import MagicMock
 
 from connexion.decorators.parameter import parameter_to_arg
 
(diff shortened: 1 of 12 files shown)


hjacobs:
Makes sense, thanks!

hjacobs:
:+1:

tests/decorators/test_parameter.py
```diff
@@ -1,5 +1,5 @@
 # we are using "mock" module here for Py 2.7 support
```
hjacobs, line 1: can you fix the comment here?

tests/test_validation.py
```diff
@@ -3,7 +3,7 @@
 import flask
 import pytest
 # we are using "mock" module here for Py 2.7 support
```
hjacobs, line 5: can you fix (remove) the comment here?

hjacobs:
:+1:

Issue #1193, opened 2020-03-27
RequestBody with both form-data and json payload does not work


ohurley:
### Description

I am trying to upload an audio file. The content contains the a file (form-data) and my schema component (application/json). The code can read the file but it cannot identify the json

Could you provide me with an example spec that can cater for this functionality? 

### Expected behaviour

I added content for both application/json & multipart/form-data (or audio/*) and the request proceeds without error

### Actual behaviour
It fails on validation against the json payload
```
{
    "_schema": [
        "Invalid input type."
    ]
}
```
The reason for this is that `get_json()` method returns None. It looks like connexion does not like 2 different requestBody components

### Additional info:
Code snippet
```
def put():
    req_data = request.get_json()
    errors = None
    try:
        data = task_schema.load(req_data)  # , partial=True)
    except ValidationError as err:
        errors = err.messages  # It throws the error
       
```

spec:
```/task:
    put:
      tags:
        - Tasks
      summary: Endpoint to create new or update existing Task object
      operationId: resources.task.put
      responses:
        '200':
          description: Updated
        '201':
          description: Created
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TaskResponse'
        '202':
          description: Patient has not been created yet
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Task'
          audio/*:
            schema:
              type: string
              format: binary


....

Task:
      type: object
      properties:
        t_id:
          type: string
          description: Unique identifier for a Task object
          example: '2305a17b-6aa0-4250-9c60-e9828f493057'
        a_id:
          type: string
          description: Unique identifier for a Assessment object (FK)
          example: '2305a17b-6aa0-4250-9c60-e9828f493057'
        t_type:
          type: string
          description: Task type
          example: 'T1'
        description:
          type: string
          description: Task name
          example: ' Sustaining vowel– a'
        uri:
          type: string
          description: Storage location for the audio recording file
          example: 'aws:bucket/assessments/file.wav'
        created:
          type: string
          format: date-time
          description: Creation time
          example: '2015-07-07T15:49:51.230+02:00'
        modified_at:
          type: string
          format: date-time
          description: Creation time
          example: '2015-07-07T15:49:51.230+02:00'
#        fileName:
#          type: object
#          description: file
#          properties:
#            content-type:
#              type: 'application/octet-stream'

```

Output of the commands:

- `python --version`
Python 3.6.8
[...]

Issue #992, opened 2019-07-08
Multiple files upload (mutlipart/form-data) is ignored for Flask


kornicameister:
### Description

Consider following endpoint definition:
```yaml
  /test-formData-file-upload:
    post:
      summary: 'Test formData with file type, for file upload'
      operationId: fakeapi.hello.test_formdata_file_upload
      responses:
        '200':
          description: OK
      requestBody:
        content:
          multipart/form-data:
            schema:
              x-body-name: formData
              type: object
              properties:
                formData:
                  type: array
                  items:
                    type: string
                    format: binary
              required:
                - formData
```

What we are doing in here is uploading multiple files from the same mulipart field.
That gets correctly pulled from `flask_request` but ignored later in `connexion.operations.abstract.AbstractOperation#_get_file_arguments`

### Expected behaviour

Request to Flask-based application with multiple files is correctly processed and we get multiple files available inside of handler.

### Actual behaviour

Multiple files are ignored, only first is taken into account.

### Steps to reproduce

1. open `connexion/tests/fixtures/simple/openapi.yaml`
2. find `/test-formData-file-upload` definition
3. Use snippet from above to replace the definition
4. Go to `tests.api.test_parameters.test_formdata_file_upload`
5. Use following code to replace the function
```python
def test_formdata_file_upload(simple_app):
    app_client = simple_app.app.test_client()
    resp = app_client.post('/v1.0/test-formData-file-upload',
                           data={'formData': [(BytesIO(b'file contents'), 'filename1.txt'), (BytesIO(b'file contents'), 'filename2.txt')]})
    assert resp.status_code == 200
    response = json.loads(resp.data.decode('utf-8', 'replace'))
    assert response == {'filename1.txt': 'file contents'}
```
6. Observe test to pass, which it should not do.

### Additional info:

Output of the commands:

- `python --version` -> `3.7.3`
- `pip show connexion | grep "^Version\:"` -> `master branch`
- `flask --version` -> `1.1.0`

kornicameister:
cc @dtkav @hjacobs

simondrabble:
Looks like a duplicate of #510. You can see my fork here https://github.com/simondrabble/connexion for a fix.

dtkav:
I think this is closed by #1000 
Please re-open if there's something I missed.

guillaumedsde:
> I think this is closed by #1000
> Please re-open if there's something I missed.

[...]

PR #976, opened 2019-06-14
Fix for issue #975

vasiliy-garenskikh:
Fixes #975  .

Changes proposed in this pull request:

 - `files` added to aiohttp `ConnexionRequest` class

PR #1222, opened 2020-04-22
Fix for aiohttp and multipart/form-data uploads

ddurham2:
I'm trying to clean some things up around aiohttp and file uploads, and close a few tickets in the process.  There are a couple of abandoned PRs (mentioned below) which I'm trying to revitalize efforts around.

Changes proposed by this fix:
- in aiohttp request handling, splitting out multipart POSTs into files[] and form[], handling duplicate keys as the rest of connexion expects
- rewrote how operations/openapi.py::_get_body_argument() works to better build the arguments[] list according to what the spec says and what the handler accepts.  This fixes a bug when requests contain mixed files and form values and the handler is expecting variable names matching the request property names
- introduce unit tests to demonstrate the problem

**Replaces PRs:**
- #987 - [aiohttp] Multipart handling
- #976 - Fix for issue #975

**Fixes:**
- #975 - File uploading with aiohttp (status 500)
- #992 - Multiple files upload (multipart/form-data) is ignored for Flask

**Should Fix:** (I consulted the original authors, but no response)
- #1193 - RequestBody with both form-data and json payload does not work
- #909 - TypeError: cannot convert dictionary update sequence element #0 to a sequence

**Relates To:**
- #935 - multipart/form-data content validation


setup.py
```diff
@@ -32,7 +32,7 @@ def read_version(package):
 swagger_ui_require = 'swagger-ui-bundle>=0.0.2'
 flask_require = 'flask>=1.0.4'
 aiohttp_require = [
-    'aiohttp>=2.3.10',
+    'aiohttp>=2.3.10,<4',
```
ddurham2, line 35: added < v4 because it was pulling a version 4 (tho it's not stable yet), and failing badly.

dfeinzeig:
@ddurham2 did you get a chance to confirm the "may fixes" as confirmed fixes? i desperately need the fix for 992 but also need 1193. thanks!

dfeinzeig:
oh, these fixes are specific to aiohttp and don't seem to work with flask?

ddurham2:
Some of the changes could actually influence Flask behavior since they change shared code.

tests/aiohttp/test_aiohttp_multipart.py
```diff
@@ -0,0 +1,100 @@
...
+    data = await resp.json()
+    assert resp.status == 200
+    assert data['fileName'] == f'{__name__}.py'
+
+
+async def test_many_files_upload(aiohttp_app, aiohttp_client):
+    app_client = await aiohttp_client(aiohttp_app.app)
+
+    dir_name = os.path.dirname(__file__)
+    files_field = [('files', open(f'{dir_name}/{file_name}', 'rb')) for file_name in os.listdir(dir_name) if file_name.endswith('.py')]
```
kornicameister, line 46: I'd suggest breaking this line, each file goes in its own line.

tests/aiohttp/test_aiohttp_multipart.py
```diff
@@ -0,0 +1,100 @@
...
+
+    resp = await app_client.post(
+        '/v1.0/upload_files',
+        data=form_data(),
+    )
+
+    data = await resp.json()
+
+    assert resp.status == 200
+    assert data['filesCount'] == len(files_field)
```
kornicameister, line 58: apart from that, do check if all file names matches, might be even for the best to see if content matches; you don't want anything affecting file content.

tests/aiohttp/test_aiohttp_multipart.py
```diff
@@ -0,0 +1,100 @@
...
+    app_client = await aiohttp_client(aiohttp_app.app)
+
+    resp = await app_client.post(
+        '/v1.0/upload_file',
+        data=aiohttp.FormData(fields=[('funky_funky', open(__file__, 'rb'))])(),
+    )
+
+    data = await resp.json()
+    assert resp.status == 200
+    assert data['fileName'] == f'{__name__}.py'
```
kornicameister, line 39: ditto about asserts

tests/aiohttp/test_aiohttp_multipart.py
```diff
@@ -0,0 +1,100 @@
...
+    assert resp.status == 200
+    assert data['dirName'] == os.path.dirname(__file__)
+    assert data['fileName'] == f'{__name__}.py'
+
+
+async def test_mixed_multipart_many_files(aiohttp_app, aiohttp_client):
+    app_client = await aiohttp_client(aiohttp_app.app)
+
+    dir_name = os.path.dirname(__file__)
+    files_field = [('files', open(f'{dir_name}/{file_name}', 'rb')) for file_name in os.listdir(dir_name) if file_name.endswith('.py')]
```
kornicameister, line 84: ditto breaking the line

connexion/apis/aiohttp_api.py
```diff
@@ -293,13 +293,50 @@ async def get_request(cls, req):
...
+                'url': url,
+            },
+        )
 
         query = parse_qs(req.rel_url.query_string)
         headers = req.headers
         body = None
-        if req.body_exists:
+
+        # if request is not multipart, `data` will be empty dict
```
kornicameister, line 314: oh, hey :D that's mine :D

connexion/apis/aiohttp_api.py
```diff
@@ -293,13 +293,50 @@ async def get_request(cls, req):
...
+        files = {}
+        form = {}
+
+        if post_data:
+            logger.debug('Reading multipart data from request')
+            for k, v in post_data.items():
+                if isinstance(v, web.FileField):
+                    if k in files:
+                        try:
+                            files[k].append(v)
```
kornicameister, line 328: unless you want to push for #987 being merged first, I'd suggest that put a decent home here explaining that this is nasty workaround or something ;)

connexion/apis/aiohttp_api.py
ddurham2: Uh yeah.. much of this is yours ;)  ... including the funny-formatted unit tests above.  I started with your PR.  I thought I said that somewhere.  Maybe it's in my internal notes or a commit msg somewhere.

Anyway, it was much helpful in getting started.

connexion/apis/aiohttp_api.py
ddurham2: I indicated in the opening notes that this would replace PR #987 (since it's based on it). I was just trying to finish what you started since you indicate you had moved on to axion.

by "nasty workaround", are you referring to the try/except nature of it?  It's just trying to put files which have the same name into a list.. not sure how that should happen, this isn't a query string.  But hey.

tests/aiohttp/test_aiohttp_multipart.py
ddurham2: What do you mean "breaking this line"?  Just that it's formatted too long?  What do you mean "each **file** goes in its own **line**"?

tests/aiohttp/test_aiohttp_multipart.py
ddurham2: Well, I thought that too the first time I saw it, but you wrote it :)   I'll fix.

ddurham2: done

tests/aiohttp/test_aiohttp_multipart.py
ddurham2: done

tests/aiohttp/test_aiohttp_multipart.py
ddurham2: I committed some reformatting.  Is that what you had in mind?

kornicameister: yep

connexion/apis/aiohttp_api.py
kornicameister: well, just look at that, does that look ok to you? figuring that out was one pain point, understanding that will be another one :(

ddurham2: and done

ddurham2:
@hjacobs any thoughts on merging this?  Anything I need to do to prepare it further?  Thanks!

connexion/operations/openapi.py
```diff
@@ -268,36 +268,79 @@ def body_definition(self):
...
+        # argument, or the broken out arguments, or both?
+        #
+        # #1 If 'x-body-arg' is explicitly given and it exists in [arguments], then the
+        # body, as a whole, will be passed to the handler with that name. STOP.
+        #
+        # #2 If kwargs is given, then we don't know what the handler cares about, so we
+        # pass the body as a whole as an argument named, 'body', along with the
+        # individual body properties. STOP.
+        #
+        # #3 Else, we pass the body's individual properties which  exist in [arguments].
```
dtkav, line 325: I don't think this piece makes sense to add. IIRC this was the behavior for swagger2, but I think it's an anti-pattern. Passing in a body object avoids some issues like parameter name conflicts, and is much less "magical".

ddurham2: It'd sure break a lot of code to remove #3.  It's the whole "Automatic Parameter Handling" that's mentioned in the examples in README.rst.  This PR is only preserving that behavior, not changing it.

ddurham2:
bump?

vgonisanz:
I have arrived to this PR this week. I'm using openapi generator to create an auto-generated server stub with python-flask generator and it uses the latest version of connexion (v.2.7.0).

In my OpenAPI specification I've defined a POST endpoint to accept a 'requestBody' with two properties, an audio file and a string. When I try to make a request to this endpoint by filling correctly both properties (file and string), the file property is parsed correctly and I receive it at my controller. However, the string is never parsed and my controller doesn't receive it, so my core logic doesn't work properly.

After debugging connexion code, I've found that the problem is at '_get_body_argument(...)' function inside 'operations/openapi.py'. It seems to be looking for a 'x-body-name' tag equal to one of the 'request.form' arguments (line 298), but this condition is never met.

Looking into your repo, I've found that the patch of this pull request fixes the issue (I've tested it) if you don't use the `pythonic_params=True` flag (It seems a value 'value', after the regen try to look for ' value_'  and don't find it but that's another problem).

Is this PR planned to be merged in the next version of the connexion library?

Thanks for your time and @ddurham2 for the patch ;-)

ddurham2:
Your welcome @vgonisanz, you and I are apparently using it for the exact same purpose (POST of an audio file, plus some additional data.. ha)

@hjacobs, @dtkav, @jmcs -- sorry to bug.  Is there any chance this could be merged?

Tohnmeister:
@ddurham2 @vgonisanz 
Same here. This specific issue appeared to be fixed with PR https://github.com/zalando/connexion/pull/954 but for some reason that PR was never merged to main. No clue why.

RichardBruskiewich:
@dtkav @kornicameister (cc: @ddurham2)

I don't know if this is the best context to raise this, but it isn't just the **multipart/form-data** which is broken here, but also, the **application/x-www-form-urlencoded** processing.

A similar runtime error to the multipart upload occurs in the current [validation.py](https://github.com/zalando/connexion/blob/1ed14fbde5be3d144760fec4982e90064f151e88/connexion/decorators/validation.py#L162) which has the following code: ```
                data = dict(request.form.items()) or (request.body if len(request.body) > 0 else {})
                data.update(dict.fromkeys(request.files, ''))  # validator expects string..
```

The problem arises in that request.body from a **application/x-www-form-urlencoded** POST is a binary (byte) string of urlencoded parameters. Needless to say, a `byte` string doesn't have an `update` method, so data.update throws an exception.

What I notice is the following in [PR #1222](https://github.com/zalando/connexion/pull/1222) is that that `form` argument is added to the ConnexionRequest object:

```
        return ConnexionRequest(url=url,
@@ -309,7 +349,8 @@ async def get_request(cls, req):
                                headers=headers,
                                body=body,
                                json_getter=lambda: cls.jsonifier.loads(body),
                                files={},
                                form=form,
                                files=files,
                                context=req)
```
That's a good step forward. However, where is it initialized with the contents of the body taken as urlencode parameters?

I'm not quite sure where to intervene in the code, but this bug is currently a show stopper for my use of AIOHTTP in my project (which is in a somewhat urgent _need-to-get-the-prototype-running_ state for an NIH funded biomedical information system project).

Cheers
Dr. Bruskiewich

**Comment by RichardBruskiewich** [comment]
@ddurham2, thank you for your diligent work on this PR. I actually forked 'connexion' then merged your PR code, with perhaps just a few local tweaks (not sure if some of them are still necessary), and the PR essentially fixed my problems with the 'multipart/form-data' uploads, so that I was finally able to get an aiohttp version of my application working.

I hope that the Connexion team will soon consider merging this PR or an equivalent, to make things work for everyone.

**Comment by ddurham2** [comment]
Thanks for you kind words.  Glad it worked for you too.  As for merging, sadly, I'm increasingly fearful that this project is abandoned by the authors.  I'm hoping a new official port can start or at least that the maintainers will happily give maintenance perms to someone else.

**Comment by ddurham2** [comment]
humble bump

File: connexion/apis/aiohttp_api.py [hunk 9]
```diff
@@ -293,13 +293,53 @@ async def get_request(cls, req):
...
+        )
 
         query = parse_qs(req.rel_url.query_string)
         headers = req.headers
         body = None
-        if req.body_exists:
+
+        # if request is not multipart, `data` will be empty dict
+        # and stream will not be consumed
+        post_data = await req.post()
```
zwerheim, line 316: @ddurham2 
I this approach may be problematic.

Referencing this doc:
https://docs.aiohttp.org/en/stable/web_quickstart.html?highlight=multipart#file-uploads
And the source code here:
https://github.com/aio-libs/aiohttp/blob/v3.7.4/aiohttp/web_request.py#L639

The above code comment seems to be assuming that the stream will not be consumed, but as per the above source code it does read the entire multipart request into memory. 

What I was finding, was that running something like 
```
multipart = await request.multipart()
part = await multipart.next()
```

results in the following error:
```
ValueError: Could not find starting boundary b'--94aa6f10ca444684a68e350816dbb465'
```

This is because connexion already reads and drains the stream before it gets to my handler. The `File Uploads` link above points out that file uploads should be handled using `request.multipart` to avoid OOM errors.

I'm not familiar with the rest of connexion yet to suggest solution, but I wanted to raise it here as we've been troubleshooting why `request.multipart` was not working.

ddurham2: Thanks for your feedback.

I believe what I meant by that comment was that if it's not multipart, then getting the value of post() will result in the post_data variable being empty (I should correct the comment to say 'post_data') and that it won't consume the request stream.  
However, looking at the code ( https://github.com/aio-libs/aiohttp/blob/v3.7.4/aiohttp/web_request.py#L629 ) that should actually say if it's not 'application/x-www-form-urlencoded' nor 'application/x-www-form-urlencoded' then it will be empty... 
I will clarify the comment, and it appears that my change will fix x-www-form-encoded forms as well, which, it appears, previously wouldn't extract those values from the request.

On the concern about it reading into memory, it actually won't necessarily read and hold the entire request data in memory.  My main concern there would be uploaded files (as seems to be your concern too), being too large to fit in memory.  But as you can see in https://github.com/aio-libs/aiohttp/blob/v3.7.4/aiohttp/web_request.py#L656 , it efficiently streams any files in the request from the network into temp files.
However, any non-files values will be momentarily read into memory (one at a time) but will discard them if they are larger than a max.  See https://github.com/aio-libs/aiohttp/blob/v3.7.4/aiohttp/web_request.py#L663  aiohttp could perhaps do a better job at reading only up to the maximum acceptable field size into memory rather than reading the whole thing in before rejecting it.  But that's an enhancement for them to make.

Now, to your point about not being able to read the data from the request yourself... Yes, it means that if it's multipart (or x-www-form-encoded forms), then the request object passed to an application's handler cannot be used to read the request a second time using multipart().  However, `await post()` should still work and return a dict with the form values.
Is using the request object for this purpose something you _needed_ to do, or just something you tried to do (with my patch in place) due to the assumed memory issues?

ddurham2:
@RobbeSneyders  - Hi Robbe.. I am SO pleased to see connexion getting the needed attention and development it needs.  I have a project that heavily depends on it, and it has the most convenient design for implementing an openapi service based on python, and best of all it implements an aiohttp backend which is important for me.

That said, I would request that you take a look at this pull request and consider merging it.  It fixes a few linked issues.  And a few other people complaining about the same problem.  The main problem solved is that connexion cannot handle a form submission that has files and other parameters at the same time.

I've had to maintain a custom fork with these changes since finding its problem and creating the fix, and really would like to get back on the main release.  

Thank You

connexion/apis/aiohttp_api.py
```diff
@@ -305,13 +305,52 @@ async def get_request(cls, req):
...
+                'url': url,
+            },
+        )
 
         query = parse_qs(req.rel_url.query_string)
         headers = req.headers
         body = None
-        if req.body_exists:
+
+        # Note: if request is not 'application/x-www-form-urlencoded' nor 'application/x-www-form-urlencoded',
```
RobbeSneyders, line 326: I'm guessing you meant for one of these to be 'multipart/form-data'?

connexion/operations/openapi.py
RobbeSneyders: Agree with @dtkav here. This isn't the current behavior for OpenAPI 3 and I don't think we should support it.

RobbeSneyders (changes requested):
Thanks for the PR @ddurham2!

If you can drop the automatic body parsing, we should be able to merge.

connexion/operations/openapi.py
ddurham2: @RobbeSneyders, thanks for taking the time to look at this.

I just wanted to get clarification on this.  So you're saying that if there's a body parameter named 'x' declared in the spec and there's an argument named 'x' in the handler, then it should not try to match that body parameter to the handler argument?  Isn't that inconsistent with how query and path parameters work?  Is there a reason they should differ in behavior?  I'm guessing what @dtkav was getting at was that body params may conflict with the others?  But couldn't they conflict with each other just the same?  Is there a conflict resolution strategy for path and query?

If the fear is breaking existing apps, that's understandable.  My thought was that dealing with body directly is breaking the abstraction that connexion is trying to put over those lower level details.

Again, I'm willing to change it.. just trying to understanding the reasoning for future reference.

Thanks

RobbeSneyders: > So you're saying that if there's a body parameter named 'x' declared in the spec and there's an argument named 'x' in the handler, then it should not try to match that body parameter to the handler argument?

Indeed, we should only pass the body as a whole. I don't think this is inconsistent with the query and path parameters. Query and path parameters are top level, just as the body. The content of the body is one level below. We only pass the top level.

Next to this, passing the body is a lot less magical as @dtkav mentioned. The user knows what to expect, which isn't the case with the current implementation in this PR, shown by the 5 paragraph comment needed to explain the behavior :slightly_smiling_face:

Path and query parameters can indeed also conflict with each other. Currently [the query parameters overwrite the path parameters](https://github.com/spec-first/connexion/blob/main/connexion/operations/abstract.py#L280). Probably not ideal, however conflicting parameters should ideally be avoided by the user.

connexion/apis/aiohttp_api.py
ddurham2: So, I reverted back to the original behavior for _get_body_argument() which no longer breaks body parameters into handler arguments.

However, in the process I learned, or.. remembered (this original change being so long ago now).. that the reason I though body should be split out is because uploaded files in the body _are_ broken out as individual handler arguments 
 which is an existing behavior apparently.

My main change here was to make the aiohttp handler consistent with flask.  I did that by constructing the ConnexionRequest() object in aiohttp_api.py the same as flask does regarding files.

So, I'm just pointing out that some future cleanup might want to make the file elements stay inside of body, where I'd think you'd say they belong due to the reasoning that they're not top-level parameters.  But I think it doesn't work that way because this is basically flask's existing behavior.

ddurham2: Hopefully, we're ready to merge now.  Thanks!

connexion/operations/openapi.py
ddurham2: So, I reverted back to the original behavior for _get_body_argument() which no longer breaks body parameters into handler arguments.

However, in the process I learned, or.. remembered (this original change being so long ago now).. that the reason I though body should be split out is because uploaded files in the body _are_ broken out as individual handler arguments 
 which is an existing behavior apparently.

My main change here was to make the aiohttp handler consistent with flask.  I did that by constructing the ConnexionRequest() object in aiohttp_api.py the same as flask does regarding files.

So, I'm just pointing out that some future cleanup might want to make the file elements stay inside of body, where I'd think you'd say they belong due to the reasoning that they're not top-level parameters.  But I think it doesn't work that way because this is basically flask's existing behavior.

ddurham2: I think we're ready to merge now.  Thanks!

tests/fixtures/aiohttp/openapi_multipart.yaml
```diff
@@ -6,11 +6,11 @@ info:
   title: "{{title}}"
   version: "1.0"
 paths:
-  "/mixed_single_file":
```
ddurham2: Here, I just moved the bottom two endpoints to the top because that's the same order they are in in the test handler and test requester files

RobbeSneyders (approved):
Thanks @ddurham2, LGTM :)

PR #1233, opened 2020-05-04
Doc/security handler

Jiehong:
Fixes warning on request.rst at doc generation time, and inline the example for the security with basic auth.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -11,7 +11,7 @@ With Connexion, the API security definition **must** include a
 ``x-tokenInfoFunc`` must contain a reference to a function
 used to obtain the token info. This reference should be a string using
 the same syntax that is used to connect an ``operationId`` to a Python
-function when routing. For example, an ``x-tokenInfoFunc`` with a value of 
+function when routing. For example, an ``x-tokenInfoFunc`` with a value of
 ``auth.verifyToken`` would pass the user's token string to the function
 ``verifyToken`` in the module ``auth.py``. The referenced function accepts
 a token string as argument and should return a dict containing a ``scope``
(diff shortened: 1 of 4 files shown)


docs/security.rst
```diff
@@ -56,10 +56,18 @@ semantics as for ``x-tokenInfoFunc``, but the function accepts three
 parameters: username, password and required_scopes. If the security declaration
 of the operation also has an oauth security requirement, required_scopes is
 taken from there, otherwise it's None. This allows authorizing individual
-operations with oauth scope while using basic authentication for
+operations with `oauth scope`_ while using basic authentication for
 authentication.
 
-You can find a `minimal Basic Auth example application`_ in Connexion's "examples" folder.
+You can find a `minimal Basic Auth example application`_ in Connexion's "examples" folder:
```
scoulomb, line 62: maybe we could also point to the readme of the example

docs/security.rst
```diff
@@ -56,10 +56,18 @@ semantics as for ``x-tokenInfoFunc``, but the function accepts three
...
+
+
+.. literalinclude:: ../examples/openapi3/basicauth/app.py
```
scoulomb, line 65: In the example I am wondering if  this line https://github.com/zalando/connexion/blob/bed4b95205205861bd6014282d0f61d50231d3ac/examples/openapi3/basicauth/app.py#L31 makes sense?

Jiehong: The example was showing the oauth2 as well, so I fixed that by improving the documentation and linking to the correct example for oauth2, and I removed that dummy_func function as well, as it no longer makes sense.

docs/security.rst
Jiehong: Yes, you're right. I removed the inline code, and used the link instead.

Jiehong:
@szborows : hi, what do you think of this PR?

jmcs:
:+1:

PR #1186, opened 2020-03-21
Python 3.8: replace @asyncio.coroutine/yield from with async/await #1182

szborows:
Fixes #1182 .

Changes proposed in this pull request:

 - replaced `@asyncio.coroutine` and `yield from` with `async` and `await` respectively

PR #1234, opened 2020-05-07
Python 3.8: replace asyncio.coroutine/yield from with async/await in decorators

vmarkovtsev:
This is the place missed in https://github.com/zalando/connexion/pull/1186 that gives annoying `DeprecationWarning`-s in `pytest`.

### DIFF
--- a/connexion/decorators/coroutine_wrappers.py
+++ b/connexion/decorators/coroutine_wrappers.py
@@ -13,23 +13,23 @@ def get_request_life_cycle_wrapper(function, api, mimetype):
     :rtype asyncio.coroutine
     """
     @functools.wraps(function)
-    def wrapper(*args, **kwargs):
+    async def wrapper(*args, **kwargs):
         connexion_request = api.get_request(*args, **kwargs)
         while asyncio.iscoroutine(connexion_request):
-            connexion_request = yield from connexion_request
+            connexion_request = await connexion_request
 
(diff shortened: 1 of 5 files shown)


vmarkovtsev:
@hjacobs can you please merge this no-brainer.

connexion/decorators/coroutine_wrappers.py
```diff
@@ -13,23 +13,23 @@ def get_request_life_cycle_wrapper(function, api, mimetype):
     :rtype asyncio.coroutine
```
cognifloyd, line 13: There's a `yield from` on line 9. Though commented, we might as well update it.
You can see this in my [await](https://github.com/cognifloyd/connexion/commit/0ad493e89179642b06a2c7b8eb2a2bffca75f85c) branch.

vmarkovtsev:
> I wonder about inlining coroutine_wrappers as we no longer support python 2.7

Yes, such refactorings are completely out of scope of this PR and I really don't want to do any extra maintenance work here. Please don't make me do random related work.

vmarkovtsev:
I cherry-picked https://github.com/cognifloyd/connexion/commit/b908e44b7efb7c140338b69b392c42eff9b62437

RobbeSneyders:
Thx for the PR @vmarkovtsev, but this change has already been merged as part of #1372.

PR #1257, opened 2020-06-24
avoid unnecessary object attribute lookup

houqp:

 - router_controller has already been access before the if state, so we can just reuse it


jmcs:
:+1:

jmcs:
Thanks for the PR.

Issue #401, opened 2017-02-09
Returning files via connexion.send_file or connexion.send_from_directory return an error


thedylman:
### Description
Using connexion.send_file or connexion.send_from_directory return an error. I think this is because they don't return an ResponseContainer object.

### Expected behaviour
File should be sent in response.

### Actual behaviour
File not sent in response, console shows call stack:
```
  Traceback (most recent call last):
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/test/download/venv/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/decorator.py", line 118, in wrapper
    response_container = function(*args, **kwargs)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/produces.py", line 59, in wrapper
    response = function(*args, **kwargs)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/response.py", line 90, in wrapper
    response = function(*args, **kwargs)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/decorator.py", line 99, in wrapper
    return self.get_response_container(operation_handler_result)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/decorator.py", line 61, in get_response_container
    self.mimetype, response=operation_handler_result)
  File "/home/test/download/venv/lib/python3.5/site-packages/connexion/decorators/decorator.py", line 153, in __init__
    self.data = self._response.get_data()
  File "/home/test/download/venv/lib/python3.5/site-packages/werkzeug/wrappers.py", line 906, in get_data
[...]

PR #1259, opened 2020-06-25
Don't validate streamed responses

### LINKED ISSUE #904: Response validation with streaming
### Description

When response validation is turned on, it’s not possible to use Flask streaming. The response is buffered by Connexion and flushed after being validated. That’s not necessarily incorrect, but requires to entirely disable response validation.

### Expected behaviour

It’d be great to have some way to have response validation turned on yet still be able to use Flask streaming. Even disabling the validation for specific operation would be sufficient.

### Actual behaviour

Now it’s only possible to disable response validation for the whole API.

### Steps to reproduce

Add the API with response validation turned on.

```python
    connexion_app.add_api("api_spec.yaml", validate_responses=True)
```

Define a streaming operation

```yaml
/stream
  get:
    operationId: api.stream
    summary: Streaming operation
    description: A stream of Newline Delimited JSON objects.
    responses:
      '200':
        description: Successfully streamed some JSONs.
        content:
          application/x-ndjson: {}
```

```python
from flask import Response
from flask.json import dumps


def stream():
    def generate():
        for i in range(10):
            item = dumps({"i": i})
            yield(item)
    return Response(generate(), mimetype="application/x-ndjson")
```

### Additional info:

Output of the commands:

- `python --version`
   `Python 3.6.7`
   `Version: 2.2.0`

cshorler:
prior to validation get_connexion_response ultimately calls response.get_data()

I haven't found a workaround for this and the old PR was closed due to inactivity.

I changed it to use getattr as the object may not be a response (could be tuple).

Fixes #904 .

there are also a few other issues raised.

 - bypass for response.is_streamed, otherwise no change in behaviour
 -
 -

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -163,6 +163,7 @@ class FlaskApi(AbstractAPI):
             content_type=response.content_type,
             headers=response.headers,
             body=response.get_data(),
+            is_streamed=response.is_streamed
         )
 
     @classmethod
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -162,7 +162,7 @@ class FlaskApi(AbstractAPI):
             mimetype=response.mimetype,
             content_type=response.content_type,
             headers=response.headers,
-            body=response.get_data(),
+            body=response.get_data() if not response.direct_passthrough else None,
             is_streamed=response.is_streamed
         )
 
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -86,9 +86,9 @@ class ResponseValidator(BaseDecorator):
         """
 
         def _wrapper(request, response):
-            if not getattr(response, 'is_streamed', False):
-                connexion_response = \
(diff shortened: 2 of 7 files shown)


RobbeSneyders:
Hi @cshorler, thx for the PR. Hopefully you're still interested in merging this :)

In the perfect case, we might want to try to validate the response in a streaming fashion, but this already looks like an improvement. I see two downside as well though:

- Response headers are also not validated anymore
- Users might not expect that validation is deactivated on streaming responses

I think we can address this by logging a warning about this when we skip validation.

Also, we moved to the main branch, so can you please rebase and change the target of your PR?

RobbeSneyders:
Also, since `is_streamed` is Flask / werkzeug specific, we should add this as a property to the `ConnexionResponse` class instead of accessing it directly.

RobbeSneyders:
Fixes #401 as well

Ruwann (approved):
Thanks!

This is currently only for flask. For aiohttp, we might need to check what the response class is (https://docs.aiohttp.org/en/stable/web_reference.html#response-classes) but I'm not that familiar with aiohttp myself

Issue #937, opened 2019-05-09
Attach operation specification to connexion.request


hubertgrzeskowiak:
Flask does only rudimentary JSON deserialisation to primitives or dictionaries. With named types defined in the swagger spec (object declarations) we could deserialise the dicts to custom DTO classes (which also could be easily generated). However during decoding in `app.json_decoder` we only get a dictionary and there is no way of telling which type we should transform to. It would be great if connexion exposed some kind of information about the spec for the current view.

### Example code I played around with
```
class DataclassJSONDecoder(flask.json.JSONDecoder):
  def __init__(self, *args, **kwargs):
    kwargs["object_hook"] = self.map_to_object
    super(DataclassJSONDecoder, self).__init__(*args, **kwargs)

  def map_to_object(self, _dict):
    # What are the parameters for the current request?
	# To which type should we transform?
    pass


app = connexion.App(...)
app.app.json_decoder = DataclassJSONDecoder
```

### Actual behaviour
I couldn't find any such meta information on the request.

### Additional info:

- Python 3.7.3
- Connexion 1.1.15

rafaelcaricio:
Right, I agree. That info could be exposed to the views, somehow. At this moment I don't know how we should do that though.

frjonsen:
Was any progress made on this? I am attempting to implement role-based access by specifying allowed roles by adding `x-roles` in the specification, but need a way to actually read these either inside the controllers.

brockhaywood:
+1 I'm trying to find a way to get at the schema of the current operation because the defaults are provided in json body objects
re: https://github.com/zalando/connexion/issues/1184

brockhaywood:
Perhaps the same way that the request context is passed ?
https://github.com/zalando/connexion/blob/master/connexion/decorators/parameter.py#L118

frjonsen:
@brockhaywood We have since moved away from connexion to using FastAPI instead, but if it is of any help to you, this is how we solved this issue, specifically with our role based access functionality.

During the setup of the app, it looked like this. This is how we would get access to the OpenAPI specification at arbitrary points in the application.

Assume we had a specification which defines a path like this. Note the custom `x-roles` attribute.
```YAML
paths:
  /admin/grant-tokens:
    post:
      tags:
        - tokens
      x-roles:
        - admin
      summary: Grant pre-paid tokens
      description: >
        Generate and save pre-paid tokens to a user
      requestBody:
        required: true
        description: Information about the tokens and who to grant tokens to
[...]

PR #1281, opened 2020-08-13
adds pass_operation_arg_name to `parameter_to_arg`

brockhaywood:
Fixes https://github.com/zalando/connexion/issues/937

 - adds `pass_operation_arg_name` to allow passing the current operation to the related handler (similar approach to pass_context_arg_name)


rjkooijman:
Found this PR through issue #937. Since the PR is open since 2020, is there a reason this isn't merged yet?

RobbeSneyders:
@rjkooijman Happy to have this change in connexion, however I would move away from the `pass_operation_arg_name` and provide the operation under a fixed name. I'd like to do the same for `pass_context_arg_name` moving to v3.0.

Feel free to submit this to the `main` branch.

PR #1283, opened 2020-08-25
Fix: #878. Support OAS3 requestBody.required.

### LINKED ISSUE #878: requestBody: required: true doesn't work
### Description
Connexion 2.2.0
OpenAPI 3
```
requestBody:
  required: true
```
Connexion doesn't validate that a requestBody was provided.

### Expected behaviour

400 Bad request -> requestBody is required

### Actual behaviour

request is passed through to controller; KeyErrors abound because no requestBody was provided

### Steps to reproduce

Submit a POST with no requestBody

### Additional info:

Output of the commands:

```
Python 3.6.8
Version: 2.2.0
```

ioggstream:
Fixes #878 

 - pass `requestBody.required` to connexion
 - check if requestBody is not None when required

### DIFF
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -424,9 +424,14 @@ class AbstractOperation(SecureOperation):
                                      self.api,
                                      strict_validation=self.strict_validation)
         if self.body_schema:
+            try:
+                body_required = self.request_body['required']
+            except (AttributeError, KeyError):
+                body_required = False
             yield RequestBodyValidator(self.body_schema, self.consumes, self.api,
                                        is_nullable(self.body_definition),
-                                       strict_validation=self.strict_validation)
+                                       strict_validation=self.strict_validation,
+                                       body_required=body_required)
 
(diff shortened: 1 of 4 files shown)


Ruwann:
Hi @ioggstream , thanks for the PR, are you still interested in moving this forward? If so, as we have moved from `master` to `main`, could you rebase on `main`?

ioggstream:
@Ruwann yes, but I need some time to do it ... please wait.

connexion/operations/abstract.py
```diff
@@ -424,9 +424,14 @@ def __validation_decorators(self):
                                      self.api,
                                      strict_validation=self.strict_validation)
         if self.body_schema:
+            try:
```
RobbeSneyders, line 427: Since this is OpenAPI 3 specific, we should probably implement this in `OpenAPIOperation` instead of `AbstractOperation`.

RobbeSneyders:
Outdated.

PR #1290, opened 2020-09-07
Add support for multiple security schemes in AND fashion

### LINKED ISSUE #586: Support for multiple security requirements (x-tokenInfoFunc)
### Description
In security_decorator(), in case multiple security-requirements are defined, all of them are being ignored.

### Expected behaviour
1. Logical OR operation: Authenticate if at least one function succeeds
2. Logical AND operation: Authenticate if all functions succeeded

In any case the behavior should not be to ignore all security requirements

### Actual behaviour
All security requirements are ignored and security_passthrough() is applied

### Additional info:
security_decorator() on operation.py line 74

Ruwann:
Fixes #586 .

 - Add support for combining [multiple authentication schemes in AND fashion](https://swagger.io/docs/specification/authentication/#multiple). The `token_info` and `user` argument will differ as a dict mapping security scheme names to token info results for that scheme is now returned as `token_info`.

Multiple OAuth2 authentication schemes in AND fashion is not supported.

The diff is somewhat bigger than the actual changes because of an extra level of indentation. The changes can be summarized as:

- Iterate over `security_req.items()` instead of limiting it to 1 element (so we also have to change the `continue` to `break` for missing info funcs).
- If there is only a single item, we do the same as before and add it "as is" to the `auth_funcs` list
- If there are multiple items in security_req, we generate a `dict` mapping security scheme names to the security checking function. This dict is an input to `verify_multiple_schemes`, which is then added to `auth_funcs`

Except for 2 additional tests, these changes were previously reviewed in ml6team/connexion#1.


RobbeSneyders:
Thx Ruwan. LGTM!

RobbeSneyders (approved):

hjacobs:
Sorry, for the slow response, I think we can merge it after resolving the conflicts :smile:

hjacobs (approved):

PR #1293, opened 2020-09-11
Allow "Required: False" on Response headers - Fix #1261

### LINKED ISSUE #1261: Respect "required: false" specification on headers
### Description
Based on the OpenAPI 3, the "headers" key on a response, respects the `Map[string, Header Object | Reference Object]` interface, and checking the [Header Object](https://swagger.io/specification/#header-object) the description says:

> The Header Object follows the structure of the Parameter Object with the following changes:

Checking the [Parameter Object](https://swagger.io/specification/#parameter-object), one of the properties says:

> required | boolean | Determines whether this parameter is mandatory. If the parameter location is "path", this property is REQUIRED and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

But Connexion doesn't take into account this, and always force to be defined the headers on the response, even if it's optional (like for example when following the [RFC 8288](https://tools.ietf.org/html/rfc8288)).

### Expected behaviour
If a header is `required: false`, the lack of the header on the response shouldn't raise `NonConformingResponseHeaders`

### Actual behaviour
It raises `NonConformingResponseHeaders`

https://github.com/zalando/connexion/blob/master/connexion/decorators/response.py#L62

### Steps to reproduce

With this openapi schema

```
openapi: 3.0.0
info:
  title: Test
  version: 0.0.1
paths:
  /test:
    get:
      operationId: test
      responses:
        '200':
          description: Example headers
          content:
            application/json:
              schema:
                type: object
                required:
                 - data
                properties:
                  data:
                    type: string
                    nullable: true
                    example: null
          headers:
            link:
              schema:
                type: string
              required: false
```

With a response without `link` header, it will fail with:

```
{
  "detail": "Keys in header don't match response specification. Difference: link",
  "status": 500,
  "title": "Response headers do not conform to specification",
  "type": "about:blank"
}
```

### Additional info:

Output of the commands:

- `python --version`: 3.8.2
- `pip show connexion | grep "^Version\:"`: 2.7.0

miguelgf:
Fixes #1261 

 - Allows response headers to have `required: False` without `NonConformingResponseHeaders` error


miguelgf:
Hello @jmcs 

Any change this gets reviewed and merged? I would like to use in a project and we will need to fork the project to support our `required: false` on our openapi specs

Regards

connexion/decorators/response.py
```diff
@@ -52,9 +52,10 @@ def validate_response(self, data, status_code, headers, url):
 
         if response_definition and response_definition.get("headers"):
             # converting to set is needed to support python 2.7
-            response_definition_header_keys = set(response_definition.get("headers").keys())
+            required_header_keys = {k for (k, v) in response_definition.get("headers").items()
+                                    if 'required' not in v or v['required']}
```
Ruwann, line 56: you can simplify this by using `dict.get()`:
```
if v.get('required')
# or to be explicit:
if v.get('required', False)
```

Ruwann (changes requested):
Hi @miguelgf , thanks for the PR! We have moved from `master` to `main`, could you rebase your PR (there will probably be a conflict) ?

connexion/decorators/response.py
miguelgf: Great tip, changed

miguelgf:
@Ruwann rebased against `main` 👍

connexion/decorators/response.py
```diff
@@ -52,9 +52,9 @@ def validate_response(self, data, status_code, headers, url):
                 raise NonConformingResponseBody(message=str(e))
 
         if response_definition and response_definition.get("headers"):
-            response_definition_header_keys = response_definition.get("headers").keys()
-            header_keys = headers.keys()
-            missing_keys = response_definition_header_keys - header_keys
+            required_header_keys = {k for (k, v) in response_definition.get("headers").items()
+                                    if v.get('required', False)}
+            missing_keys = required_header_keys - headers.keys()
```
Ruwann, line 57: Seems like the tests are failing because of this line: apparently, `headers` (in the case of Flask) is actually a `werkzeug.datastructures.Headers` object instead of a regular Python `dict`. 😞 

Guess we'll have to change it back to the "python2.7-style" with sets:
```suggestion
            missing_keys = required_header_keys - set(headers.keys())
```

miguelgf: Applied the change

Ruwann:
It seems that your code was correct, but another test failed because, imo, the test itself is incorrect. 

For OpenAPI 3, it is clearly in the spec what the expected behaviour for optional and required headers is, but it is not very clear for Swagger 2. This issue (https://github.com/OAI/OpenAPI-Specification/issues/321) and this response (https://github.com/OAI/OpenAPI-Specification/issues/321#issuecomment-89958502) seem to indicate that that the response headers in Swagger 2 are optional.

In OpenAPI 3, when the `required` field is not present in the header, it should be considered `false`. So you can revert that last change.

If you are up for it, you can adjust the corresponding test and swagger/openapi. But I'm also happy to do it myself in a PR to your branch or in a follow-up PR.

https://github.com/zalando/connexion/blob/c014299435990ff1d4c20596396ca563f7b266b4/tests/api/test_headers.py#L20-L30

https://github.com/zalando/connexion/blob/c014299435990ff1d4c20596396ca563f7b266b4/tests/fixtures/simple/swagger.yaml#L559-L572

https://github.com/zalando/connexion/blob/c014299435990ff1d4c20596396ca563f7b266b4/tests/fixtures/simple/openapi.yaml#L704-L720

connexion/decorators/response.py
```diff
@@ -52,9 +52,10 @@ def validate_response(self, data, status_code, headers, url):
                 raise NonConformingResponseBody(message=str(e))
 
         if response_definition and response_definition.get("headers"):
-            response_definition_header_keys = response_definition.get("headers").keys()
-            header_keys = headers.keys()
-            missing_keys = response_definition_header_keys - header_keys
+            required_header_keys = {k for (k, v) in response_definition.get("headers").items()
+                                    if v.get("required", True)}
```
Ruwann, line 56:
```suggestion
                                    if v.get("required", False)}
```

miguelgf:
@Ruwann feel free to change the code, I won't be able to make any changes until late today. Thanks!

miguelgf:
@Ruwann I've reading what you sent and I see the issue, yes, the test is not testing the proper behaviour, but we can't test it on the Swagger 2, as there is no way to say that a header is required with the Header Object specification (https://swagger.io/specification/v2/#headerObject).

This test can be perform on openapi 3, on swagger 2 this test has no sense (every header is always optional, based on their specification).

We can change the test so it only pass on OpenAPI 3 and we don't pass it on Swagger 2, what do you think?

Ruwann (approved):

miguelgf:
Thanks @Ruwann and @hjacobs !!

PR #1298, opened 2020-09-23
Add support for array of files with same name in request

fgsalomon:
 - Add support for array of files with same name in request


RobbeSneyders:
Hi @fgsalomon could you provide some additional context with this change? Ideally in the form of a test that used to fail and now succeeds because of this change.

fgsalomon:
I'm sorry about the lack of tests. I needed this for an old project but have moved to FastAPI since then so I'm gonna discard the PR.

PR #1299, opened 2020-09-24
Pass through Werkzeug exception headers to problem

languitar:
Some Werkzeug exceptions such as Unauthorized define headers to present
to the client. Pass through these headers to the problem render method
to actually deliver those headers to the client.

Might potentially fix #1234


languitar:
Is there still any work going on on this project?

RobbeSneyders (approved):
Thx!

PR #1303, opened 2020-10-12
Fix reverseproxy example for OpenAPI 3

cuongqn:
In the current example, the run() method is called on the app variable, which does not have its path altered.
This small PR updates app to flask_app to fix that problem.


RobbeSneyders (approved):
Thx!

PR #1306, opened 2020-10-16
More liberal flask number converters for float and int in paths

### LINKED ISSUE #1040: path arguments of type number or integer can't be negative
The Swagger and OpenAPI specs don't say anything about arguments of type number or integer being restricted to non-negative values. However, passing a negative number in for a number or integer variable results in a `404 Not Found`.

This seems to be behaviour inherited from Flask's `int` and `float` converters which they inherited from Werkzeug. However, they added a `signed` flag to both of those converters in https://github.com/pallets/werkzeug/pull/1355 in December 2018 and that version seems to now be pulled in as a dependency by connexion

So now negative numbers can be supported by changing the `PATH_PARAMETER_CONVERTERS` lookup table in `flask_utils.py` to be:
```python
PATH_PARAMETER_CONVERTERS = {
    'integer': 'int(signed=True)',
    'number': 'float(signed=True)',
    'path': 'path'
}
```

This change won't be enough for us to change our strings to number since we would also need support for integers passed in for number parameters and so I won't set up an environment for making a pull request myself unless I can also fix #1041

### LINKED ISSUE #1041: path arguments of type number can't accept integers
The Swagger and OpenAPI specs don't say anything about arguments of type number being restricted to non-integer values. However, passing an integer in for a number or integer variable results in a 404 Not Found.

This seems to be behaviour inherited from Flask's `int` and `float` converters which they inherited from Werkzeug and reading their code, they still validate agains `regex = r"\d+\.\d+"`

This is the less tractable sister of #1040 since it requires the introduction of a flag in Werkzeug's `FloatConverter` class to relax the regular expression above.

logi:
This change installs more liberal flask converters for `float` and `int`. These don't try to enforce a "single representation" of paths but instead try to convert the numbers that callers pass in.

This is the minimal change to get this effect and installs the custom number converters where the custom JSON converter is being installed already. An alternative would be to register the new converters as `integer` and `number` in Flask and update the `PATH_PARAMETER_CONVERTERS` mapping, but that's a larger change and would change the generated flask paths which could affect existing code.

Fixes #1040
Fixes #1041

The tests for numbers in paths are modified to return not just `int` or `float` but also the value seen by the handler so we can ascertain that the conversion works and not just the routing. Then a few number formats are added.

### DIFF
--- a/tests/api/test_parameters.py
+++ b/tests/api/test_parameters.py
@@ -140,7 +140,7 @@ def test_strict_formdata_param(strict_app):
 def test_path_parameter_someint(simple_app, arg, result):
     assert isinstance(arg, str)  # sanity check
     app_client = simple_app.app.test_client()
-    resp = app_client.get('/v1.0/test-int-path/'+arg)  # type: flask.Response
+    resp = app_client.get(f'/v1.0/test-int-path/{arg}')  # type: flask.Response
     assert resp.data.decode('utf-8', 'replace') == f'"{result}"\n'
 
 
@@ -170,7 +170,7 @@ def test_path_parameter_someint__bad(simple_app):
 def test_path_parameter_somefloat(simple_app, arg, result):
     assert isinstance(arg, str)  # sanity check
     app_client = simple_app.app.test_client()
-    resp = app_client.get('/v1.0/test-float-path/' + arg)  # type: flask.Response
+    resp = app_client.get(f'/v1.0/test-float-path/{arg}')  # type: flask.Response
     assert resp.data.decode('utf-8', 'replace') == f'"{result}"\n'
 
 
--- a/tests/api/test_parameters.py
+++ b/tests/api/test_parameters.py
@@ -1,6 +1,8 @@
 import json
 from io import BytesIO
 
+import pytest
+
 
 def test_parameter_validation(simple_app):
     app_client = simple_app.app.test_client()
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Hi @logi, thanks for this PR. We switched to the `main` branch, could you rebase / merge?

logi:
@RobbeSneyders AFAICT that's done now

tests/fakeapi/hello/__init__.py
```diff
@@ -266,11 +266,11 @@ def test_schema_int(test_int):
 
 
 def test_get_someint(someint):
-    return type(someint).__name__
+    return '%s %g' % (type(someint).__name__, someint)
```
hjacobs, line 269: Please use "new" Python f-strings (`f"{x} {y}")

tests/fakeapi/hello/__init__.py
```diff
@@ -266,11 +266,11 @@ def test_schema_int(test_int):
...
 
 
 def test_get_somefloat(somefloat):
-    return type(somefloat).__name__
+    return '%s %g' % (type(somefloat).__name__, somefloat)
```
hjacobs, line 273: Please use "new" Python f-strings (`f"{x} {y}"`)

tests/api/test_parameters.py
```diff
@@ -124,22 +126,57 @@ def test_strict_formdata_param(strict_app):
...
+    ['+0', 'int 0'],
+    ['-0', 'int 0'],
+    ['-123', 'int -123'],
+])
+def test_path_parameter_someint(simple_app, arg, result):
+    assert isinstance(arg, str)  # sanity check
     app_client = simple_app.app.test_client()
-    resp = app_client.get('/v1.0/test-int-path/123')  # type: flask.Response
-    assert resp.data.decode('utf-8', 'replace') == '"int"\n'
+    resp = app_client.get('/v1.0/test-int-path/'+arg)  # type: flask.Response
```
hjacobs, line 143:
```suggestion
    resp = app_client.get(f"/v1.0/test-int-path/{arg}")  # type: flask.Response
```

tests/fakeapi/hello/__init__.py
logi: Sure. When I submitted this PR 9 months ago I reviewed your code style and there were no f-strings around so I carefully converted all my f-strings into old-school % formats.

hjacobs: True, we are still catching up with Python features and have now removed some of the old-school stuff (probably a lot of cleaning up still needed).

logi: Yeah, been there.

hjacobs (approved):

hjacobs:
:+1:

Issue #1309, opened 2020-10-22
MethodViewResolver initializes view object for every method which uses it


exaroth:
### Description
I noticed an unexpected behaviour when using `MethodViewResolver` https://github.com/zalando/connexion/blob/master/connexion/resolver.py#L138 - namely that  for every method defined against a method view connexion will instantiate separate view object.

### Expected behaviour

There should be only single class instantiated for each `add_api` call with `MethodViewResolver` 

### Actual behaviour

Resolver will initialize separate View object for each method defined in the spec

### Steps to reproduce

Given example swagger spec:

``` yaml
---
openapi: "3.0.0"
info:
  title: "Test API"
  description: "Test api"
  version: "1.0.0"

paths:
  /recipes:
    post:
      responses:
          "200":
             description: "post"
    delete:
      responses:
          "200":
             description: "delete"
```

example `MethodView`:

``` python
from flask.views import MethodView


class RecipesView(MethodView):

    def __init__(self):
        print("------> initializing ", self.__class__)

    def delete(self):
        pass

    def post(self):
        pass
```

and some boilerplate:
``` python
import connexion
from connexion.resolver import MethodViewResolver

connexion = connexion.FlaskApp(__name__, skip_error_handlers=True)
connexion.add_api('swagger/spec.yml', resolver=MethodViewResolver("service.api"))
app = connexion.app

```

This will output:

```
------> initializing  <class 'service.api.RecipesView'>
------> initializing  <class 'service.api.RecipesView'>
 * Serving Flask app "service" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
```

### Additional info:

Output of the commands:

- `python --version`
`Python 3.8.0`
- `pip show connexion | grep "^Version\:"`
`Version: 2.7.0`

PR #1310, opened 2020-10-22
Fix issue with MethodViewResolver instantiating multiple instances of a view

exaroth:
Fixes  https://github.com/zalando/connexion/issues/1309

This fixes issue I reported earlier today, which results in `MethodViewResolver` instantiating separate View object for each method defined in spec file.

My proposal is to maintain an array of view objects initialized by the resolver and before instantiating each view class check if it was instantiated before (by comparing against `__class__` attribute of instantiated view) - if so reuse it, if not create new object and add it to the list.


RobbeSneyders (approved):
Thx!

PR #1326, opened 2020-11-22
Send request exception signal in common handler

Agalin:
Flask's default error handler sends `got_request_exception` signal (see [this method](https://github.com/pallets/flask/blob/dc11cdb4a4627b9f8c79e47e39aa7e1357151896/src/flask/app.py#L1446)). Handler provided by `connexion` doesn't do that and breaks integration of other apps, e.g. Sentry (see [Flask integration](https://github.com/getsentry/sentry-python/blob/e873bdb071146b1fd31814ae5f742f6a4f7abe39/sentry_sdk/integrations/flask.py#L80) implementation).

 - Send `got_request_exception` signal in `common_exception_handler` registered for Flask application.
 - Due to `got_request_exception` expecting an instance `common_exception_hanndler` is changed from static method to a normal instance method.


RobbeSneyders (approved):
Thx @Agalin!

PR #1335, opened 2021-01-11
Fix response validator docstrings

Glutexo:
The default value of _ResponseBodyValidator.validator_ is not _jsonschema.Draft4Validator_ for some time already. It is _Draft4ResponseValidator_ now.

_ResponseValidator.validator_ does not have a default value. If not set, _ResponseBodyValidator_ default is used, which is eventually _Draft4ResponseValidator_, but it is not _ResponseValidator_ that decides it.


Glutexo:
Travis is failing with

> The job exceeded the maximum log length, and has been terminated.

This shouldn’t be related to my changes.

hjacobs:
:+1:

PR #1344, opened 2021-03-06
Limit `openapi-spec-validator`

### LINKED ISSUE #1343: Connexion incompatible with openapi-spec-validator==0.3.0
### Description
Connexion is incompatible with the latest version of the openapi-spec-validator (0.3.0 minimum).
This line: https://github.com/zalando/connexion/blob/1ed14fbde5be3d144760fec4982e90064f151e88/connexion/decorators/validation.py#L169
forces users to define a `formData` type parameters for each one of the attributes in the form data body (`multipart/form-data` content type).
This is not allowed by the spec, and as a result, an error is raised by 'openapi-spec-validator==0.3.0' upon starting the flask app.
Removal of the incorrect parameters lead to validation errors on request in Connexion (in strict mode), because connexion requires the parameters to be specified incorrectly in the spec file.

### Expected behaviour
With no `formData` parameters present in the spec, no error should be raised upon validating an incoming request with form data payload.

### Actual behaviour
With no `formData` parameters in the spec, an error is received upon validating incoming requests in strict mode:
`Extra formData parameter(s) user_id not in spec`, with status code 400

### Steps to reproduce
Install connexion==2.4.0
Using the following spec:
```
---                                                     
openapi: 3.0.0                                          
info:                                                   
  title: Premium API                                    
  version: 1.0.0                                        
paths:                                                  
  /{profile_id}/attachments:                            
    post:                                               
      description: Upload attachment to profile         
      operationId: some operation                       
      parameters:                                       
       - $ref: '#/components/parameters/ProfileIDParam' 
       - $ref: '#/components/parameters/UserIDParam'    
      requestBody:                                      
        required: true                                  
        content:                                        
          multipart/form-data:                          
            schema:                                     
              type: object                              
              properties:                               
                user_id:                                
                  type: string                          
                file:                                   
                  type: string                          
                  format: binary                        
      responses:                                        
        200:                                            
          description: OK                               
schemas:                                                
  UserIDParam:                                          
    name: user_id                                       
    in: formData                                        
    required: false                                     
    description: userid                                 
    schema:                                             
      type: string                                      
                                                        
  ProfileIDParam:                                       
    name: profile_id                                    
    in: path                                            
    description: ID                                     
    required: true                                      
    schema:                                             
      type: string                                      
```
Start the server, and observe the error (from `openapi-spec-validator`) that states that UserIdParam does not validate (1in: formData1 is incorrect in the standard.
Then remove the UserIDParam parameter, start the server again, and try calling the api endpoint. With strict validation, you will get a 400 error saying that the `formData` parameter is missing.

### Additional info:

Output of the commands:

- `python --version` 3.7.9
- `pip show connexion | grep "^Version\:"` 2.4.0

briankinney:
Recent 0.3.0 release is not compatible with connexion in its current state. This limits to the latest compatible release.

Fixes #1343  .

 - Limit `openapi-spec-validator` dependency to <=0.2.10

### DIFF
--- a/requirements-all.txt
+++ b/requirements-all.txt
@@ -1,4 +1,4 @@
 -r requirements-aiohttp.txt
 -r requirements-devel.txt
-openapi_spec_validator
+openapi_spec_validator<=0.2.10
 pytest-aiohttp
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ install_requires = [
     'PyYAML>=5.1',
     'requests>=2.9.1',
     'inflection>=0.3.1',
-    'openapi-spec-validator>=0.2.4',
+    'openapi-spec-validator>=0.2.4,<=0.2.10',
 ]
 
 swagger_ui_require = 'swagger-ui-bundle>=0.0.2'


briankinney:
Looking at the Travis logs, it looks like the tests pass in the one environment where the new version specification have resulted in the newer release not being used. I think that the problem is due to some details of tox/travis/whatever other tools you are using for CI. I'm going to have to leave it here and let someone more familiar with the CI stack to figure out why the requirements are not installed correctly.

briankinney:
Perhaps the issue was adding the version limit to the `setup.py`. Tests still fail though. I don't know why.

RobbeSneyders:
Closing in favor of #1374, which fixes the underlying issue.

PR #1348, opened 2021-03-11
Fix RuntimeError exception from **application/x-www-form-urlencoded**  (see also PR# 1222)

RichardBruskiewich:
Adds to PR #1222  to fix a RuntimeError exception from **application/x-www-form-urlencoded**  processes here, as well as the **multipart/form-data** issues already addressed by the PR.

 - https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L162 response.body is a byte string from an **application/x-www-form-urlencoded** form **POST**, thus, is not a dictionary with an `update` method (next line). Parsing the byte string out creates a dictionary which is better behaved (coded a local method, `parse_body_parameters(self, body)` , to achieve this). 
 - Note that if the value of the parameters in the URL encoded byte string are singular values (not true arrays of values), then the dictionary value is simplified to a scalar.
 -

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ htmlcov/
 .vscode/
 venv/
 src/
+venv/
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -2,6 +2,8 @@ import collections
 import copy
 import functools
 import logging
+from urllib.parse import parse_qs
+
 
 import pkg_resources
 from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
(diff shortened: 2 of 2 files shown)


RichardBruskiewich:
I'm getting an odd RuntimeError with this patch. 

```
Error handling request
   File "/usr/src/app/src/connexion/connexion/apis/aiohttp_api.py", line 49, in problems_middleware
     response = await handler(request)
   File "/usr/local/lib/python3.8/site-packages/aiohttp/web_middlewares.py", line 110, in impl
     return await handler(request)
   File "/usr/src/app/src/connexion/connexion/decorators/coroutine_wrappers.py", line 21, in wrapper
     connexion_response = function(connexion_request)
   File "/usr/src/app/src/connexion/connexion/decorators/uri_parsing.py", line 144, in wrapper
     response = function(request)
   File "/usr/src/app/src/connexion/connexion/decorators/validation.py", line 177, in wrapper
     (self.parse_body_parameters(request.body) if len(request.body) > 0 else {})
   File "/usr/src/app/src/connexion/connexion/decorators/validation.py", line 131, in parse_body_parameters
     parsed_body = parse_qs(body.decode(encoding))
 UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 285: invalid start byte
```

Alas, although my application seems to be returning a byte-based request body, decoding it as UTF-8 in my application doesn't seem to work.  I'm unsure about the specific source of this error. If anyone has an idea, please let me know.

Thus, I've converted this PR to 'draft' pending arriving at a better understanding as to what is going on.

RobbeSneyders:
Closing due to inactivity.
Might be outdated since #1491

Issue #1031, opened 2019-08-31
Is OpenAPI 3.0.0 supported by connexion?


mcocosila:
### Description
I'm using OpenAPI 3.0.0 in my spec yaml file.

I keep getting errors like 'no module found' or 'empty function name'. I moved around the python file and openapi spec file trying to work around the operationId not being resolved correctly.

Sample:
File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/connexion/resolver.py", line 61, in resolve_function_from_operation_id
    return self.function_resolver(operation_id)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/connexion/utils.py", line 75, in get_function_from_name
    raise ValueError("**Empty function name**")

### Expected behaviour
operationId to be resolved correctly

### Actual behaviour
Errors like 'no module found' or 'empty function name'.

### Steps to reproduce
python file named api.py located at project top level
function named hello in api.py python module

in openapi spec file:
operationId: api.hello

### Additional info:

Output of the commands:

- `python --version`
Python 3.7.4

- `pip show connexion | grep "^Version\:"`
Version: 2.3.0

danie1k:
Take a look on #983 & #996

danie1k:
You cannod define view functions (operations) in the same file where you define Connextion API. It won't work.
There are some cyclic imports that take place or something like that.

Just move your functon `hello` to any other py file.

Abdullahabid3691:
sharing some related content of your .yml file would be helpful in order to trace the issue.
This can happen when you define `x-swagger-router-controller` instead of `x-openapi-router-controller` (in case of OAS3)
or the `operationId` parameter is not set correctly.

RobbeSneyders:
OpenAPI 3.0.0 is supported by Connexion.

PR #1353, opened 2021-04-20
Add note about x-openapi-router-controller when using OAS

paulsbruce:

 - Add a note about x-openapi-router-controller when using OAS

Fantastic work, this repo is. There are quite a few confusions out on the interwebs about the RestyResolver, one of which I stumbled on when using OpenAPI 3. A simple docs update should help others not spin their wheels on this like I did.


paulsbruce:
See this comment for where I finally found the hint from Googling a lot: https://github.com/zalando/connexion/issues/1031#issuecomment-668771368

hjacobs:
:+1:

PR #1358, opened 2021-05-31
Update exceptions.rst

GabrieleLabanca:
Added import line for Request in example code.

Probably obvious for most people, but for those not familiar with flask (like me) it can be a minor hassle.


hjacobs:
:+1:

PR #1360, opened 2021-06-03
schema validation error message creation moved to its own function

scgbckbone:
Fixes # .
Nothing. This is just a refactor. 

Rationale:
I use connexion to validate JSON requests for me. While I validate YAML files with josnschema. To match error response I copy your code to our codebase - it would be nice if I'm able to import "error message creation routine". Not sure whether classmethod is right place to do it - yet at least with this PR we can start a discussion.

 - schema validation error message creation moved to its own function (RequestBodyValidator)


RobbeSneyders:
Thx for the PR @scgbckbone.

This refactoring seems fine to me, but I would mark the method private (rename to `_error_path_message`). I don't think it's beneficial for future Connexion development to expand the public interface.

Also, we moved to the `main` branch, so can you please rebase and change the target of your PR?

scgbckbone:
incorporated @RobbeSneyders comments

RobbeSneyders (approved):
Thanks!

Issue #1365, opened 2021-06-21
Moving Connexion forward


RobbeSneyders:
Hi all,

Like a lot of other people here, we are convinced of the benefits of Connexion but worried by about the state of maintenance of the project (#1262, #1334). After a comparison with other available and popular tools, we decided that Connexion is still the best framework for our needs, as long as it is actively maintained. Since we're not aware of any existing maintained fork, we want to commit to help maintain the project ourselves.

We have created a [fork](https://github.com/ml6team/connexion) and are actively working on adding our own improvements. So far we have:
- Added support for multiple authentication methods (ml6team/connexion#1)
- Moved CI/CD to github actions so the tests can run on our fork
- Fixed the tests and added version upper bounds for breaking dependencies
- Added high level documentation (ml6team/connexion#6)

Our goal for the project is two-fold
- Provide regular releases with continuous improvements and bug fixes
- Continue the work towards a framework agnostic Connexion 3.0 with support for frameworks such as quart, sanic & starlette (#380, #496, #1250, #828, #1095, Ruwann/connexion#3)

The goal of this fork is not to do a hostile takeover, but to ensure that Connexion and its API-first approach can stay alive and be extended to other frameworks. We created the new fork to start moving forward again, but are convinced that it would be better for the project to be able to keep its home and name. @hjacobs @jmcs @rafaelcaricio @dtkav, if you read this, please reach out to discuss any possibilities of collaboration on the current repository instead.

If there is no possibility to keep working on the current repository, we'll open discussions on how to move forward with the fork in the coming weeks. So make sure to star / watch the fork or this thread to stay up to date. In the meantime, feel free to already submit any issues or pull requests on the fork as well.

@RobbeSneyders @Ruwann @svetlana-v

dtkav:
Awesome! I don't have write permissions to the repo (I was also a community contributor), so I can't help you. I might be interested in contributing to the fork! I have some open PRs I'm excited about.

IMO the ideal course of action would be for Zalando to donate the project / branding. That way they aren't legally bound to review every PR when they don't have the resources to do so.

Best of luck to you!

rafaelcaricio:
I'm in the same position as @dtkav . When I worked at Zalando I used to have permissions to the repository. In the last 4 years I didn't have any access. I need someone from Zalando to merge PRs, approve, etc.

I agree that it would be best for the project to move to its own organization in Github. But that is my personal opinion.

I hope this can be fixed.

hjacobs:
[...]

PR #1367, opened 2021-06-25
Move CICD from Travis to Github Actions

RobbeSneyders:
This PR moves CICD from Travis to Github Actions as discussed in #1365.

It also fixes some issues currently breaking the pipeline
- Fix test API configuration to comply with updated spec validation
- Fix isort commands & order for newest version
- Add upper limits to dependencies to prevent breaking versions

These changes are a combination of the following PRs on the ML6 fork with some small adaptations:
- ml6team/connexion#2
- ml6team/connexion#5


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install tox tox-gh-actions setuptools coveralls
```
rafaelcaricio, line 22: I think would be nice if we considered to pin those versions. Those packages may release major versions with breaking changes that can break Connexion CI/CD pipeline in an unfortunate moment. I know we have more maintainers now, but if we have multiple PRs, we need to coordinate with external contributors to ask them to rebase their PRs or commit to each PR to fix the pipeline. It might be wise to just pin the versions here and avoid the headache. We can still update the versions whenever we need, of course.

rafaelcaricio (approved):
LGTM. I just have one suggestion.

hjacobs:
:+1:

hjacobs (approved):

PR #1371, opened 2021-06-25
Mark endpoints for streaming uploads

### LINKED ISSUE #592: Handling streaming request data
### Description

Using just Flask, a user can (manually) use the `request.input_stream` to handle streaming uploads. The use case is because I need to handle binary uploads that won't fit in memory. I use the JSON validation features of `connexion` when the Content-Type is `application/json` but in this case I have a `application/octet-stream`.

With the `connexion` library I can't seem to work out how to access the raw data stream - it looks like the data all gets read in  [FlaskApi.get_request](https://github.com/zalando/connexion/blob/master/connexion/apis/flask_api.py#L226) from `EndOfRequestLifecycleDecorator` - which is happening well before calling any user defined handler function.

Is it possible to not wrap the `flask.request` with `ConnexionRequest` for some endpoints?

### Expected behaviour

For an unrecognized content-type, or for a specific binary stream, connexion should - like Flask - not `read()` the full request body before calling a handler. Instead the stream should be pulled on demand - usually when a user calls `request.data()` or `request.json()`

### Actual behaviour

Deep in the bowels of connexion the stream gets fully read and I couldn't see how to disable it for one endpoint.

### Steps to reproduce

Upload a file e.g. with `requests`:
```python
file_path = 'large.bin'
with open(file_path, 'rb') as f:
    r = requests.post(
        url + '/upload',
        headers={
            'Content-Type': 'application/octet-stream',
            'Content-Length': '123456789'
        },
        data=f.read()
    )
```

On the server side:

```python
if headers['Content-Type'] == "application/octet-stream":
        # This should work
        data = request.input_stream.read()
        print(len(data))
```

But the input_stream is empty because connexion has already slurped it up - which lead to problems if we are dealing with uploads that won't fit in memory (obviously I don't normally just read the stream and print the length)

### Additional info:

- `python --version` 3.7 (in alpine docker image)
- `pip show connexion | grep "^Version\:"` 1.4

### LINKED ISSUE #1332: Trying to upload large files into chunks using multipart/form-data but connexion.request.stream seems to be empty
### Description

I'm trying to implement a POST request to upload large files (~4GB) using connexion set up with a Flask server. The physical is an embedded one so the performances are critical.

Having read the [OpenAPI 3.0 File Upload specs](https://swagger.io/docs/specification/describing-request-body/file-upload/), it suggests to use `multipart/form-data` to upload a file with additional data (e.g filename, user ID, ...). Here is the API I'm using :

**api.yaml**
```yaml
paths:
  /records:
    post:
      tags:
          - records
      summary: Upload a new record to the server
      operationId: server.controllers.file_controller.records_post
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              additionalProperties: false
              type: object
              properties:
                record:
                  type: string
                  format: binary
                  description: Record to be uploaded to the server
                filename:
                  type: string
                  description: New filename for the record being uploaded
                  required:
                      - record
              required:
                  - record
      responses:
        201:
          description: Created, the new record has been successfully uploaded to the server
        409:
          description: Conflict, a record already exists with that name on the server
```

**file_controller.py**
```python
def records_post():
    # Create destination directory if it does not exist
    os.makedirs(os.path.dirname(RECORDS_PATH), exist_ok=True)

    record = connexion.request.files['record']
    filename = secure_filename(record.filename)

    if 'filename' in connexion.request.form and connexion.request.form['filename']:
        filename = secure_filename(connexion.request.form['filename'])

    filepath = os.path.join(RECORDS_PATH, filename)

    if os.path.isfile(filepath):
        return {'status': 'Conflict', 'message': 'A record already exists with that name on the server'}, 409

    try:
        record.save(filepath)
    except Exception as e:
        logging.error("Failed to save '{}' to: {}".format(filename, RECORDS_PATH))
        logging.error(str(e))
        return {'status': 'Internal Server Error', 'message': 'Failed to upload record to the server',
                'error': str(e)}, 500

    return {'status': 'Created', 'message': 'Record uploaded successfully !', 'location': filename}, 201
```

When using `multipart/form-data` I can reach to the uploaded file using `connexion.request.files['record']` but it seems that it has to save the file into disk when file size exceeds 500KB and I'm limited to the max available space in `/tmp` on my server (which is ~1.8GB).

Some people in the issues suggest to use [streaming-form-data](https://github.com/siddhantgoel/streaming-form-data), so I've taken a look at [an implementation for Flask server](https://git.io/fjPoA) but it seems that `connexion.request.stream` is empty. Apparently, [the stream would be empty if it was previously read by request.data](https://stackoverflow.com/a/11000778/4200279)... and if I use `request.data` or `request.get_data()`, I can't find a way to split the bytestring into chunks to save memory and write to the right destination.

If I use `application/octet-stream`, I can stream the file directly but I'm loosing any chance to upload additional data and I encounter the same problem as above which is: `connexion.request.stream` is empty anyway and I can't read `connexion.request.get_data()` into chunks.

Is there any way to read non-empty `connexion.request.stream` or read `connexion.request.get_data()` into chunks ?

### Additional info:

Output of the commands:

- `python --version` **3.8.5**
- `pip show connexion | grep "^Version\:"` **2.7.0**
- OpenAPI version : **3.0.0**

tst-lsavoie:
Fixes #592 based on the suggestion by @MatteoRagni: https://github.com/zalando/connexion/issues/592#issuecomment-775153162.
Also Fixes #1332, I think.

 - Adds an `x-stream-upload` option for requests. When this is specified, connexion will pass `flask.request.stream` as the `request.body` instead of calling `get_data()`. This allows streaming uploads to disk instead of reading the entire file into memory.


RobbeSneyders:
Hi @tst-lsavoie, thanks for the PR.

- I'm a bit hesitant to define a custom extension. Is there a reason we cannot inspect `Content-Type` instead?
- We moved to the `main` branch. Can you rebase or merge and change the target branch of the PR?

tst-lsavoie:
@RobbeSneyders I didn't use the `Content-Type` because it's less reliable since a client could leave the content type out or set it incorrectly. There's also the question of which content types should be streamed, which might be different for each application, or even different for different endpoints in the same application. I don't love the idea of adding a new extension either but it seemed like the least bad option.

I changed the base branch to `main` but it looks like there are non-trivial merge conflicts I need to figure out. I'm going to hold off on that until we have a clearer picture of how the PR will need to change.

RobbeSneyders:
I've been looking into this a bit more, and I would suggest the following solution.

Define a `body` property with `body_getter` on `ConnexionRequest`, similar to the `json` property with `json_getter`.

https://github.com/zalando/connexion/blob/f46551cee8056dca65c1177e8252f19805d95d50/connexion/lifecycle.py#L25-L27

We can then assign `request.get_data` as the `boddy_getter` for Flask, which would delay stream consumption until the `body` property is evaluated. There's only two places in Connexion where `request.body` is evaluated:
1) During request body validation. Here the body is only consumed if Connexion supports validation for the provided `Content-Type`, which is currently only for `json` or `form`:
https://github.com/zalando/connexion/blob/f46551cee8056dca65c1177e8252f19805d95d50/connexion/decorators/validation.py#L137-L190

2) During parameter parsing. Here the body is currently always consumed:
https://github.com/zalando/connexion/blob/f46551cee8056dca65c1177e8252f19805d95d50/connexion/decorators/parameter.py#L79-L84
But it's actually not used if the body parameter is not defined in the view function parameters:
https://github.com/zalando/connexion/blob/f46551cee8056dca65c1177e8252f19805d95d50/connexion/operations/swagger2.py#L277-L282
If we refactor the parameter parsing a bit, we can prevent body consumption here as well if the body parameter is not present in the view function parameters.

You can then prevent stream consumption in connexion by
- Defining a `Content-Type` for which Connexion does not support validation, this should be the case for all streaming types
- Not defining a body parameter in the view function

WDYT?

RobbeSneyders:
You can find a quick and dirty PoC of the solution here:
https://github.com/thermopylae/connexion/compare/mark-endpoints-for-streaming-uploads...zalando:streaming-uploads-poc

tst-lsavoie:
I think I understand the proposed solution. So, if the content type was not form or JSON, and if the user didn't add a parameter for the body to their function, then the stream would be available.

I could see that working. I worry about it for the future, though. One small, innocent looking update that accesses the request body could break it. Hopefully testing would catch that, but it's not guaranteed.

We'd have to do some refactoring for the OpenAPI 3 version of the code as well:
https://github.com/zalando/connexion/blob/c014299435990ff1d4c20596396ca563f7b266b4/connexion/operations/openapi.py#L270-L300

The biggest problem is going to be time. I'm not sure when I'll next have a good chunk of time to focus on this issue.

clementperon:
This is great! 

I have used it marking my endpoint with x-stream-upload and work like a charm, thanks :)

RobbeSneyders:
Fixed by https://github.com/spec-first/connexion/pull/1618/files

PR #1374, opened 2021-06-28
Fix formdata parameter validation for OAS3

### LINKED ISSUE #1020: Connexion claims extra formData parameters that are in the spec

### LINKED ISSUE #1160: Extra formData parameter(s) query_param not in spec
### Description

using Connexion 2.4.0: if I use path, query and payload parameters with strict_validation=True in a POST route, I get an input validation error. This does work in version 2.3.0

### Expected behaviour

```
curl -X POST "http://localhost:6060/example/3212?query_param=42" -H "accept: application/json" -H "Content-Type: application/json" -d "{ \"data\": \"string\"}"
```

I'm simply returning the path, query and payload parameters all in a single response body.
the response should be:

```
{
  "path": "3212",
  "payload": {
    "data": "string"
  },
  "query": 42
}

```
### Actual behaviour

400 Error: BAD REQUEST

```
{
  "detail": "Extra formData parameter(s) query_param not in spec",
  "status": 400,
  "title": null,
  "type": "about:blank"
}
```

### Steps to reproduce

I'm using this swagger.yaml

```
swagger: "2.0"
info:
  title: Bugfix example API
  description: There's a bug in Connexion 2.4.0 in the POST route if you use path, query and payload parameters with strict_validation=True
  version: 1.0.0
tags:
  - name: Example
    description: This causes validation errors if you use Connexion 2.4.0 (it does work in 2.3.0)
consumes:
  - application/json
produces:
  - application/json
paths:
  /example/{path_param}:
    post:
      operationId: app.flask.routes.Example.post
      tags:
        - Example
      summary: Example endpoint to post information.
      parameters:
        - name: path_param
          in: path
          description: Example path parameter
          required: true
          type: string
        - name: query_param
          in: query
          description: Example query parameter
          required: true
          type: integer
          default: 42
        - description: Example payload
          in: body
          name: payload
          required: true
          schema:
            $ref: "#/definitions/payload"
      responses:
        201:
          description: OK
definitions:
  payload:
    type: object
    properties:
      data:
        description: Test data
        type: string
```

It works with version 2.3.0, but not with 2.4.0

### Additional info:

Output of the commands:

- `python --version`: `Python 3.7.5`
- `pip show connexion | grep "^Version\:"`:  `Version: 2.4.0`

### LINKED ISSUE #1340: parameter validator in connexion module expecting formData key which we never use in open api 3
### Description

i am working on open api 3. For my PUT request content-type is  set to multipart/form-data and also I have path parameter too. I enabled strict_validation for validating parameters. Parameter validator in connexion/decorators/validation.py is validating form parameters but in open api 3 we are not defining formData parameters like this( in:formData) anymore we are using request body only for media types.

def validate_formdata_parameter_list(self, request):
    request_params = request.form.keys()
   
    try:
        spec_params = [x['name'] for x in self.parameters['formData']]
    except KeyError:
        # OAS 3
        return set()
    return validate_parameter_list(request_params, spec_params)

we are using only in:path and in:query in openapi 3 so no need of checking for  formData parameters in parameter validator class so my suggestion is to remove validate_formdata_parameter_list function from parameter validator

### Additional info:

Output of the commands:

- `python --version`

### LINKED ISSUE #1343: Connexion incompatible with openapi-spec-validator==0.3.0

RobbeSneyders:
Fixes #1020. Fixes #1160. Fixes #1340. Fixes #1343.

The previous fix which tries to catch a `KeyError` does not work since `self.parameters` is a `defaultdict`. Changed to an explicit `if` check instead.

### DIFF
--- a/tests/decorators/test_validation.py
+++ b/tests/decorators/test_validation.py
@@ -1,6 +1,7 @@
 from unittest.mock import MagicMock
 
 import pytest
+from connexion.apis.flask_api import FlaskApi
 from connexion.decorators.validation import ParameterValidator
 from connexion.json_schema import (Draft4RequestValidator,
                                    Draft4ResponseValidator)
@@ -180,3 +181,21 @@ def test_writeonly_required_error():
     }
     with pytest.raises(ValidationError):
         Draft4RequestValidator(schema).validate({"bar": "baz"})
+
+
+def test_formdata_extra_parameter_strict():
+    """Tests that connexion handles explicitly defined formData parameters well across Swagger 2
+    and OpenApi 3. In Swagger 2, any formData parameter should be defined explicitly, while in
+    OpenAPI 3 this is not allowed. See issues #1020 #1160 #1340 #1343."""
+    request = MagicMock(form={'param': 'value', 'extra_param': 'extra_value'})
+
+    # OAS3
+    validator = ParameterValidator([], FlaskApi, strict_validation=True)
+    errors = validator.validate_formdata_parameter_list(request)
+    assert not errors
+
+    # Swagger 2
+    validator = ParameterValidator([{'in': 'formData', 'name': 'param'}], FlaskApi,
+                                   strict_validation=True)
+    errors = validator.validate_formdata_parameter_list(request)
+    assert errors
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -305,10 +305,9 @@ class ParameterValidator(object):
 
     def validate_formdata_parameter_list(self, request):
         request_params = request.form.keys()
-        try:
+        if 'formData' in self.parameters:  # Swagger 2:
             spec_params = [x['name'] for x in self.parameters['formData']]
-        except KeyError:
-            # OAS 3
+        else:  # OAS 3
             return set()
         return validate_parameter_list(request_params, spec_params)
 


jmcs (approved):

tests/decorators/test_validation.py
```diff
@@ -1,7 +1,9 @@
 from unittest.mock import MagicMock
 
 import pytest
+from connexion.apis.flask_api import FlaskApi
 from connexion.decorators.validation import ParameterValidator
+from connexion.exceptions import ExtraParameterProblem
```
Ruwann, line 6: Why is this import now needed? `ExtraParameterProblem` doesn't seem to be used in this file.

Ruwann (approved):

tests/decorators/test_validation.py
RobbeSneyders: Good catch. This is left over from a different implementation of the test I tried first.

PR #1375, opened 2021-07-02
Bugfix/missing infofuncs

Ruwann:
In PR #869 , the check for `token_info` was changed from `if token_info is None` to `if token_info is cls.no_value`. However, the initial value for `token_info` is `None`, so if there are no auth_funcs (due to missing `x-XXXInfoFunc`s), the `OAuthProblem` branch isn't entered and `token_info.get()` will raise an AttributeError, leading to a 500 Internal Server Error.

 - Change the default value of `token_info` from `None` to `cls.no_value` so that an `OAuthProblem` is raised


hjacobs:
:+1:

PR #1376, opened 2021-07-03
Remove deprecated static analysis

RobbeSneyders:
This PR removes a bunch of deprecated static analysis config files:

- The Catwatch repository has been archived
- The Codevalidator repository has been archived
- Coverage: Python 2 support has been dropped and the remaining Python 3 config is the same as the default config
- No idea which tool relies on `.checkignore`, but it doesn't seem to break anything
- @hjacobs can you confirm that it's ok to remove the Zappr config?

We should also change the default branch to `main` on [codeclimate](https://codeclimate.com/github/zalando/connexion). @jmcs is this something you can do?


hjacobs:
:+1:

PR #1377, opened 2021-07-03
Don't import concrete security factories in init

Ruwann:
The concrete `SecurityHandlerFactory` implementations require the corresponding libraries to be installed.
When importing them in `security.__init__.py`, this can raise an ImportError.

For example, when running the helloworld example which uses `FlaskApp` and shouldn't require `aiohttp`:
```
$ python hello.py 
  File "hello.py", line 10, in <module>
    app = connexion.FlaskApp(__name__, port=9090, specification_dir='swagger/')
  File "/home/ruwan/projects/connexion/connexion/__init__.py", line 22, in _required_lib
    raise exc
  File "/home/ruwan/projects/connexion/connexion/__init__.py", line 30, in <module>
    from .apis.flask_api import FlaskApi, context  # NOQA
  File "/home/ruwan/projects/connexion/connexion/apis/flask_api.py", line 13, in <module>
    from connexion.security import FlaskSecurityHandlerFactory
  File "/home/ruwan/projects/connexion/connexion/security/__init__.py", line 8, in <module>
    from .aiohttp_security_handler_factory import AioHttpSecurityHandlerFactory  # NOQA
  File "/home/ruwan/projects/connexion/connexion/security/aiohttp_security_handler_factory.py", line 3, in <module>
    import aiohttp
ModuleNotFoundError: No module named 'aiohttp'
```

This PR addresses this by not importing the concrete classes in `security.__init__.py`. Another option is to do something similar to `connexion.__init__.py`, but I opted for this approach because the `SecurityHandlerFactory` isn't exposed directly to the end user, but happy to change it so this also uses that approach.

### DIFF
--- a/connexion/security/__init__.py
+++ b/connexion/security/__init__.py
@@ -3,7 +3,3 @@
 # abstract
 from .async_security_handler_factory import AbstractAsyncSecurityHandlerFactory  # NOQA
 from .security_handler_factory import AbstractSecurityHandlerFactory  # NOQA
-
-# concrete
-from .aiohttp_security_handler_factory import AioHttpSecurityHandlerFactory  # NOQA
-from .flask_security_handler_factory import FlaskSecurityHandlerFactory  # NOQA
--- a/connexion/security/__init__.py
+++ b/connexion/security/__init__.py
@@ -3,3 +3,24 @@
 # abstract
 from .async_security_handler_factory import AbstractAsyncSecurityHandlerFactory  # NOQA
 from .security_handler_factory import AbstractSecurityHandlerFactory  # NOQA
+
+
+def not_installed_error(exc):  # pragma: no cover
+    import functools
+
+    def _required_lib(exc, *args, **kwargs):
+        raise exc
+
+    return functools.partial(_required_lib, exc)
+
+
+# concrete
+try:
+    from .flask_security_handler_factory import FlaskSecurityHandlerFactory
+except ImportError as err:  # pragma: no cover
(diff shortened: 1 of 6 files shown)


RobbeSneyders:
Good catch! Not a big preference for one solution, but I do like the cleaner imports by importing them in `__init__`.

I think we should update our tests to prevent this from happening again. We can [group tests by marking them](https://docs.pytest.org/en/6.2.x/example/markers.html) and run separate test suites for aiohttp & Flask with only the necessary imports installed.

Ruwann:
Thanks! I have updated the PR to use the approach of catching the `ImportError`s in `security.__init__.py`.

I have also opened issue #1389 to discuss how to best set up the testing suites for optional dependencies, as there are different ways of doing this.

connexion/security/__init__.py
```diff
@@ -4,6 +4,23 @@
 from .async_security_handler_factory import AbstractAsyncSecurityHandlerFactory  # NOQA
 from .security_handler_factory import AbstractSecurityHandlerFactory  # NOQA
 
+
+def not_installed_error(exc):  # pragma: no cover
```
RobbeSneyders, line 8: We have the same function in `connexion.__init__.py`.
Can we move it to `utils.py`?

RobbeSneyders (approved):
Ok sounds good!
Left one small comment.

connexion/security/__init__.py
Ruwann: Good point, done!

PR #1383, opened 2021-07-04
Upgrade syntax for Python >= 3.6

cclauss:
These changes were made with [`pyupgrade --py36-plus **/*.py`](https://github.com/asottile/pyupgrade)

Python f-strings simplify the code and [should speed up execution](https://www.scivision.dev/python-f-string-speed).


RobbeSneyders (approved):
Thx!

PR #1384, opened 2021-07-04
tox.ini: isort==5.9.1

cclauss:
Fixes #1382
Run `isort .` with the current version of https://github.com/PyCQA/isort

Changes proposed in this pull request:

 -
 -
 -

PR #1382, opened 2021-07-04
tox.ini: isort==5.9.1

cclauss:
https://pypi.org/project/isort

Fixes # .

Changes proposed in this pull request:

 -
 -
 -

PR #1388, opened 2021-07-05
Upgrade isort==5.9.1

cclauss:
Fixes #1382
Fixes #1384
Fixes #1387
https://pypi.org/project/isort

### DIFF
--- a/tests/decorators/test_validation.py
+++ b/tests/decorators/test_validation.py
@@ -1,12 +1,11 @@
 from unittest.mock import MagicMock
 
 import pytest
-from jsonschema import ValidationError
-
 from connexion.apis.flask_api import FlaskApi
 from connexion.decorators.validation import ParameterValidator
 from connexion.json_schema import (Draft4RequestValidator,
                                    Draft4ResponseValidator)
+from jsonschema import ValidationError
 
 
 def test_get_valid_parameter():
--- a/tests/decorators/test_validation.py
+++ b/tests/decorators/test_validation.py
@@ -1,11 +1,12 @@
 from unittest.mock import MagicMock
 
 import pytest
+from jsonschema import ValidationError
+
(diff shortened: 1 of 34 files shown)


RobbeSneyders:
Thx @cclauss.

There seem to be two reasons isort is failing:
- Isort thinks `aiohttp` is local in the tests because of the `tests/aiohttp` directory. We can pass `aiohttp` as a `--thirdparty` to fix this.
- You made some changes which treat `conftest` as third party. I think we need to revert those changes.

cclauss:
Bullet 1: I did try `isort --thirdparty connexion,aiohttp ./tests` and that might have gotten us closer but still not yet passing.

Bullet 2: The only [PR that I landed](https://github.com/zalando/connexion/pulls?q=is%3Apr+is%3Aclosed+author%3Acclauss) that modified `./tests` was #1383 but that did not modify import ordering.  There seem to be [isort issues with conftest.py](https://github.com/PyCQA/isort/search?q=conftest&type=issues).

tests/decorators/test_validation.py
```diff
@@ -1,11 +1,12 @@
 from unittest.mock import MagicMock
 
 import pytest
+from jsonschema import ValidationError
+
 from connexion.apis.flask_api import FlaskApi
```
RobbeSneyders, line 6: This change doesn't treat connexion as third party anymore.

RobbeSneyders (commented):
I meant the changes to `tests` in this PR. Added an in-line comment with one example.

cclauss:
Now I understand...

`tox` does `cd tests` before it runs `isort` which produces different results that `isort tests`.  Thanks for the guidance.

RobbeSneyders (approved):
Ah, that explains it. Thanks a lot for this and your other PRs!

PR #1391, opened 2021-07-07
Add readthedocs config with docs requirements

RobbeSneyders:
Fixes #1390

https://docs.readthedocs.io/en/stable/faq.html#my-documentation-requires-additional-dependencies

PR #1394, opened 2021-07-08
update-routing-docs-add-pythonic-params

### LINKED ISSUE #1329: Connexion does not convert query parameters with hyphens to underscores in python code
### Description

Hi, I have a set of query parameters like the following:

```yaml
openapi: '3.0.3'
# ...
components:
    parameters:
        CreatedByQuery:
            name: created-by
            in: query
            schema:
                type: string
```

I expected that when using this query parameter on an endpoint, that the call to the python controller function would have a `created_by="someone"` keyword argument as a snake_case representation of the kebab-case query (or even `createdBy="someone"` if we are in Javascript lowerCamelCase land) , but instead I simply get `createdby="someone"` without any spaces, which was unexpected! 

Is this intended behaviour and is there any way to specify how the query names get converted to python kwarg names?

### Additional info:

Output of the commands:

- `python --version` -> `Python 3.8.5`
- `pip show connexion | grep "^Version\:"` -> `Version: 2.7.0`

aminalaee:
Fixes #1329

 - Adds docs for *pythonic_params* in the routing sanitsation docs

Maybe we can have another heading `Parameter Name Conversions` in the docs.

### DIFF
--- a/docs/routing.rst
+++ b/docs/routing.rst
@@ -241,6 +241,17 @@ Without this sanitation it would e.g. be impossible to implement an
 `OData
 <http://www.odata.org>`_ API.
 
+You can also convert *CamelCase* parameters to *snake_case* automatically using `pythonic_params` option:
+
+.. code-block:: python
+
+    app = connexion.FlaskApp(__name__)
+    app.add_api('api.yaml', ..., pythonic_params=True)
+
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Thx! Don't pay attention to the failing docs build, #1391 should fix this.

RobbeSneyders (approved):

PR #1404, opened 2021-07-21
Render correct template in API

### LINKED ISSUE #1289: Serving swagger 2.0 and openapi 3.0 at the same time
### Description
We use  have an application which needs to support two versions of APIs at the same time. Version 1 uses swagger 2.0 and version 2 uses openapi 3.0. When both versions are enabled, the swagger UI of version 1 API can be accessed successfully but the version 2 API cannot be accessed. The applications gives the following error

```
werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
```

But when only the version 2 API is enabled, the swagger UI can be accessed successfully.

### Expected behaviour

When the application serve both API versions, the swagger UI of both should be accessible

### Actual behaviour

When the application serve both API versions, only the swagger UI of the API version which uses swagger 2.0 can be accessed.

### Additional info:

Output of the commands:

- `python --version`: 3.6.5
- `pip show connexion | grep "^Version\:"`: 2.6

RobbeSneyders:
Fixes #1289.

https://github.com/zalando/connexion/issues/1289#issuecomment-884105076 for more info.

I also included a commit which cleans up an unused constant which I noticed while debugging this issue.

### DIFF
--- a/connexion/apis/flask_api.py
+++ b/connexion/apis/flask_api.py
@@ -4,6 +4,7 @@ Connexion requests / responses.
 """
 
 import logging
+import pathlib
 import warnings
 from typing import Any
 
@@ -296,7 +297,13 @@ class InternalHandlers:
         }
         if self.options.openapi_console_ui_config is not None:
             template_variables['configUrl'] = 'swagger-ui-config.json'
-        return flask.render_template('index.j2', **template_variables)
+
+        # Use `render_template_string` instead of `render_template` to circumvent the flask
+        # template lookup mechanism and explicitly render the template of the current blueprint.
+        # https://github.com/zalando/connexion/issues/1289#issuecomment-884105076
+        template_dir = pathlib.Path(self.options.openapi_console_ui_from_dir)
+        index_path = template_dir / 'index.j2'
+        return flask.render_template_string(index_path.read_text(), **template_variables)
 
     def console_ui_static_files(self, filename):
         """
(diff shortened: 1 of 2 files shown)


hjacobs:
I guess you mean `fixes #1289`? (#289 is a PR)

RobbeSneyders:
Good catch! Updated.

connexion/apis/flask_api.py
```diff
@@ -296,7 +297,12 @@ def console_ui_home(self):
         }
         if self.options.openapi_console_ui_config is not None:
             template_variables['configUrl'] = 'swagger-ui-config.json'
-        return flask.render_template('index.j2', **template_variables)
+
+        # Use `render_template_string` instead of `render_template` to circumvent the flask
+        # template lookup mechanism and explicitly render the template of the current blueprint.
```
Ruwann, line 302: Could be nice to include a link to the relevant flask docs or the github issue as well for full context?

Ruwann (approved):
Does it make sense to add a test for this? Would be nice imo, but I don't see a good way of testing this at first sight as it seems to be about the extra static content that is loaded when `/ui` is called that isn't rendered, while the page itself returns a `200` response

connexion/apis/flask_api.py
RobbeSneyders: Added link to my comment on the issue.

RobbeSneyders:
I think adding a test for this isn't straightforward indeed, and don't think it's worth the effort.

Issue #1392, opened 2021-07-07
Python syntax error in type comments


cclauss:
### Description
It might be useful to use [`mypy`](http://mypy-lang.org) or similar tools to implement checking of Python type hints but they fail on:
$ `flake8 --count --select=F723 --show-source --statistics`
```
./api/test_parameters.py:60:76: F723 syntax error in type comment '[str] multi array with csv format'
    array_response = json.loads(response.data.decode('utf-8', 'replace'))  # type: [str] multi array with csv format
                                                                           ^
./api/test_parameters.py:64:76: F723 syntax error in type comment '[str] multi array with csv format'
    array_response = json.loads(response.data.decode('utf-8', 'replace'))  # type: [str] multi array with csv format
                                                                           ^
./api/test_parameters.py:68:76: F723 syntax error in type comment '[int] multi array with pipes format'
    array_response = json.loads(response.data.decode('utf-8', 'replace'))  # type: [int] multi array with pipes format
                                                                           ^
./api/test_parameters.py:90:76: F723 syntax error in type comment '[str] multi array with csv format'
    array_response = json.loads(response.data.decode('utf-8', 'replace'))  # type: [str] multi array with csv format
                                                                           ^
./api/test_parameters.py:95:76: F723 syntax error in type comment '[int] multi array with pipes format'
    array_response = json.loads(response.data.decode('utf-8', 'replace'))  # type: [int] multi array with pipes format
                                                                           ^
5     F723 syntax error in type comment '[str] multi array with csv format'
5
```
Does the project use a different Python type hint tool that is incompatible with [`mypy`](http://mypy-lang.org)?

### Additional info:

Output of the commands:

- `python --version`
- `pip show connexion | grep "^Version\:"`

RobbeSneyders:
Hi @cclauss, thanks for the report.
We are currently not checking the type hints, but now that only Python 3.6+ is supported, we should move away from type comments to actual type annotations and start checking them. I was planning to submit a proposal for a gradual approach soon, but anyone is free to take this up already.

We already have #998 to track the introduction of typing though, so I'll close this one if you don't mind.

Issue #1111, opened 2019-12-19
Add type annotations and mypy validation


hjacobs:
As we only support Python 3 now, we can safely use type annotations (`typing` module) and use [mypy](http://mypy-lang.org/) for validation, e.g. like here https://github.com/hjacobs/kube-resource-report/blob/master/Makefile#L11

This is a follow-up to #2.

cclauss:
`mypy --exclude '/(__init__|api|api.pets|app|hello|orm|resty)\.py$' --install-types --non-interactive .` --> #1405

PR #1405, opened 2021-07-24
Tox.ini: Run mypy in ignore_outcome mode

cclauss:
Related to #2, #998, #1111
Proves #1392

`mypy --exclude '/(__init__|api|api.pets|app|hello|orm|resty)\.py$' --install-types --non-interactive .`

 - Run [`mypy`](https://mypy.readthedocs.io) in [tox ignore_outcome](https://tox.readthedocs.io/en/latest/config.html#conf-ignore_outcome) mode to show progress on #1111
 - Making mypy work is a pain in the neck!!

```
tox: mypy
  mypy run-test-pre: PYTHONHASHSEED='866419681'
  mypy run-test: commands[0] | mypy --exclude '/(__init__|api|api.pets|app|hello|orm|resty)\.py$' \
       --install-types --non-interactive .
  tests/api/test_parameters.py:62: error: syntax error in type comment "[str] multi array with csv format"
  tests/api/test_parameters.py:66: error: syntax error in type comment "[str] multi array with csv format"
  tests/api/test_parameters.py:70: error: syntax error in type comment "[int] multi array with pipes format"
  tests/api/test_parameters.py:92: error: syntax error in type comment "[str] multi array with csv format"
  tests/api/test_parameters.py:97: error: syntax error in type comment "[int] multi array with pipes format"
  Found 5 errors in 1 file (errors prevented further checking)
  WARNING: command failed but result from testenv is ignored
  cmd: InvocationError for command /home/runner/work/connexion/connexion/.tox/mypy/bin/mypy \
       --exclude '/(__init__|api|api.pets|app|hello|orm|resty)\.py$' --install-types --non-interactive . \
       (exited with code 2)
```

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -21,8 +21,9 @@ jobs:
         pip install --upgrade pip
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
-      run: mkdir .mypy_cache
-      run: tox
+      run: |
+        mkdir .mypy_cache
+        tox
     - name: Coveralls
       run: coveralls --service github
       env:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -21,9 +21,7 @@ jobs:
         pip install --upgrade pip
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
-      run: |
-        mkdir -p .mypy_cache
-        tox
+      run: tox
     - name: Coveralls
       run: coveralls --service github
       env:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -21,6 +21,7 @@ jobs:
         pip install --upgrade pip
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
+      run: mkdir .mypy_cache
       run: tox
     - name: Coveralls
       run: coveralls --service github
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -22,7 +22,7 @@ jobs:
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
       run: |
-        mkdir .mypy_cache
+        mkdir -p .mypy_cache
         tox
     - name: Coveralls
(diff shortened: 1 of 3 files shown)


.github/workflows/pipeline.yml
```diff
@@ -21,7 +21,9 @@ jobs:
         pip install --upgrade pip
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
-      run: tox
+      run: |
+        mkdir .mypy_cache
```
Ruwann, line 25: Is this only necessary for the `--install-types` argument? How about explicitly adding the necessary dependencies for typechecking, so these packages are also pinned?

tox.ini
```diff
@@ -57,3 +58,8 @@ basepython=python3
 deps=isort==5.9.1
 changedir={toxinidir}/tests
 commands=isort --thirdparty aiohttp,connexion --check-only --diff .
+
+[testenv:mypy]
+deps=mypy==0.910
+ignore_outcome=true
+commands=mypy --exclude '/(__init__|api|api.pets|app|hello|orm|resty)\.py$' --install-types --non-interactive .
```
Ruwann, line 65: Looking at the commits you made, I believe using `--exclude 'examples/'` should do the trick?
Or could we simply do `mypy connexion` (not sure about a potential `-m` or `-p` flag)?

.github/workflows/pipeline.yml
```diff
@@ -21,7 +21,9 @@ jobs:
         pip install --upgrade pip
         pip install "tox<4" "tox-gh-actions<3" "setuptools<58" "coveralls<4"
     - name: Test with tox
-      run: tox
+      run: |
+        mkdir -p .mypy_cache
```
Ruwann, line 25: Do we explicitly need to create this? Don't think this actually does something as we are running mypy once?

tox.ini
```diff
@@ -57,3 +58,8 @@ basepython=python3
 deps=isort==5.9.1
 changedir={toxinidir}/tests
 commands=isort --thirdparty aiohttp,connexion --check-only --diff .
+
+[testenv:mypy]
+deps=mypy==0.910
+ignore_outcome=true
+commands=mypy --exclude 'examples/' --install-types --non-interactive .
```
Ruwann, line 65: I'm no expert, but using `--install-types` installs the latest versions I believe? Perhaps we should explicitly add the types packages as dependencies for this step (`types-setuptools`, `types-PyYAML`, `types-requests`) to limit the upper version as in other steps? What do you think?

PR #1411, opened 2021-08-08
Update basicauth example

### LINKED ISSUE #1406: HTTP basic auth example probably should not have scopes
### Description

I am fairly new to OpenAPI, so I apologize if I got this wrong.
In the [HTTP basic auth example](https://github.com/zalando/connexion/blob/2229831e80f55e1f8bfb536f73bc77013566df21/examples/openapi3/basicauth/openapi.yaml#L18) it says:
```yaml

openapi: 3.0.0
[...]
      security:
        - basic: ['secret']
components:
  securitySchemes:
    basic:
      type: http
      scheme: basic
```
Doesn't this violate the [OAS 3.0.0](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/3.0.0.md#securityRequirementObject) which states: `If the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names required for the execution. For other security scheme types, the array MUST be empty.`

### Expected behaviour

```yaml
        - basic: []
```

### Actual behaviour

```yaml
        - basic: ['secret']
```

aminalaee:
Fixes #1406

 - Remove security scope

### DIFF
--- a/examples/openapi3/basicauth/openapi.yaml
+++ b/examples/openapi3/basicauth/openapi.yaml
@@ -15,7 +15,7 @@ paths:
               schema:
                 type: string
       security:
-        - basic: ['secret']
+        - basic: []
 components:
   securitySchemes:
     basic:


Ruwann:
Thanks for the PR!

PR #1415, opened 2021-08-22
#1256 AioHttpAPP.get_root_path return Path object

### LINKED ISSUE #1256: AioHttpApp.get_root_path returns string, not pathlib.Path
### Description

AioHttpApp.get_root_path returns string
FlaskApp.get_root_path returns pathlib.Path

### Expected behaviour

Both return the same type, eg. pathlib.Path

### Actual behaviour

They use different types. This complicates tests.

### Steps to reproduce

See code.

### Additional info:

Python 3.7.7
Version: 2020.0.dev1

nickyfoto:
Fixes #1256 

 - update AioHTTPAPP get_root_path from str to Path
 - Added unit tests

### DIFF
--- a/connexion/apps/aiohttp_app.py
+++ b/connexion/apps/aiohttp_app.py
@@ -4,6 +4,7 @@ This module defines an AioHttpApp, a Connexion application to wrap an AioHttp ap
 
 import logging
 import os.path
+import pathlib
 import pkgutil
 import sys
 
@@ -29,7 +30,8 @@ class AioHttpApp(AbstractApp):
     def get_root_path(self):
         mod = sys.modules.get(self.import_name)
         if mod is not None and hasattr(mod, '__file__'):
-            return os.path.dirname(os.path.abspath(mod.__file__))
+            root_path = os.path.dirname(os.path.abspath(mod.__file__))
+            return pathlib.Path(root_path)
 
         loader = pkgutil.get_loader(self.import_name)
         filepath = None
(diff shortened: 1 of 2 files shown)


connexion/apps/aiohttp_app.py
```diff
@@ -29,7 +30,8 @@ def create_app(self):
     def get_root_path(self):
         mod = sys.modules.get(self.import_name)
         if mod is not None and hasattr(mod, '__file__'):
-            return os.path.dirname(os.path.abspath(mod.__file__))
+            root_path = os.path.dirname(os.path.abspath(mod.__file__))
+            return pathlib.Path(root_path)
```
Ruwann, line 34: We can also leverage pathlib to do path operations:
```suggestion
            return pathlib.Path(mod.__file__).resolve().parent
```

See: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

connexion/apps/aiohttp_app.py
```diff
@@ -40,7 +42,8 @@ def get_root_path(self):
         if filepath is None:
             raise RuntimeError(f"Invalid import name '{self.import_name}'")
 
-        return os.path.dirname(os.path.abspath(filepath))
+        root_path = os.path.dirname(os.path.abspath(filepath))
+        return pathlib.Path(root_path)
```
Ruwann, line 46:
```suggestion
        return pathlib.Path(filepath).resolve().parent
```

connexion/apps/aiohttp_app.py
```diff
@@ -4,6 +4,7 @@
 
 import logging
 import os.path
```
Ruwann, line 6:
```suggestion
```

Ruwann (changes requested):
Thanks for the PR! I left a comment on using `pathlib` for file manipulations as well instead of using `os.path` (as a result, we don't need to import `os.path` anymore as well)

Ruwann (approved):

PR #1418, opened 2021-08-25
Reload Flask on API file changes

aparcar:
Flask can automatically reload in debug mode whenver specified files
change. By default that includes templates, however changes to the API
files (i.e. openapi.yml) should result in a reload as well.

This PR adds all added API files to be monitored automatically by Flask.

WIP since I have to figure out the used `tox` setup first.

Signed-off-by: Paul Spooren <mail@aparcar.org>

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -23,6 +23,7 @@ logger = logging.getLogger('connexion.app')
 class FlaskApp(AbstractApp):
     def __init__(self, import_name, server='flask', **kwargs):
         super().__init__(import_name, FlaskApi, server=server, **kwargs)
+        self.extra_files = kwargs.get("extra_files", [])
 
     def create_app(self):
         app = flask.Flask(self.import_name, **self.server_args)
@@ -64,6 +65,8 @@ class FlaskApp(AbstractApp):
     def add_api(self, specification, **kwargs):
         api = super().add_api(specification, **kwargs)
         self.app.register_blueprint(api.blueprint)
+        if isinstance(specification, (str, pathlib.Path)):
+            self.extra_files.append(specification)
         return api
 
     def add_error_handler(self, error_code, function):
@@ -101,7 +104,8 @@ class FlaskApp(AbstractApp):
 
         logger.debug('Starting %s HTTP server..', self.server, extra=vars(self))
         if self.server == 'flask':
-            self.app.run(self.host, port=self.port, debug=self.debug, **options)
+            self.app.run(self.host, port=self.port, debug=self.debug,
+                         extra_files=self.extra_files, **options)
         elif self.server == 'tornado':
             try:
                 import tornado.httpserver
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -73,7 +73,13 @@ class FlaskApp(AbstractApp):
         # type: (int, FunctionType) -> None
         self.app.register_error_handler(error_code, function)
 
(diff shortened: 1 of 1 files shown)


Ruwann:
Hi @aparcar , thanks for the PR, let me know if you need some help

aparcar:
@Ruwann Thanks for your response. I'll look into adding test cases to this PR.

I'm not sure why the current CI is failing on Python 3.9, do you have any ideas?

Also do you require any doc updates for this PR?

connexion/apps/flask_app.py
```diff
@@ -101,7 +103,8 @@ def run(self, port=None, server=None, debug=None, host=None, **options):  # prag
 
         logger.debug('Starting %s HTTP server..', self.server, extra=vars(self))
         if self.server == 'flask':
-            self.app.run(self.host, port=self.port, debug=self.debug, **options)
+            self.app.run(self.host, port=self.port, debug=self.debug,
+                    extra_files=self.extra_files, **options)
```
Ruwann, line 107: The checks are failing because this line is not correctly indented. It should be aligned with the `self.host` in the line above (add 5 spaces I think).

connexion/apps/flask_app.py
```diff
@@ -64,6 +65,7 @@ def common_error_handler(self, exception):
     def add_api(self, specification, **kwargs):
         api = super().add_api(specification, **kwargs)
         self.app.register_blueprint(api.blueprint)
+        self.extra_files.append(specification)
```
Ruwann, line 68: This can be more than just `str`, it can also be a `pathlib.Path` or dict instance

Ruwann:
For the test cases, I'm not sure whether it is possible to write a test to check whether it is reloading in debug mode? You could write a test for checking whether they are added to the `extra_files`, but not sure how useful that test is.

connexion/apps/flask_app.py
aparcar: Now it will only be added in case it's either string or Path

Ruwann (changes requested):
We also need to add this option now to `FlaskApp.run()` because users might already use the `extra_files` argument for other things. So we need to append the `extra_files` in the `run()` function to the `self.extra_files`.
Otherwise, this would lead to a `TypeError: run() got multiple values for keyword argument 'extra_files'`

Is it clear what I mean?

Ruwann (approved):

Ruwann:
@RobbeSneyders This PR looks good to me, but I have made some changes myself as well, let me know if you'd like to have another look at it before it is merged.

connexion/apps/flask_app.py
```diff
@@ -21,8 +21,9 @@
 
 
 class FlaskApp(AbstractApp):
-    def __init__(self, import_name, server='flask', **kwargs):
+    def __init__(self, import_name, server='flask', extra_files=None, **kwargs):
```
RobbeSneyders, line 24: Nit, but all other parameters are documented in the super function. Would be good to document the `extra_files` parameter here? We currently also don't document its type, but assume it's a `list` by calling `.extend()`.

RobbeSneyders (approved):
Thx @aparcar @Ruwann.
One small nit, but LGTM.

connexion/apps/flask_app.py
Ruwann: True, wasn't sure what happens to the auto-generated documentation but seems that it already doesn't include the documentation of the parent class.
I'll add the `extra_files` documentation and refer to the parent class for the other parameters.

PR #1419, opened 2021-08-28
Feature/relative resolver

### LINKED ISSUE #938: Allow specifying the default controller for an API
### Description
If all operations of an API are handled by the same controller, we still need to define `x-swagger-router-controller` on each operation. It would be nice if instead we could specify it in one spot.

Here's my attempt:
```
class ControllerResolver(resolver.Resolver):
  """Connexion allows to specify the controller through """
  def __init__(self, controller: str, *args, **kwargs):
    self.controller = controller
    super(ControllerResolver, self).__init__(*args, **kwargs)

  def resolve_operation_id(self, operation) -> str:
    op = super(ControllerResolver, self).resolve_operation_id(operation)
    if "." in op:
      return op
    return f"{self.controller}.{op}"
```
```
app.add_api(pathlib.Path("swagger.yaml"), resolver=ControllerResolver(controller="my_service.api.rest.controller"))
```

It would be even nicer if we could pass in the actual module instead of a string.

Ruwann:
Fixes #938 .

 - Add `RelativeResolver` class to resolve add `operationId`s relative to the same module/path without having to specify the same `x-swagger-router-controller` or `x-openapi-router-controller` for every operation individually

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -2,13 +2,18 @@ import connexion.apps
 import pytest
 from connexion.exceptions import ResolverError
 from connexion.operations import Swagger2Operation
-from connexion.resolver import Resolver, RestyResolver
+from connexion.resolver import RelativeResolver, Resolver, RestyResolver
 
 PARAMETER_DEFINITIONS = {'myparam': {'in': 'path', 'type': 'integer'}}
 
 
 def test_standard_get_function():
     function = Resolver().resolve_function_from_operation_id('connexion.FlaskApp.common_error_handler')
+    assert function == connexion.FlaskApp.common_error_handler
+
+
+def test_relative_get_function():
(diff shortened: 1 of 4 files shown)


tests/test_resolver.py
```diff
@@ -12,6 +12,10 @@ def test_standard_get_function():
     assert function == connexion.FlaskApp.common_error_handler
 
 
+def test_relative_get_function():
+    function = RelativeResolver('connexion').resolve_function_from_operation_id('connexion.FlaskApp.common_error_handler')
+    assert function == connexion.FlaskApp.common_error_handler
+
```
RobbeSneyders, line 18: Extreme nit: missing newline.

RobbeSneyders (approved):
Sorry @Ruwann, lost track of this PR due to my holidays.

Looks (almost) perfect ;)

PR #1425, opened 2021-09-10
#1424: add keywords to the list of pythonic words

### LINKED ISSUE #1424: Parameters named "class" (or other python reserved words) don't work, even with pythonic_params=True?
### Description

I have a test endpoint with a query param called `class`, like so:

```
    /test:
        get:
            operationId: "operations.test_this"
            parameters:
                - in: query
                  name: class
                  schema:
                      type: string
                - in: query
                  name: yoohoo
                  schema:
                      type: string
                - in: query
                  name: filter
                  schema:
                      type: string
```

and in the routing,

```
def test_this(class_="blank", yoohoo="dd", filter_="sdfsdfs"):
    response = locals()
    http_status_code = 200
    return response, http_status_code
```

### Expected behaviour

`curl "http://localhost:3333/htsget/v1/test?class=header&yoohoo=yoohoo&filter=dddd"` should result in:
```
{
  "class_": "header",
  "filter_": "dddd",
  "yoohoo": "yoohoo"
}
```

### Actual behaviour

It actually results in:
```
{
  "class_": "blank",
  "filter_": "dddd",
  "yoohoo": "yoohoo"
}
```

### Steps to reproduce
As above

### Additional info:

Output of the commands:

- `python --version`
-- `Python 3.7.12`

-- `Version: 2020.0.dev1`

daisieh:
Fixes #1424 .

 - Add Python's keywords to the list of reserved words for underscoring.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -43,10 +43,7 @@ def snake_and_shadow(name):
     :return:
     """
     snake = inflection.underscore(name)
-    all_reserved = []
-    all_reserved.extend(list(builtins.__dict__.keys()))
-    all_reserved.extend(keyword.kwlist)
-    if snake in all_reserved:
+    if snake in builtins.__dict__ or keyword.iskeyword(snake):
         return f"{snake}_"
     return snake
 
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -3,6 +3,7 @@ This module defines a decorator to convert request parameters to arguments for t
 """
 
 import builtins
+import keyword
 import functools
 import inspect
 import logging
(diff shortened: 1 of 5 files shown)


connexion/decorators/parameter.py
```diff
@@ -42,7 +43,10 @@ def snake_and_shadow(name):
     :return:
     """
     snake = inflection.underscore(name)
-    if snake in builtins.__dict__.keys():
+    all_reserved = []
+    all_reserved.extend(list(builtins.__dict__.keys()))
+    all_reserved.extend(keyword.kwlist)
+    if snake in all_reserved:
```
Ruwann, line 49: We can also use the `keyword.iskeyword()` function to avoid needing to build a list:

```suggestion
    if snake in builtins.__dict__ or keyword.iskeyword(snake):
```

What do you think?

Ruwann (commented):
Thanks for the PR (and the updated tests)!
I left one comment to avoid building a list

Ruwann (approved):
Thanks!

PR #1435, opened 2021-10-11
Fix fail to parse body for open api schema defining multiple content …

MarcoRizk:
…types

Fixes
When using an open api schema with multiple content type definition like:
```
requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/DownTimeLogCreate'
            schema:
              $ref: '#/components/schemas/DownTimeLogCreate'
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/DownTimeLogCreate'
        required: true
```
sending a request with a json body causes the following error:
```
  File "/home/marco/Projects/Synapse/azka_vision_mock_server/venv/lib/python3.6/site-packages/connexion/decorators/parameter.py", line 97, in wrapper
    request.files, arguments, has_kwargs, sanitize)
  File "/home/marco/Projects/Synapse/azka_vision_mock_server/venv/lib/python3.6/site-packages/connexion/operations/abstract.py", line 282, in get_arguments
    has_kwargs, sanitize))
  File "/home/marco/Projects/Synapse/azka_vision_mock_server/venv/lib/python3.6/site-packages/connexion/operations/openapi.py", line 296, in _get_body_argument
    body_arg.update(body or {})
TypeError: cannot convert dictionary update sequence element #0 to a sequence
```

 - I am proposing that the request body parsing also considers the CONTENT-TYPE header sent with the request not just the open api schema definition

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -14,7 +14,7 @@ import inflection
 
 from ..http_facts import FORM_CONTENT_TYPES
 from ..lifecycle import ConnexionRequest  # NOQA
-from ..utils import all_json
+from ..utils import all_json, is_json_mimetype
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Hi @MarcoRizk, thanks for the PR. 

There is currently indeed a limitation in Connexion when using multiple content types, but I don't think this is sufficient to solve it. I agree that we can use the headers, but they should be used to select the correct definition in the API contract, both for parameter parsing and validation.

gfleischer-meta:
hey any news on this PR
@MarcoRizk did u manage to fix this issue?

RobbeSneyders:
Fixed in https://github.com/spec-first/connexion/pull/1588

PR #1440, opened 2021-11-19
Allow object in multipart/form-data body

vpohl:
Hi there,
when you define a multipart request, having both an object and a file upload, i.e. e.g., 
```yaml
requestBody:
  content: 
    multipart/form-data: # Media type
      schema:            # Request payload
        type: object
        properties:      # Request parts
          id:            # Part 1 (string value)
            type: string
            format: uuid
          address:       # Part2 (object)
            type: object
            properties:
              street:
                type: string
              city:
                type: string
          profileImage:  # Part 3 (an image)
            type: string
            format: binary
```
(cf. https://swagger.io/docs/specification/describing-request-body/multipart-requests/)
you will get validation errors.

 - Convert the request body into the correct data type
   - Altering the request.form is not very clean but if you are not doing this, the parameter validation fails and if you circumvent that validation, it won't appear in the `body`.

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -166,6 +166,12 @@ class RequestBodyValidator:
                 if data is not None or not self.has_default:
                     self.validate_schema(data, request.url)
             elif self.consumes[0] in FORM_CONTENT_TYPES:
+                # Convert multipart object items into dictionaries
+                for k, v in request.form.items():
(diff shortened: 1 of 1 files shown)


Ruwann:
Thanks for the PR!
Could you have a look at the tests? There are some that are failing now.
It would also be nice if you could include a specific test case for this behaviour

PR #1442, opened 2021-12-03
Fix sanitization of function arguments

### LINKED ISSUE #1183: "pythonic_params" causes strange behavior with incorrectly-cased query parameters
### Description

I have an app that sets `pythonic_params=True`. The app has an endpoint with a camel case query parameter (eg `orderBy=ascending`). If a request erroneously provides an `order_by=ascending` query parameter a mangled value is passed to the view function.

### Expected behaviour

The snake case query parameter is ignored, or correctly deserialized. 

### Actual behaviour

The route is passed `order_by = "['ascending']"`.

#### openapi.yml

```
openapi: '3.0.0'
info:
  title: 'pythonic_params bug'
  version: '1.0.0'

paths:
  /query:
    get:
      operationId: "app.test_query"
      parameters:
        - name: orderBy
          in: query
          required: false
          schema:
            type: string
      responses:
        default:
          description: 'A test'

```

#### app.py

```
from typing import Optional

import connexion


def test_query(order_by: Optional[str] = None):
    return {'order_by': str(order_by), 'type': str(type(order_by))}


if __name__ == "__main__":
    app = connexion.App(__name__)
    app.add_api('openapi.yml', pythonic_params=True)
    app.run(port=8000)

```

#### Request
```
curl "localhost:8000/v1/query?order_by=ascending"
```

#### Response

```
{
  "order_by": "['ascending']",
  "type": "<class 'str'>"
}
```

### Additional info:

Output of the commands:

- `python --version` - Python 3.7.4 
- `pip show connexion | grep "^Version\:"` - Version: 2.6.0
- `pip freeze`

```
certifi==2019.11.28
chardet==3.0.4
click==7.1.1
clickclick==1.2.2
connexion==2.6.0
Flask==1.1.1
idna==2.9
inflection==0.3.1
itsdangerous==1.1.0
Jinja2==2.11.1
jsonschema==2.6.0
MarkupSafe==1.1.1
openapi-spec-validator==0.2.8
PyYAML==5.3.1
requests==2.23.0
six==1.14.0
urllib3==1.25.8
Werkzeug==1.0.0
```

Ruwann:
Fixes #1183.

 - Only perform sanitization when checking the python view function arguments and move the parameter sanitization as much as possible to the same place

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -83,7 +83,7 @@ def parameter_to_arg(operation, function, pythonic_params=False,
         if all_json(consumes):
             request_body = request.json
         elif consumes[0] in FORM_CONTENT_TYPES:
-            request_body = {sanitize(k): v for k, v in request.form.items()}
+            request_body = request.form
         else:
             request_body = request.body
 
--- a/connexion/operations/abstract.py
+++ b/connexion/operations/abstract.py
@@ -191,18 +191,21 @@ class AbstractOperation(SecureOperation, metaclass=abc.ABCMeta):
                            function_arguments, has_kwargs, sanitize):
         res = {}
         for key, value in query_arguments.items():
-            key = sanitize(key)
-            if not has_kwargs and key not in function_arguments:
-                logger.debug("Query Parameter '%s' not in function arguments", key)
+            sanitized_key = sanitize(key)
+            if not has_kwargs and sanitized_key not in function_arguments:
+                logger.debug("Query Parameter '%s' (sanitized: '%s') not in function arguments",
+                             key, sanitized_key)
             else:
-                logger.debug("Query Parameter '%s' in function arguments", key)
+                logger.debug("Query Parameter '%s' (sanitized: '%s') in function arguments",
+                             key, sanitized_key)
                 try:
                     query_defn = query_defns[key]
                 except KeyError:  # pragma: no cover
-                    logger.error(f"Function argument '{key}' not defined in specification")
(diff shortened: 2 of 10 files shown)


Ruwann:
Yes, agreed, the goal of this PR was not to accept and to attempt validating parameters in a different format than what's specified in the spec. Right now, these two interfaces are not very clear and the "translation" between the two happens in multiple places which leads to errors such as in #1183 .

Imo, the translation of `orderBy` to `order_by` should only be done right when checking the view function's parameters and passing in the corresponding value. The checking of the request's parameters should all be based on the spec (before any sanitization and "pythonic param conversion").

I'm not entirely sure, but I feel we should accept a parameter `order_by` but not try to validate it based on what's specified in the contract for `orderBy`. I don't immediately find an answer on whether this extra parameter should be rejected or silently ignored in the OAS.

RobbeSneyders:
Ok, I think we agree and I actually misread your tests. Looking back at them, they describe the exact behavior I was aiming at. Sorry for the confusion!

> I'm not entirely sure, but I feel we should accept a parameter `order_by` but not try to validate it based on what's specified in the contract for `orderBy`. I don't immediately find an answer on whether this extra parameter should be rejected or silently ignored in the OAS.

Agree. There doesn't seem to be a clear consensus, but the discussion in OAI/OpenAPI-Specification#568 is interesting, and I tend to agree with the view of OpenAPI being an open world contract. So I would stick with the current behavior and ignore it.

RobbeSneyders (approved):
Thanks!

PR #1447, opened 2022-01-16
Bump upperbound version for jsonschema to 5.0.0

### LINKED ISSUE #1430: Support jsonschema >= 4.0.0
### Description

connexion uses the `_util` part of `jsonschema` package. There was a breaking change in the latest release 4.0 release of `jsonschema` removing the `types_msg` function.
Removal commit: https://github.com/Julian/jsonschema/commit/00031cb043763842266877923552e40f0c8f36b5

https://github.com/zalando/connexion/blob/main/connexion/json_schema.py#L64 uses the removed function.

### Expected behaviour

Not use missing private function.

### Actual behaviour

Uses missing private function. 

### Steps to reproduce

Install `jsonschema` with version 4.0 and up.

### Additional info:

Output of the commands:

- `python --version`: 3.6.10
- `pip show connexion | grep "^Version\:"`: 2.7.0

RobbeSneyders:
Fixes #1430

 - Drop usage of private jsonschema modules to prevent breaking on non-major version changes.
 - Create validate_defaults function by passing in the instance_validator instead of using an attribute on the OpenApiValidator. This is necessary since the attribute is no longer maintained while the validator descends into the spec. I presume due to the usage of `attrs.resolve` by jsonschema.

### DIFF
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -6,6 +6,7 @@ import abc
 import copy
 import json
 import pathlib
+import typing as t
 from collections.abc import Mapping
 from urllib.parse import urlsplit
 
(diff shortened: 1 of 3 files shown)


connexion/spec.py
```diff
@@ -24,33 +25,39 @@
...
 
 
-def validate_defaults(validator, properties, instance, schema):
-    """Validate `properties` subschema.
+def create_validate_default_fn(instance_validator: Draft4Validator) -> t.Callable:
+    """Creates a validation function for property defaults. This validation function will be used
+    by a validator that validates an openapi spec against the openapi schema. The default value
+    however needs to be validated against the openapi spec itself.
+
+    Args:
```
Ruwann, line 33: Don't think the Google docstring format is used elsewhere. To keep it consistent, let's use [the Sphinx docstring](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) format here as well?

Ruwann (approved):
Thanks, looks great!
Left a small comment on the docstring format

connexion/spec.py
RobbeSneyders: Good catch. Wrong project defaults on new laptop :)

patrickelectric:
Can we get this merged ?

patrickelectric:
Thanks!

alex-zywicki:
When can we expect this to be released?

PR #1448, opened 2022-01-18
don't rely on  jsonschema._utils in json_schema.py

slobinger:
In the `json_schema.validate_type` function `jsonschemas_utils.types_msg`is used, wich leads to an
`AttributeError` if the instance to check has no valid type.
In general it is not a good idea to rely on other libs utils especially if they start with `_`.
So I also replaced the call of `_utils.ensure_list` with it's adjusted implementation.

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -5,7 +5,7 @@ Module containing all code related to json schema validation.
 from collections.abc import Mapping
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Hi @slobinger, thanks for the PR, but this change is already part of PR #1447 as well.
Hope you don't mind if I close this one.

slobinger:
@RobbeSneyders never mind, thanks for the info.

Issue #1452, opened 2022-02-05
[BUG] x-body-name violates the OAS3 spec by being required to be within the schema field of media type


Johnlon:
### Description

The docs and tests promote a schema that is not valid OAS3.

This problem means tools can't work together and forces hacks in other libraries and user code.
For instance this can't work wit the swagger tools or openapi generator.

```
requestBody:
  content:
    application/json:
      schema:
        x-body-name: pet
        $ref: '#/components/schemas/Pet'
```

A media type's "schema" field may be either a single "$ref" OR a full blown schema so you can't has an extension AND a ref.

The current impl alternatively permits the extension to be the referred to schema nit this pollutes the datatype with info that relates only to the usage within a requrst.

So can we support the approach below in addition to whatever you currently have ...

```
requestBody:
  content:
    application/json:
       x-body-name: pet
      schema:
        $ref: '#/components/schemas/Pet'
```

### Expected behaviour

The tool should follow the spec.

### Actual behaviour

x-body-name requires schemata to violate te spec OR pollute datatypes.

### Found in

Current master and cli 2.10.0

Johnlon:
See https://github.com/zalando/connexion/pull/1453

Johnlon:
I've added a feature and updated the docs.

PR #1453, opened 2022-02-05
 issue 1452 - placement of x-body-name isnt valid OAS3 location for an extension

Johnlon:
Fixes # [1452 ](https://github.com/zalando/connexion/issues/1452) - placement of x-body-name isnt valid OAS3 location for an extension

 - permit use of x-body-name at a location that does not violate the OAS spec

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -266,9 +266,21 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody schema
+          provide the x-body-name parameter in your requestBody media-type
+          (or non-OAS compliant legacy position within the requestBody schema)
           to override the name of the parameter that will be passed to your
           handler function.
+
+.. code-block:: yaml
+
+    requestBody:
+      content:
+        application/json:
+          x-body-name: stack
+          schema:
+            # legacy location here should be ignored because the preferred location is at the media-type level above
+            x-body-name: this_should_be_ignored
+            $ref: '#/components/schemas/someComponent'
 
 .. warning:: When you define a parameter at your endpoint as *not* required, and
     this argument does not have default value in your Python view, you will get
--- a/README.rst
+++ b/README.rst
@@ -266,7 +266,7 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody media-type
+          provide the x-body-name parameter in your requestBody 
           (or non-OAS compliant legacy position within the requestBody schema)
           to override the name of the parameter that will be passed to your
           handler function.
@@ -274,11 +274,11 @@ of the endpoint parameter ``message`` to your view function.
 .. code-block:: yaml
 
     requestBody:
+      x-body-name: stack
       content:
         application/json:
-          x-body-name: stack
           schema:
-            # legacy location here should be ignored because the preferred location is at the media-type level above
+            # legacy location here should be ignored because the preferred location is at the requestBody level above
             x-body-name: this_should_be_ignored
             $ref: '#/components/schemas/someComponent'
 
--- a/README.rst
+++ b/README.rst
@@ -266,21 +266,24 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody 
+          provide the x-body-name parameter in your operation 
           (or non-OAS compliant legacy position within the requestBody schema)
           to override the name of the parameter that will be passed to your
           handler function.
 
 .. code-block:: yaml
 
-    requestBody:
-      x-body-name: stack
-      content:
-        application/json:
-          schema:
-            # legacy location here should be ignored because the preferred location is at the requestBody level above
-            x-body-name: this_should_be_ignored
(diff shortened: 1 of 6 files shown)


Johnlon:
@jmcs @hjacobs @rafaelcaricio @dtkav @RobbeSneyders     Please review thanks

Solves a compat issue with OpenApi generator and a bunch of long standing open issues I think.

RobbeSneyders:
Hi @Johnlon, thanks for the report and PR!

Looking at the [OpenAPI documentation](https://swagger.io/docs/specification/openapi-extensions/), it seems like extensions are also not permitted under the media type though:

> Extensions are supported on the root level of the API spec and in the following places:
> - info section
> - paths section, individual paths and operations
> - operation parameters
> - responses
> - tags
> - security schemes

I would propose to place it under the operation, which is allowed and neatly matches the scope of `x-body-name`:

```
  /path:
    post:
      operationId: ...
      x-body-name: ...
      requestBody:
        content:
          application/json:
            schema:
              # should be ignored because the preferred location is at the media-type level above
              x-body-name: ...
```

Johnlon:
Hi see here where it indicates media type may be extended.

https://spec.openapis.org/oas/v3.1.0#media-type-object

Johnlon:
But whatever works

Johnlon:
If we can fix then I'll update my PR on openapi generator.

Thanks

Johnlon:
x-codegen-request-body-name is on operation in openapi-generator  so operation is fine for x-body-name

Johnlon:
However, a practical problem with your suggestion is that it appears to change a public method

To call '_get_body_argument' with the path then we'd need to change the signature ...

```python
    def get_arguments(self, path_params, query_params, body, files, arguments,
                      has_kwargs, sanitize):
        """
        get arguments for handler function
        """
        ret = {}
        ret.update(self._get_path_arguments(path_params, sanitize))
        ret.update(self._get_query_arguments(query_params, arguments,
                                             has_kwargs, sanitize))

        if self.method.upper() in ["PATCH", "POST", "PUT"]:
            ret.update(self._get_body_argument(body, arguments,
                                               has_kwargs, sanitize))
            ret.update(self._get_file_arguments(files, arguments, has_kwargs))
        return ret
```

... whereas putting it on operation is a minor change.

A better compromise is Request Body as it also permits extension and is already available in context..

https://spec.openapis.org/oas/v3.0.2#requestBodyObject

````
    def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
```

Johnlon:
The request body for some strange reason is where the existing OpenApi generator samples seem to think that it's supposed to be  ....

See 
[ samples/server/petstore/python-aiohttp-srclayout/src/openapi_server/openapi/openapi.yaml
](https://github.com/OpenAPITools/openapi-generator/blob/master/samples/server/petstore/python-aiohttp-srclayout/src/openapi_server/openapi/openapi.yaml)


```yml
paths:
  /pet:
    post:
      operationId: add_pet
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
          application/xml:
            schema:
              $ref: '#/components/schemas/Pet'
        description: Pet object that needs to be added to the store
        required: true
        x-body-name: body
```

Johnlon:
@[RobbeSneyders](https://github.com/RobbeSneyders)  Ok Pushed with it at Operation level

Johnlon:
My change to OpenAPI Generator is now contingent on this Connexions fix.
Thanks

connexion/operations/openapi.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import logging
+import io
```
RobbeSneyders, line 6: The pipeline is failing on this import as it's unused.

RobbeSneyders (commented):
Thanks @Johnlon!

Since the extension of `requestBody` is allowed as you showed, it actually might make more sense to put it on that level. With the additional benefit that we match the current expectations of openapi-generator.

Johnlon:
Will check the import.

Open api generator puts it's own x-codegen-request-body-name at the
operation level so it's more sense to be also at operation IMHO.

If it were are operation level then a later discussion would be whether we
need both.

The pr is with operation.

Johnlon:
Like I say  OpenApi generator (see also AbstractPythonConnexionServerCodegen) has their equiv setting x-codegen-request-body-name at the op level.

```yaml
paths:
  /pet:
    post:
      tags:
        - pet
      summary: Add a new pet to the store
      description: ''
      operationId: addPet
      x-codegen-request-body-name: new_body_name
```

Johnlon:
There is a further slight benefit of keeping at the op level and that's that in openapi generator one clould insert x-codegen-request-body-name if x-body-name is seen. This is useful because really the two of these MUST line up otherwise the generated code will not run - both must equate so that the generated controller params match the calls the connextion makes.

Johnlon:
If you want to go at the Request then it's not a big deal - at some point on can change OpenApi Gen to also support at Req level.
Lever me know as I'd like to put myself out of my misery.

README.rst
```diff
@@ -266,10 +266,25 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody schema
+          provide the x-body-name parameter in your operation 
+          (or non-OAS compliant legacy position within the requestBody schema)
```
Ruwann, line 270: I don't think that the current position is necessarily breaking OAS by design, rather, the example of using it alongside a `$ref` at the same level is

Johnlon:
There are examples ( that don't work of course  )  that have it at req
level. I have no idea why someone thought the current situation would work.
No testing I suppose?
Openapi generators own equivalent annotation is at op level so it's
possible to push down from op to req

RobbeSneyders:
Was there a reason to close this PR @Johnlon?

We'd like to get this change merged, but with the extension in the `requestBody`.

PR #1455, opened 2022-02-07
Drop openapi-spec-validator for custom ref handler

### LINKED ISSUE #930: Use jsonschema for validation

RobbeSneyders:
Fixes #930 
Continues on #936 

This PR completely removes the dependency on `openapi-spec-validator`. There's multiple reasons to do this:
- One dependency less
- We don't use `openapi-spec-validator` for it's main purpose, but for utility functionality
- It's not production ready (https://github.com/p1c2u/openapi-schema-validator/issues/31#issuecomment-1014141023)
- We recently had some issues with dependency conflicts (https://github.com/spec-first/connexion/issues/1430#issuecomment-1007737974)

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -2,21 +2,70 @@
 Module containing all code related to json schema validation.
 """
 
+import contextlib
+import io
+import os
 import typing as t
+import urllib.parse
+import urllib.request
 from collections.abc import Mapping
 from copy import deepcopy
 
+import requests
+import yaml
 from jsonschema import Draft4Validator, RefResolver
 from jsonschema.exceptions import RefResolutionError, ValidationError  # noqa
 from jsonschema.validators import extend
-from openapi_spec_validator.handlers import UrlHandler
 
 from .utils import deep_get
 
+
+class ExtendedSafeLoader(yaml.SafeLoader):
+    """Extends the yaml SafeLoader to coerce all keys to string so the result is valid json."""
+
+    def __init__(self, stream):
+        self.original_construct_mapping = self.construct_mapping
+        self.construct_mapping = self.extended_construct_mapping
+        super().__init__(stream)
+
+    def extended_construct_mapping(self, node, deep=False):
+        data = self.original_construct_mapping(node, deep)
+        return {str(key): data[key] for key in data}
+
+
+class FileHandler:
+    """Handler to resolve file refs."""
+
+    def __call__(self, uri):
+        filepath = self._uri_to_path(uri)
+        with open(filepath) as fh:
+            return yaml.load(fh, ExtendedSafeLoader)
+
+    @staticmethod
+    def _uri_to_path(uri):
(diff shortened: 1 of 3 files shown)


connexion/json_schema.py
```diff
@@ -2,21 +2,70 @@
...
 from jsonschema import Draft4Validator, RefResolver
 from jsonschema.exceptions import RefResolutionError, ValidationError  # noqa
 from jsonschema.validators import extend
-from openapi_spec_validator.handlers import UrlHandler
 
 from .utils import deep_get
 
+
+class ExtendedSafeLoader(yaml.SafeLoader):
+    """Extends the yaml SafeLoader to coerce all keys to string so the result is valid json."""
```
Ruwann, line 24: Why do we need to do this? Shouldn't the loaded file be json as well if it adheres to OpenAPI? Perhaps I'm simply overlooking something.

examples/openapi3/methodresolver/openapi/pets-api.yaml
```diff
@@ -42,7 +42,8 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/Pet"
+              $ref: "file:///home/robbe/workspace/oss/connexion/examples/openapi3/methodresolver/openapi/test.json#/components/schemas/Pet"
+#              $ref: "http://0.0.0.0:8000/test.yaml#/components/schemas/Pet"
```
Ruwann, line 46: Can we add a separate test for this and change this back to the original value?

Ruwann (approved):
Thanks, Robbe! Think we just need to adjust the tests for this?

connexion/json_schema.py
RobbeSneyders: The loaded file can be a `yaml` file as well, which allows integer keys. According to the OpenAPI spec, keys should indeed be strings so the `yaml` is a representation of valid `json`, and we shouldn't need this. However `openapi-spec-validator` allowed this, and I tried to provide a drop-in replacement so kept the same behavior.

We could mark this for deprecation in the next major version though.

examples/openapi3/methodresolver/openapi/pets-api.yaml
RobbeSneyders: Woops, this is leftover from my local testing.
I'll remove this and add a separate test. 
Strange that this doesn't fail in the Github pipeline though.

Ruwann: Agreed, I don't see a good way of adding a test for this, so I would merge without it.

PR #1454, opened 2022-02-07
Automate release process via github actions

RobbeSneyders:
This PR automates the release process via Github actions.

It runs on two triggers:
- When merged to `main`, the package is only published to TestPypi
- When a new release is published, the package is published to PyPi.

I temporarily changed the triggers to get a test run on this PR:
https://github.com/spec-first/connexion/runs/5098593555

PR #1456, opened 2022-02-07
Change test release trigger to tags

RobbeSneyders:
[The test release pipeline failed](https://github.com/spec-first/connexion/actions/runs/1809057983) after merging #1454 because you can only upload a package on PyPi with the same version once.

This PR changes the trigger for the test release pipeline to tags, so it only runs when there's a new version.

This way, releasing works as follows:
1. Push tag
2. Test release pipeline
3. Publish release
4. Release pipeline

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,8 @@
 name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 on:
   push:
-    branches:
-      - main
+    tags:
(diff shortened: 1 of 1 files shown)


Ruwann (approved):
LGTM, another option could be to replace version by the commit SHA when publishing to Test PyPI and still run on every push to `main`?

RobbeSneyders:
A commit SHA is unfortunately not a valid version. See [PEP 440](https://www.python.org/dev/peps/pep-0440/).

PR #1459, opened 2022-02-09
Add Python 3.9 to supported versions in setup.py

RobbeSneyders:
This PR updates our `setup.py` to correctly indicate support for Python 3.9 which we're already using as target version in our test pipeline.


Ruwann (approved):
Good catch, thanks!

PR #1463, opened 2022-02-16
Take into account (x-)nullable when validating defaults

RobbeSneyders:
Fixes # 1462.

CC: @nielsbox

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -94,15 +94,18 @@ def validate_writeOnly(validator, wo, instance, schema):
     yield ValidationError("Property is write-only")
 
 
+NullableTypeValidator = allow_nullable(Draft4Validator.VALIDATORS['type'])
+NullableEnumValidator = allow_nullable(Draft4Validator.VALIDATORS['enum'])
+
 Draft4RequestValidator = extend(Draft4Validator, {
-                                'type': allow_nullable(Draft4Validator.VALIDATORS['type']),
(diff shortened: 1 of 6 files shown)


Ruwann (commented):
Oops, overlooked something!

tests/fixtures/json_validation/swagger.yaml
```diff
@@ -67,3 +67,16 @@ paths:
           description: User object
           schema:
             $ref: '#/definitions/User'
+
+  /nullable_default:
```
Ruwann, line 71: Is this endpoint actually tested? The added test doesn't seem to call it?

RobbeSneyders: It's the loading of the spec, since the validation of the spec itself failed.

connexion/json_schema.py
```diff
@@ -94,15 +94,18 @@ def validate_writeOnly(validator, wo, instance, schema):
...
+                                'type': NullableTypeValidator,
+                                'enum': NullableEnumValidator,
                                 'required': validate_required,
                                 'readOnly': validate_readOnly})
 
 Draft4ResponseValidator = extend(Draft4Validator, {
-                                 'type': allow_nullable(Draft4Validator.VALIDATORS['type']),
-                                 'enum': allow_nullable(Draft4Validator.VALIDATORS['enum']),
+                                 'type': NullableTypeValidator,
+                                 'enum': NullableTypeValidator,
```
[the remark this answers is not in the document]
RobbeSneyders: Good catch

PR #1465, opened 2022-02-18
Bump Flask lower bound to 1.1.4

RobbeSneyders:
To prevent a dependency conflict between Flask and itsdangerous.
See https://github.com/pallets/flask/issues/4455

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,10 +30,15 @@ install_requires = [
 ]
 
 swagger_ui_require = 'swagger-ui-bundle>=0.0.2,<0.1'
-flask_require = 'flask>=1.0.4,<3'
+
+flask_require = [
+    'flask>=1.0.4,<3',
+    'itsdangerous>=0.24',
+]
 aiohttp_require = [
     'aiohttp>=2.3.10,<4',
-    'aiohttp-jinja2>=0.14.0,<2'
+    'aiohttp-jinja2>=0.14.0,<2',
+    'MarkupSafe>=0.23',
 ]
 
 tests_require = [
@@ -41,7 +46,7 @@ tests_require = [
     'pytest>=6,<7',
     'pytest-cov>=2,<3',
     'testfixtures>=6,<7',
-    flask_require,
+    *flask_require,
     swagger_ui_require
 ]
 
@@ -96,7 +101,7 @@ setup(
     license='Apache License Version 2.0',
     setup_requires=['flake8'],
     python_requires=">=3.6",
-    install_requires=install_requires + [flask_require],
+    install_requires=install_requires + flask_require,
     tests_require=tests_require,
     extras_require={
         'tests': tests_require,


davidism:
Preferably, bump it to 2.0.3, since the 1.x line is no longer supported.

RobbeSneyders:
I'm reluctant to drop support for flask 1.X on a minor version update @davidism.
We're planning a new major release where we can do this.

Seems like we have the same issue with jinja and markupsafe.
I see three options:
- Bump the lower bound for flask (1.1.4) and jinja (3.0.x)
- Fix upper bound for itsdangerous (2.0.1) and markupsafe (2.0.1) until Connexion 3.0
- Remove our test suites for the minimum versions. The issues can be fixed by the users themselves by pinning their dependencies in the way they choose. The test suites provide useful info when they break due to code changes though.

wdyt @Ruwann?

davidism:
Sorry, I didn't know the context of this specific PR, I just saw a link to it on one of the issues. You're right, you should wait until a non-patch release to bump minimum versions. Just wanted to make sure you were aware, thanks!

Ruwann (approved):
Thx!

RobbeSneyders:
The tests still fail with the current fix due to a similar issue with Jinja @Ruwann.
We need to decide how to fix the conflicts completely (see my options above).

Ruwann:
Aha, didn't see we were still having issues with the current patch. Now looking at the options.
Trying to understand better why this bump isn't sufficient per https://github.com/pallets/flask/issues/4455#issuecomment-1044275070

For bumping the minimum version of Flask to 2.x , I'm inclined to keep it for a major bump of connexion as well.

Ruwann:
- Seems like Flask 1.x and Jinja>=3 are actually incompatible? https://github.com/pallets/flask/blob/1.1.x/setup.py#L57
- While not ideal to keep these versions limited, it seems like the most sensible option for now to support both Flask 1.x and Flask 2.x, although this might lead to some future problems and might have to limit Flask 2.x as well if it updates its dependencies for these packages?
- For bumping the minimum version of Flask to 2.x , I'm inclined to keep it for a major bump of connexion as well. Although I'm not sure whether this is actually required for SemVer, I guess there are people relying on Flask-specific stuff that will break their application by going from Flask 1.x to Flask 2.x, so for connexion it's likely to be breaking for our users.

RobbeSneyders:
Ok, I think I found a good solution

The difficulty is that there are two sets of dependencies and subdependencies that work.

**Backwards compatible:**
```
flask>=1.0.4,<3
itsdangerous<=2.0.1
jinja2>=2.7,<4
markupsafe<=2.0.1
```
**Future proof:**
```
flask>=2,<3
itsdangerous>=2
jinja2>=3,<4
markupsafe>=2
```

If we go with one of the choices that change the dependency boundaries as mentioned above, we are effectively dropping support for one of these sets.  

The issue is that our `*-min` environments only install the minimum version of the dependencies that we list. So it will install for instance `flask==1.0.4`, but it will install `itsdangerous==2.1.0` since `flask` version 1.0.4 does not specify an upper boundary for itsdangerous.

By copying the subdependencies lower bounds into our `setup.py`, the `*-min` test environment does install the lower versions.

So while not every possible combination of versions listed in our `setup.py` will work, at least we maximize the amount of possible dependency sets that work, and the user can choose.

Ruwann:
Yes, agreed to try to maximize the amount of options that work. AFAICT, this also doesn't give any problems when installing connexion with either Flask 1.x or Flask 2.x out of the box.
We should be able to drop the extra minimum dependencies once we upgrade to `flask>2.0.3`

setup.py
```diff
@@ -30,10 +30,14 @@ def read_version(package):
 ]
 
 swagger_ui_require = 'swagger-ui-bundle>=0.0.2,<0.1'
-flask_require = 'flask>=1.0.4,<3'
+flask_require = [
```
Ruwann, line 33: Bit surprised that this doesn't give any errors when installing as line 48 adds `flask_require` as an element to a list.

Similarly, I would expect that line 103 also needs to be changed:
```
install_requires=install_requires + [flask_require],
```

Ruwann (changes requested):

setup.py
```diff
@@ -30,10 +30,14 @@ def read_version(package):
...
+    'flask>=1.0.4,<3',
+    'itsdangerous>=0.24'
```
Ruwann, line 35: minor nit: but adding a trailing comma to the last element of the list makes future diffs a bit nicer

Ruwann (approved):

PR #1467, opened 2022-02-18
[FIX] MethodViewResolver CamelCase Methods

### LINKED ISSUE #1466: MethodViewResolver inconsistent documentation/incorrect(?) behavior
### Description
The [API documentation](https://connexion.readthedocs.io/en/latest/autoapi/connexion/resolver/index.html#connexion.resolver.MethodViewResolver) suggests the endpoint `/foo_bar` would translate to `FooBarView`, but [the testing suite](https://github.com/spec-first/connexion/blob/a67baa5be02c0018a2f501534fb0899a5b4e06d2/tests/test_resolver_methodview.py#L39-L179) translates `example_method` to `Example_methodView`

### Expected behaviour
`/foo-bar` -> `FooBarView`
`/foo_bar` -> `FooBarView`

### Actual behaviour
`/foo-bar` -> `Foo_barView`

### Steps to reproduce

run tests in [test_resolver_methodview](https://github.com/spec-first/connexion/blob/a67baa5be02c0018a2f501534fb0899a5b4e06d2/tests/test_resolver_methodview.py#L39-L179)

### Additional info:

[In the pep8 standard, class names should be FooBar and not Foo_bar](https://www.python.org/dev/peps/pep-0008/#class-names)

jdkent:
Fixes #1466 .

dependent on #1465 

 - change `view_name` to be `CamelCase` instead of a mix of CamelCase and snake_case.
 - update tests to output CamelCase for MethodViewResolver

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -5,8 +5,9 @@ from the operations defined in the OpenAPI spec.
 
 import inspect
 import logging
-import re
 import sys
+
+from inflection import camelize
 
 import connexion.utils as utils
 from connexion.exceptions import ResolverError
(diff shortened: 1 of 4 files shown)


RobbeSneyders:
Thanks for the PR @jdkent!
We already depend on `inflection` for this kind of string manipulation, so we could use `inflection.camelize` instead. It takes into account some more edge cases. E.g. for 'test_API', your implementation will return `TestApi`, while `inflection.camelize` will return TestAPI.

jdkent:
awesome, didn't know about `inflection`, very useful! switched to using that package instead of regular expressions.

RobbeSneyders (approved):
Thanks @jdkent!

PR #1469, opened 2022-02-19
Remove sanitization of body properties

RobbeSneyders:
This was introduced in #1222. 
We should not sanitize the body schema, since we don't sanitize body arguments. This creates a mismatch when comparing both.


RobbeSneyders:
Pipelines are failing due to dependency issue, see #1465.

Ruwann (approved):
Thanks!

PR #1474, opened 2022-02-23
Rework required_scopes checking

### LINKED ISSUE #1423: Security hash maps not properly ORed when using OAuth scopes
### Description
Security specification handling does not seem to properly apply OpenAPI "OR" semantics when trying to specify that "scope 1" or "scope 2" are acceptable for a given endpoint.  See https://swagger.io/docs/specification/authentication/ for a description of how the security section specifies a list of hashmaps where the maps specify AND criteria but the list is a list maps that are ORed.

### Expected behaviour
I should be able to use OpenAPI security OR semantics to specify the same security scheme in different `security` entries with different scopes specified.

### Actual behaviour
When the same security scheme is listed in multiple entries in the security section, the last one appears to clobber the earlier one.

### Steps to reproduce
Consider the spec fragment: 
````
{
  "components": {
    "securitySchemes": {
      "OAuthAuthCode": {
        "type": "oauth2",
        "flows": {
          "authorizationCode": {
            "authorizationUrl": "https://my-provider/authorize",
            "tokenUrl": "https://my-provider/token",
            "scopes": {
              "my.scp1" : "my test scope 1",
              "my.scp2" : "my test scope 2"
            }
          }
        }
      }
    }
  },  
  "security": [
    {
      "OAuthAuthCode": ["my.scp1"]
    },
    {
      "OAuthAuthCode": ["my.scp2"]
    }
  ],
}
````

I would expect that endpoints using the default security would accept tokens with the `my.scp1` or `my.scp2` scope.  In practice, an access token with `my.scp1` is rejected, and only a token with `my.scp2` is accepted.

### Additional info:

Output of the commands:

- `python --version`
````
Python 3.9.4
````
````
Name: connexion
Version: 2.7.0
Summary: Connexion - API first applications with OpenAPI/Swagger and Flask
Home-page: https://github.com/zalando/connexion
Author: Zalando SE
Author-email: None
License: Apache License Version 2.0
Location: /venv/lib/python3.9/site-packages
Requires: jsonschema, clickclick, requests, inflection, openapi-spec-validator, PyYAML, flask
Required-by: foo-service
````

### PR DESCRIPTION
**Pull request by Ruwann**
Fixes #1423  .

Moves the required_scopes argument as current approach assumes there is one possible set of scopes per operation/security object.

I have changed the tested behaviour for [this test](https://github.com/spec-first/connexion/blob/main/tests/api/test_secure_api.py#L99).
If there is optional auth, if an incorrect API key is provided, the `verify_none` still passes and given that the security schemes are specified in OR fashion, the security check should pass.
(For example, changing the security schemes around in the spec ([link](https://github.com/spec-first/connexion/blob/main/tests/fixtures/secure_endpoint/swagger.yaml#L169-L170)) also causes the test to fail, which shouldn't happen as the order shouldn't matter)

### DIFF
--- a/connexion/security/async_security_handler_factory.py
+++ b/connexion/security/async_security_handler_factory.py
@@ -62,12 +62,12 @@ class AbstractAsyncSecurityHandlerFactory(AbstractSecurityHandlerFactory):
         return wrapper
 
     @classmethod
-    def verify_security(cls, auth_funcs, required_scopes, function):
+    def verify_security(cls, auth_funcs, function):
         @functools.wraps(function)
         async def wrapper(request):
             token_info = cls.no_value
             for func in auth_funcs:
-                token_info = func(request, required_scopes)
+                token_info = func(request)
                 while asyncio.iscoroutine(token_info):
                     token_info = await token_info
                 if token_info is not cls.no_value:
--- a/connexion/security/async_security_handler_factory.py
+++ b/connexion/security/async_security_handler_factory.py
@@ -66,16 +66,23 @@ class AbstractAsyncSecurityHandlerFactory(AbstractSecurityHandlerFactory):
         @functools.wraps(function)
         async def wrapper(request):
             token_info = cls.no_value
+            problem = None
             for func in auth_funcs:
-                token_info = func(request)
-                while asyncio.iscoroutine(token_info):
-                    token_info = await token_info
-                if token_info is not cls.no_value:
-                    break
+                try:
+                    token_info = func(request)
+                    while asyncio.iscoroutine(token_info):
+                        token_info = await token_info
+                    if token_info is not cls.no_value:
+                        break
+                except (OAuthProblem, OAuthResponseProblem, OAuthScopeProblem) as err:
+                    problem = err
 
             if token_info is cls.no_value:
-                logger.info("... No auth provided. Aborting with 401.")
-                raise OAuthProblem(description='No authorization token provided')
+                if problem is not None:
+                    raise problem
+                else:
+                    logger.info("... No auth provided. Aborting with 401.")
+                    raise OAuthProblem(description='No authorization token provided')
 
             # Fallback to 'uid' for backward compatibility
             request.context['user'] = token_info.get('sub', token_info.get('uid'))
--- a/connexion/security/async_security_handler_factory.py
+++ b/connexion/security/async_security_handler_factory.py
@@ -66,7 +66,7 @@ class AbstractAsyncSecurityHandlerFactory(AbstractSecurityHandlerFactory):
         @functools.wraps(function)
         async def wrapper(request):
             token_info = cls.no_value
-            problem = None
+            errors = []
             for func in auth_funcs:
                 try:
                     token_info = func(request)
@@ -74,12 +74,12 @@ class AbstractAsyncSecurityHandlerFactory(AbstractSecurityHandlerFactory):
                         token_info = await token_info
                     if token_info is not cls.no_value:
                         break
-                except (OAuthProblem, OAuthResponseProblem, OAuthScopeProblem) as err:
-                    problem = err
+                except Exception as err:
+                    errors += err
 
             if token_info is cls.no_value:
-                if problem is not None:
-                    raise problem
+                if errors != []:
+                    cls._raise_most_specific(errors)
                 else:
                     logger.info("... No auth provided. Aborting with 401.")
                     raise OAuthProblem(description='No authorization token provided')
--- a/connexion/security/async_security_handler_factory.py
+++ b/connexion/security/async_security_handler_factory.py
@@ -75,7 +75,7 @@ class AbstractAsyncSecurityHandlerFactory(AbstractSecurityHandlerFactory):
                     if token_info is not cls.no_value:
                         break
                 except Exception as err:
-                    errors += err
+                    errors.append(err)
 
             if token_info is cls.no_value:
                 if errors != []:
--- a/connexion/security/security_handler_factory.py
+++ b/connexion/security/security_handler_factory.py
@@ -173,10 +173,10 @@ class AbstractSecurityHandlerFactory(abc.ABC):
             raise OAuthProblem(description='Invalid authorization header')
         return auth_type.lower(), value
 
-    def verify_oauth(self, token_info_func, scope_validate_func):
+    def verify_oauth(self, token_info_func, scope_validate_func, required_scopes):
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: connexion/security/security_handler_factory.py [hunk 1]
```diff
@@ -320,6 +320,9 @@ def wrapper(request, *args, required_scopes=None):
             if need_to_add_required_scopes:
                 kwargs[self.required_scopes_kw] = required_scopes
             token_info = func(*args, **kwargs)
+            # TODO: Multiple OAuth schemes defined in OR fashion
+            # This currently doesn't work because if the first one doesn't apply, it will raise an error
+            # and the second OAuth security func is never checked
```
Ruwann, line 325: Remove TODO

connexion/security/security_handler_factory.py
```diff
@@ -362,18 +365,26 @@ def wrapper(request, token, required_scopes):
...
+            problem = None
             for func in auth_funcs:
-                token_info = func(request, required_scopes)
-                if token_info is not cls.no_value:
-                    break
+                try:
+                    token_info = func(request)
+                    if token_info is not cls.no_value:
+                        break
+                # TODO: Catch any error that might be raised instead of specific ones?
```
Ruwann, line 378: To check which errors should be caught here

connexion/security/security_handler_factory.py
```diff
@@ -362,18 +362,25 @@ def wrapper(request, token, required_scopes):
...
             for func in auth_funcs:
-                token_info = func(request, required_scopes)
-                if token_info is not cls.no_value:
-                    break
+                try:
+                    token_info = func(request)
+                    if token_info is not cls.no_value:
+                        break
+                except (OAuthProblem, OAuthResponseProblem, OAuthScopeProblem) as err:
+                    problem = err
```
RobbeSneyders, line 376: This can be overwritten in subsequent iterations of the for loop, only retaining the last error, no? Is this a problem?

Ruwann: Yes, you are correct: we can't return _all_ errors but wasn't sure what the best approach was here and whether it's enough to simply return _any_ error that was raised to at least return some info.

Some options:
- any/last specific error that was raised (current implementation)
- a generic `401`
- retain a `403` over a `401` as it's more specific (valid credentials but not enough privileges vs invalid/no credentials)

Ruwann: I'm also not yet sure which errors to catch: if a security view function raises a custom defined error, this isn't caught here and another passing security scheme is possibly never reached. Perhaps it's better to simply catch any error?

connexion/security/async_security_handler_factory.py
```diff
@@ -62,12 +62,12 @@ async def wrapper(request, token, required_scopes):
         return wrapper
 
     @classmethod
-    def verify_security(cls, auth_funcs, required_scopes, function):
+    def verify_security(cls, auth_funcs, function):
```
RobbeSneyders, line 65: Shouldn't we apply the same updates to this function? It overwrites the one in the `AbstractSecurityHandlerFactory`.

connexion/security/security_handler_factory.py
RobbeSneyders: It's a bit hard to wrap my head around this since I'm not very familiar with the security implementation of Connexion.

But if I understand correctly, we loop through `auth_funcs` here and catch the errors since we want to apply them in an `or` fashion?

Not sure if we want to catch any error here. It would make the security check more robust as one broken function would not limit the use of the others. On the other hand, it can hide issues longer.

If we do catch any error, we should probably prioritize which one to return. E.g. if you have two functions which would return a `403` and `500` status code, we should return the `403` status code, since it's probably more relevant to the security method you're trying to use.

RobbeSneyders (approved):
Thanks @Ruwann! LGTM.

SpudInNZ:
Hi

As a result of this patch, I am quite confused about the behaviour I am now seeing. We have an API with two possible security schemes, 

```
/accounts:
    get:
      security:
        - ApiKey: [ ]
        - OAuth2: [ accounts.read ]
....

components:
  securitySchemes:
    ApiKey:
      type: apiKey
      name: Authorization
      in: header
    OAuth2:
      type: oauth2
      flows: //etc

```
We have two token validation functions defined, which we specify using environment variables:

```
TOKENINFO_FUNC=app.security.validate_oauth2_token
APIKEYINFO_FUNC=app.security.validate_token
```
Calling the GET /accounts method with an OAUTH token causes `validate_token` to be called, with `required_scopes` set to `None`. Previously this method was called, but required_scopes was set to `accounts.read`. Neither now nor before the patch was `validate_oauth2_token` ever called, which we assumed was due to a bug in Connexion (but never tried to confirm this).

1. What is the expected behaviour in this scenario?
2. Is the order of the possible security schemes in the specification important?

Thanks

Ruwann:
Hi @SpudInNZ , I think there are two things here:

1. Security schemes that do not have scopes (all except OAuth2 and OIDC) no longer receive any scopes in the security handler function. The reason is that the scopes are not tied to an operation, but to a certain security scheme. For example, in case an operation for example multiple sets of scopes (e.g. 2 different OAuth2 schemes that use different scopes or in addition also OIDC), there is no clear answer on what the `required_scopes` should be (see also #1423 for context).

> For each scheme, you specify a list of security scopes required for API calls (see [below](https://swagger.io/docs/specification/authentication/#scopes)). Scopes are used only for OAuth 2 and OpenID Connect Discovery; other security schemes use an empty array [] instead.

2. The `validate_oauth2_token` function never being called. I'm not sure about this one: can you create a separate issue for this with a minimal reproducible example?

In theory, the order of the security schemes doesn't matter as it's an unordered mapping. However, in practice, connexion will first check the first security scheme, if that succeeds, we're done. If that fails, the second security scheme is checked. So it could have some impact depending on the performance if one function is more expensive than the other.

Ruwann:
Hi @vashek , we try to limit the breaking changes as much as possible, but we didn't anticipate breaking behaviour from this patch.
We had a discussion on this sort of cases, as this fixes a bug in connexion that led to this behaviour and whether to bump the major version.
What would have the ideal case for you here? Then we can see if and how we can do better in the future :)

To go further on what you aim to achieve and how it is breaking your application: as discussed in the comment above, there is undefined behaviour when there are multiple security schemes for which scopes are defined. For example,

```yaml
security:
    - oauth: [ read ]
    - oauth: [ admin ]
    - api_key: []
```

I don't see a clear answer to what should happen in this case. What would you like to happen here?
Note that this might be solved in OpenAPI 3.1, as there the scopes can also be defined for schemes other than OAuth2 and OIDC.

PR #1475, opened 2022-02-23
Move x-body-name to requestBody level

### LINKED ISSUE #1452: [BUG] x-body-name violates the OAS3 spec by being required to be within the schema field of media type

RobbeSneyders:
Fixes #1452  .

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -266,9 +266,21 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody schema
+          provide the x-body-name parameter in your requestBody media-type
+          (or non-OAS compliant legacy position within the requestBody schema)
           to override the name of the parameter that will be passed to your
           handler function.
+
+.. code-block:: yaml
+
+    requestBody:
+      content:
+        application/json:
+          x-body-name: stack
+          schema:
+            # legacy location here should be ignored because the preferred location is at the media-type level above
+            x-body-name: this_should_be_ignored
+            $ref: '#/components/schemas/someComponent'
 
 .. warning:: When you define a parameter at your endpoint as *not* required, and
     this argument does not have default value in your Python view, you will get
--- a/README.rst
+++ b/README.rst
@@ -267,7 +267,7 @@ of the endpoint parameter ``message`` to your view function.
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
           provide the x-body-name parameter in your operation 
-          (or non-OAS compliant legacy position within the requestBody schema)
+          (or legacy position within the requestBody schema)
           to override the name of the parameter that will be passed to your
           handler function.
 
@@ -276,8 +276,8 @@ of the endpoint parameter ``message`` to your view function.
 
     /path
       post:
-        x-body-name: body
         requestBody:
+          x-body-name: body
           content:
             application/json:
               schema:
--- a/README.rst
+++ b/README.rst
@@ -266,7 +266,7 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody media-type
+          provide the x-body-name parameter in your requestBody 
           (or non-OAS compliant legacy position within the requestBody schema)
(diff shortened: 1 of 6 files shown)


tests/fixtures/simple/openapi.yaml
```diff
@@ -271,10 +271,12 @@ paths:
         '200':
           description: OK
       requestBody:
+        x-body-name: stack
         content:
           application/json:
             schema:
-              x-body-name: stack
+              # should be ignored because the preferred location is at the media-type level above
```
Ruwann, line 278: `media-type` --> `operation`

Ruwann (changes requested):
Thanks for continuing this PR!
One small comment, then good to merge

README.rst
```diff
@@ -266,10 +266,25 @@ of the endpoint parameter ``message`` to your view function.
...
+
+
+    /path
+      post:
+        requestBody:
+          x-body-name: body
+          content:
+            application/json:
+              schema:
+                # legacy location here should be ignored because the preferred location for x-body-name is at the operation level above
```
Ruwann, line 284:
```suggestion
                # legacy location here should be ignored because the preferred location for x-body-name is at the requestBody level above
```

docs/request.rst
```diff
@@ -92,10 +92,23 @@ And the view function:
...
+.. code-block:: yaml
+
+    /path
+      post:
+        requestBody:
+          x-body-name: body
+          content:
+            application/json:
+              schema:
+                # legacy location here should be ignored because the preferred location for x-body-name is at the operation level above
```
Ruwann, line 109:
```suggestion
                # legacy location here should be ignored because the preferred location for x-body-name is at the requestBody level above
```

Ruwann (changes requested):

README.rst
```diff
@@ -266,10 +266,25 @@ of the endpoint parameter ``message`` to your view function.
 
 .. note:: In the OpenAPI 3.x.x spec, the requestBody does not have a name.
           By default it will be passed in as 'body'. You can optionally
-          provide the x-body-name parameter in your requestBody schema
+          provide the x-body-name parameter in your operation 
+          (or legacy position within the requestBody schema)
```
Ruwann, line 270: There was still one I missed 😅 Although not sure what the clearest explanation is here with `requestBody` and `requestBody schema`
```suggestion
          provide the x-body-name parameter in your requestBody
          (or legacy position within the requestBody schema)
```

RobbeSneyders: I think this is fine :slightly_frowning_face:

Ruwann (approved):

Ruwann:
If I saw correctly, we still have tests that use the legacy position only. (If not, it's good to also have a tests that uses the legacy position explicitly)

RobbeSneyders:
Indeed, it's still included in a lot of tests.
I added a `DeprecationWarning` so we can remove this in the next version.

connexion/operations/openapi.py
```diff
@@ -275,7 +275,15 @@ def _get_body_argument(self, body, arguments, has_kwargs, sanitize):
         if len(arguments) <= 0 and not has_kwargs:
             return {}
 
-        x_body_name = sanitize(self.body_schema.get('x-body-name', 'body'))
+        # prefer the x-body-name as an extension of requestBody
+        x_body_name = sanitize(self.request_body.get('x-body-name', None))
+
+        if not x_body_name:
+            # x-body-name also accepted in the schema field for legacy connexion compat
+            logger.warning('x-body-name within the requestBody schema will be deprecated in the '
```
Ruwann, line 283: No expert, but I think a DeprecationWarning might be the most pythonic option here?
https://docs.python.org/3/library/warnings.html#warnings.warn

```
warnings.warn(message, DeprecationWarning, stacklevel=2)
```

RobbeSneyders: Actually what I wanted to do :see_no_evil:

PR #1477, opened 2022-03-03
Proof of concept to move connexion functionality into middleware

RobbeSneyders:
This PR implements a proof of concept to move most of the Connexion functionality to ASGI middleware. This middleware can be wrapped around any ASGI (and WSGI by using an adapter) compliant framework, greatly increasing the compatibility of connexion.

There is one `ConnexionMiddleware` component, which manages a pluggable list of ASGI middlewares. Similar to the [middlewares mechanism of Django](https://docs.djangoproject.com/en/4.0/topics/http/middleware/).

image

Note that I used Starlette for the implementation, but as an ASGI toolkit, not as a web framework. The result is not limited to Starlette, but framework-agnostic.

You can find some example apps in the [`test_middleware`](https://github.com/spec-first/connexion/tree/feature/middleware/test_middleware) directory, showing compatibility with Flask, Quart, Sanic, and Starlette.

Eg. `starlette_app.py`:
```python
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route

from connexion.middleware import ConnexionMiddleware, TestMiddleware


# Simple Starlette app. Can be any ASGI framework.

async def test(request):
    value = request.query_params.get('int')

    if value == '1':
        value = int(value)

    return JSONResponse({'int': value})


app = Starlette(debug=True, routes=[
    Route('/test', test),
])


# Add connexion middleware.

middlewares = ConnexionMiddleware.default_middlewares + [TestMiddleware]

app = ConnexionMiddleware(app, middlewares=middlewares)
app.add_api('openapi.yaml')
```

This PR is quite big, but I split the work into logical commits. Once we want to start actually reviewing this, I'll submit them as separate PRs with more info.

### DIFF
--- a/connexion/apis/__init__.py
+++ b/connexion/apis/__init__.py
@@ -13,4 +13,5 @@ on the framework app.
 """
 
 
-from .abstract import AbstractAPI, AbstractMinimalAPI, AbstractSwaggerUIAPI  # NOQA
+from .abstract import (AbstractAPI, AbstractMinimalAPI,  # NOQA
+                       AbstractSwaggerUIAPI)
--- a/connexion/apis/__init__.py
+++ b/connexion/apis/__init__.py
@@ -13,4 +13,4 @@ on the framework app.
 """
 
 
-from .abstract import AbstractAPI  # NOQA
+from .abstract import AbstractAPI, AbstractMinimalAPI  # NOQA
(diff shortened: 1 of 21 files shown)


RobbeSneyders:
I've started extracting this into separate PRs. See #1492, #1496, #1497, and more will follow.

PR #1480, opened 2022-03-07
Declare setuptools dep as a dep

### LINKED ISSUE #1479: Connexion doesn't declare setuptools dep
### Description
Connexion doesn't declare setuptools as a dep, but it uses it

### Expected behaviour
connexion can successfully be imported

### Actual behaviour
ImportError: No module named pkg_resources

### Steps to reproduce
Build a [pex](https://pex.readthedocs.io/en/v2.1.70/) with connexion as a dep and try to import it

### Additional info:

Output of the commands:

- `python --version`
 Python 3.7.9
2.7.0

jmcrawford45:
Fixes #1479 .

 - Declare setuptools dep as a dep


RobbeSneyders:
Hi @jmcrawford45, thanks for the PR.
Seems like you're working on top of an outdated branch though. Could you start from the current main branch?

RobbeSneyders:
We should specify a minimum version like we do for the other requirements @jmcrawford45.

jmcrawford45:
> We should specify a minimum version like we do for the other requirements @jmcrawford45.

@RobbeSneyders I wasn't sure which versions we should target, so I picked a minimum from a similar date range to some of the other minimum dep requirements

RobbeSneyders:
Thanks @jmcrawford45.

Setuptools versions go quickly though, and they're often installed automatically, so I would remove the upper limit here. I think the risk of dependency conflicts here is a lot bigger than the risk of breaking changes.

I think we can lower the lower bound as well. I quickly tested it locally, and seems like version 17.0 is the lowest version that works.

RobbeSneyders:
Seems like there is a version conflict in the `importlib-metadata` dependency now.

In the meantime, there's already an open PR (https://github.com/spec-first/connexion/pull/1499) dropping `setuptools` as I recommended in the issue, so I suggest we go with that one instead.

Thanks anyway @jmcrawford45!

PR #1482, opened 2022-03-10
Save memory on unused schema strings

vmarkovtsev:
Nodoby needs the schema strings vs. the parsed contents, yet those strings consume quite some memory for big specs.


RobbeSneyders (approved):
Thx @vmarkovtsev!

Issue #1485, opened 2022-03-14
Basicauth documentation and example issues


a-a-abramov:
### Description
`examples/openapi3/basicauth` example is broken right now. I assume it was broken since #869 (commit 8830d56841d29fe2ddce9d8507d60a27594cff30) - the example tries to import `connexion.decorators.security` which is now deleted. There's also almost the same issue in the docs [here](https://connexion.readthedocs.io/en/latest/security.html#oauth-2-authentication-and-authorization) - `connexion.decorators.security.validate_scope` function is obviously deleted with the whole file. 

It's very easy to fix (I will create a PR once any feedback is given), but unfortunately this example has code that makes no sence - it sets and checks scopes, while Swagger/OpenAPI 3 docs clearly state, that this method does not support scoping. Your [basic auth documentation](https://connexion.readthedocs.io/en/latest/security.html#basic-authentication) implicitly says the same:

> If the security declaration of the operation also has an oauth security requirement, required_scopes is taken from there, otherwise it’s None. This allows authorizing individual operations with [oauth scope](https://oauth.net/2/scope/) while using basic authentication for authentication.

As I understand this you can define multiple security schemes and use them in conjunction (AND). And then authn from **basic** and authz with scopes from **oauth**. But I don't really see the point of this, because you actually have to perform authn twice - I found no way to bypass/ignore oauth authn in this case. You may implement service account approach, but basic authn in this case becomes obsolete. 

It would be awesome if you could provide example of basicauth + oauth in case I'm completely wrong! Or otherwise just remove this clause completely and state that it's impossible to use scopes unless oauth2 is used. I, once again, can fix this section of docs and create another PR.

### Expected behaviour
All examples work.
Documentation is unambiguous and clear for new users like myself.

### Actual behaviour
`openapi3/basicauth` example fails due to previous refactoring.
Documentation is lacking clarity or additional examples.

### Steps to reproduce
Run `openapi3/basicauth`

### Additional info:

```
$ python --version
Python 3.10.1
$ pip show connexion | grep "^Version\:"
Version: 2.12.0
```

Ruwann:
Hi @a-a-abramov , thanks for the detailed report!

Indeed, as you say, the outdated documentation is already reported in #1471 .

[...]

PR #1486, opened 2022-03-15
Remove obsolete code from basicauth example

### LINKED ISSUE #1471: OpenAPI basicauth example is outdated
The example references old code.

```
  File "app.py", line 7, in <module>
    from connexion.decorators.security import validate_scope
ModuleNotFoundError: No module named 'connexion.decorators.security'
```

a-a-abramov:
Fixes #1471 / #1485

 - Remove obsolete code (see the beginning of #1485 for concrete reasoning)
 - Remove scopes completely except necessary arg in `x-basicInfoFunc` implementation

### DIFF
--- a/examples/openapi3/basicauth/app.py
+++ b/examples/openapi3/basicauth/app.py
@@ -10,7 +10,7 @@ PASSWD={
     'foo': 'bar'
 }
 
-def basic_auth(username, password, required_scopes=None) -> dict:
+def basic_auth(username, password, required_scopes=None):
     if PASSWD.get(username) == password:
         return {'sub': username}
     return None
--- a/examples/openapi3/basicauth/app.py
+++ b/examples/openapi3/basicauth/app.py
@@ -5,14 +5,15 @@ Basic example of a resource server
 
 import connexion
 
-PASSWD={
+PASSWD = {
     'admin': 'secret',
     'foo': 'bar'
 }
 
-def basic_auth(username, password, required_scopes=None):
+def basic_auth(username, password):
     if PASSWD.get(username) == password:
         return {'sub': username}
+    # optional: raise exception for custom error response
     return None
 
 def get_secret(user) -> str:
--- a/examples/openapi3/basicauth/app.py
+++ b/examples/openapi3/basicauth/app.py
@@ -4,8 +4,6 @@ Basic example of a resource server
 '''
 
 import connexion
-from connexion.decorators.security import validate_scope
-from connexion.exceptions import OAuthScopeProblem
 
 
 def basic_auth(username, password, required_scopes=None):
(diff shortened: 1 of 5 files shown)


examples/openapi3/basicauth/app.py
```diff
@@ -4,29 +4,16 @@
...
+    'foo': 'bar'
+}
 
-def basic_auth(username, password, required_scopes=None):
-    if username == 'admin' and password == 'secret':
-        info = {'sub': 'admin', 'scope': 'secret'}
-    elif username == 'foo' and password == 'bar':
-        info = {'sub': 'user1', 'scope': ''}
-    else:
-        # optional: raise exception for custom error response
```
Ruwann, line 17: nit but can we leave this comment in here to indicate that this is also possible to do from a security function?

examples/openapi3/basicauth/app.py
```diff
@@ -4,29 +4,16 @@
...
-    if required_scopes is not None and not validate_scope(required_scopes, info['scope']):
-        raise OAuthScopeProblem(
-                description='Provided user doesn\'t have the required access rights',
-                required_scopes=required_scopes,
-                token_scopes=info['scope']
-            )
-
-    return info
-
+def basic_auth(username, password, required_scopes=None) -> dict:
```
Ruwann, line 13: You don't have to typehint the return type here, but if you want to, it should be `typing.Optional[dict]` as it's possible to return `None`

We also don't need the `required_scopes` argument here anymore as well (the example should still work if you leave it out)

Ruwann, line 13: You don't have to typehint the return type here, but if you want to, it should be `typing.Optional[dict]` as it's possible to return `None`

examples/openapi3/basicauth/app.py
```diff
@@ -4,29 +4,16 @@
 '''
 
 import connexion
-from connexion.decorators.security import validate_scope
-from connexion.exceptions import OAuthScopeProblem
 
+PASSWD={
```
Ruwann, line 8:
```suggestion
PASSWD = {
```

Ruwann, line 8:
```suggestion
PASSWD = {
```

Ruwann (changes requested):
Thanks for the PR! Left some small comments, but should be good to merge when addressed :)

There's also still a reference in the swagger2 basicauth example to the usage of scopes here:
https://github.com/spec-first/connexion/blame/main/examples/swagger2/basicauth/README.rst#L16-L19

We can just delete these lines from the README imo

examples/openapi3/basicauth/app.py
a-a-abramov: Ah, good catch. Totally missed it!

a-a-abramov:
I sort of made 1 more logical step and made app.py-s and README.rst-s almost identical for both swagger and openapi3. Code duplication bothered me a bit, but they are examples after all.

Ruwann (approved):
Awesome, thanks a lot!

I agree that the code duplication is not that bad for examples, it's better to have each example complete on its own.

a-a-abramov:
Cool, thanks a lot!

Issue #1489, opened 2022-03-17
Move functionality into pluggable ASGI middleware stack


RobbeSneyders:
> Currently, Connexion works by injecting its functionality in between the underlying framework app and the user view function, by wrapping the view function in decorators. (See our [ARCHITECTURE.rst](https://github.com/spec-first/connexion/blob/main/ARCHITECTURE.rst)).
> 
> We propose to move most of this functionality to ASGI middleware, which is wrapped around the framework app. This allows this functionality to be framework agnostic, and actually work with any ASGI (or WSGI by using an adapter) compliant framework. This means that you can add the power of Connexion to existing apps of any of these frameworks, without Connexion having to maintain a framework-specific interface.
> 
> ![image](https://user-images.githubusercontent.com/20990866/157504165-97397ba2-b662-4ea8-85c5-d6dc011e7a41.png)
> The only functionalities that cannot be moved to this framework agnostic middleware are the resolver to automatically map operations to python functions, and the automatic parameter unpacking which needs to remain a decorator since it needs to have access to the view function.

_Originally posted in https://github.com/spec-first/connexion/issues/1395#issuecomment-1063214142_

This also makes the functionality pluggable so built-in features can easily be disabled and additional functionality can easily be added.

I've implemented a proof of concept that shows this middleware concept in https://github.com/spec-first/connexion/pull/1477.

The goal is to:
1. Make the middleware separately available for usage with any ASGI / WSGI app

```Python
from connexion import ConnexionMiddleware
from framework import App

app = App()
app = ConnexionMiddleware(app, __name__, specification_dir='openapi/')
app.add_api('my_api.yaml')
```

2. Use it within the Connexion App so we can keep offering the current interface

```Python
from connexion import App

app = App(__name__, specification_dir='openapi/')
app.add_api('my_api.yaml')
```

*Tasks*
- [x] Create ConnexionMiddleware #1492
- [x] Extract SwaggerUI functionality into middleware #1496 
- [x] Add routing middleware #1497 
- [x] Extract security into middleware #1514 
- [x] Extract (de)serialization into middleware #1524
- [x] Extract validation into middleware #1525

PR #1492, opened 2022-03-17
Add empty connexion middleware

RobbeSneyders:
Part of #1489 

This PR adds the empty ConnexionMiddleware and wraps it around the FlaskApp using the `a2wsgi.WSGIMiddleware` adapter.

The ConnexionMiddleware and FlaskApp both implement the ASGI interface, and an ASGI server is the recommended way to run it. 

The `a2wsgi.ASGIMiddleware` adapter is wrapped around the ASGI middleware to make it WSGI compliant so it can be used by the development server and test client.

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -42,7 +42,7 @@ class FlaskApp(AbstractApp):
         app.url_map.converters['int'] = IntegerConverter
         return app
 
-    def apply_middleware(self):
+    def _apply_middleware(self):
         middlewares = [*ConnexionMiddleware.default_middlewares,
                        a2wsgi.WSGIMiddleware]
         self.middleware = ConnexionMiddleware(self.app.wsgi_app, middlewares=middlewares)
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -8,12 +8,14 @@ import pathlib
 from decimal import Decimal
 from types import FunctionType  # NOQA
 
+import a2wsgi
 import flask
 import werkzeug.exceptions
 from flask import json, signals
 
 from ..apis.flask_api import FlaskApi
 from ..exceptions import ProblemException
+from ..middleware import ConnexionMiddleware
 from ..problem import problem
 from .abstract import AbstractApp
 
@@ -28,8 +30,10 @@ class FlaskApp(AbstractApp):
 
         See :class:`~connexion.AbstractApp` for additional parameters.
         """
-        super().__init__(import_name, FlaskApi, server=server, **kwargs)
         self.extra_files = extra_files or []
+        self.middleware = None
+
+        super().__init__(import_name, FlaskApi, server=server, **kwargs)
 
     def create_app(self):
         app = flask.Flask(self.import_name, **self.server_args)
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -31,6 +32,38 @@ def json(self):
         return json.loads(self.text)
 
 
+def fixed_get_environ():
```
RobbeSneyders, line 35: The `WSGIMiddleware` and test client were unfortunately not compatible (see https://github.com/abersheeran/a2wsgi/issues/21, https://github.com/pallets/werkzeug/issues/2347). The issues are fixed, but I still had to patch the test client since the fixes are not released and I don't want to restrict the versions too much.

tests/api/test_errors.py
```diff
@@ -44,23 +44,23 @@ def test_errors(problem_app):
 
     get_problem = app_client.get('/v1.0/problem')  # type: flask.Response
     assert get_problem.content_type == 'application/problem+json'
-    assert get_problem.status_code == 418
```
RobbeSneyders, line 47: `a2wsgi.ASGIMiddleware` uses the builtin `http` module underneath, which only supports status code 418 on Python 3.9+.

connexion/apps/flask_app.py
```diff
@@ -38,6 +42,14 @@ def create_app(self):
         app.url_map.converters['int'] = IntegerConverter
         return app
 
+    def apply_middleware(self):
```
Ruwann, line 45: It feels a bit weird to have this as a public function but at the same time it's not straightforward to apply any other middleware. Now, it's a function that doesn't need to be exposed as it's only done at instance creation?

I would propose to make it either a private function, or to allow for an argument for the entire middleware stack or to extend the middleware stack.

setup.py
```diff
@@ -23,16 +23,18 @@ def read_version(package):
     'clickclick>=1.2,<21',
     'jsonschema>=2.5.1,<5',
     'PyYAML>=5.1,<7',
-    'requests>=2.19.1,<3',
+    'requests>=2.27,<3',
```
Ruwann, line 26: Why do we need to bump to 2.27?

connexion/middleware/main.py
```diff
@@ -0,0 +1,65 @@
+import pathlib
+import typing as t
+
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+class MissingMiddlewareError(Exception):
```
Ruwann, line 7: What's the reason for creating this exception class?

connexion/middleware/main.py
```diff
@@ -0,0 +1,65 @@
...
+        """High level Connexion middleware that manages a list o middlewares wrapped around an
+        application.
+
+        :param app: App to wrap middleware around.
+        :param middlewares: List of middlewares to wrap around app. The list should be ordered
+                            from outer to inner middleware.
+        """
+        if middlewares is None:
+            middlewares = self.default_middlewares
+        self.app, self.apps = self._apply_middlewares(app, middlewares)
```
Ruwann, line 30: If I understand correctly, the `apps` and `middlewares` will have the opposite direction, given that we iterate on the reversed middlewares?

setup.py
RobbeSneyders: Because of a sub-dependency which is shared by both `requests` and `starlette`. We'll be able to drop the `requests` dependency once we move security into the middleware stack though, since we'll only need the `AsyncSecurityHandlerFactory` then.

connexion/middleware/main.py
RobbeSneyders: Legacy of a previous implementation :)
Will remove it.

RobbeSneyders:
Accidentally merged this one due to rebase hell @Ruwann.
Will reopen.

PR #1495, opened 2022-03-17
Rework security handler factory into sync and async

RobbeSneyders:
Follows up on https://github.com/spec-first/connexion/pull/1491 which removed the `aiohttp_security_handler_factory`.

`flask_security_handler_factory` is actually not Flask specific, so we can refactor this into a generic sync and async class. I also lifted some methods out of the abstract class since they were overwritten in 1 of 2 subclasses.


tests/fixtures/schema_references/openapi.yaml
```diff
@@ -0,0 +1,19 @@
+openapi: 3.0.0
```
RobbeSneyders, line 1: Why are these in here?

PR #1496, opened 2022-03-17
Extract Swagger UI functionality into middleware

RobbeSneyders:
Part of https://github.com/spec-first/connexion/issues/1489

This PR extracts the Swagger UI functionality into a middleware.

The middleware is implemented with a starlette router on which the swagger ui endpoints are registered. For all other routes, the router acts ass a pass-through.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -43,6 +43,7 @@ class AbstractSpecAPI(metaclass=AbstractAPIMeta):
             *args,
             **kwargs
     ):
+        """Base API class with only minimal behavior related to the specification."""
         logger.debug('Loading specification: %s', specification,
                      extra={'swagger_yaml': specification,
                             'base_path': base_path,
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -32,7 +32,83 @@ class AbstractAPIMeta(abc.ABCMeta):
         cls._set_jsonifier()
 
 
-class AbstractAPI(metaclass=AbstractAPIMeta):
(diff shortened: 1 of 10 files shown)


examples/openapi3/helloworld/hello.py
```diff
@@ -8,5 +8,6 @@ def post_greeting(name: str) -> str:
 
 if __name__ == '__main__':
     app = connexion.FlaskApp(__name__, port=9090, specification_dir='openapi/')
-    app.add_api('helloworld-api.yaml', arguments={'title': 'Hello World Example'})
+    app.add_api('helloworld-api.yaml', arguments={'title': 'Hello World Example'},
+                options={"serve_spec": False})
```
RobbeSneyders, line 12: Probably just for local testing, should revert.

connexion/apis/abstract.py
```diff
@@ -32,7 +32,83 @@ def __init__(cls, name, bases, attrs):
         cls._set_jsonifier()
 
 
-class AbstractAPI(metaclass=AbstractAPIMeta):
+class AbstractSpecAPI(metaclass=AbstractAPIMeta):
```
Ruwann, line 35: Can you add a docstring for clarity on what the purpose of this class is?

Ruwann (approved):

connexion/apis/abstract.py
RobbeSneyders: Done.

We can probably remove some of these abstract API classes in the future, since they'll only be implemented for their respective middleware. For now I've kept them so it's clear which functionality is extracted from the `AbstractAPI` class.

PR #1497, opened 2022-03-17
Add routing middleware

RobbeSneyders:
Part of https://github.com/spec-first/connexion/issues/1489

This PR adds a routing middleware, which adds the connexion operation to the scope passed to the next middlewares.

The middleware is implemented with a starlette router on which the api endpoints are registered. For all other routes, the router acts ass a pass-through.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -43,7 +43,14 @@ class AbstractSpecAPI(metaclass=AbstractAPIMeta):
             *args,
             **kwargs
     ):
-        """Base API class with only minimal behavior related to the specification."""
+        """Base API class with only minimal behavior related to the specification.
+
+        :param specification: OpenAPI specification. Can be provided either as dict, or as path
+            to file.
+        :param base_path: Base path to host the API.
+        :param arguments: Jinja arguments to resolve in specification.
+        :param options: New style options dictionary.
+        """
         logger.debug('Loading specification: %s', specification,
                      extra={'swagger_yaml': specification,
                             'base_path': base_path,
@@ -121,14 +128,13 @@ class AbstractMinimalAPI(AbstractSpecAPI):
             **kwargs
     ) -> None:
         """Minimal interface of an API, with only functionality related to routing.
-        :param specification: OpenAPI specification. Can be provided either as dict, or as path
-        to file.
-        :param base_path: Base path to host the API.
-        :param arguments: Jinja arguments to resolve in specification.
+
         :param resolver: Callable that maps operationID to a function
         :param resolver_error_handler: Callable that generates an Operation used for handling
-        ResolveErrors
+            ResolveErrors
         :param debug: Flag to run in debug mode
+        :param pass_context_arg_name: If not None URL request handling functions with an argument
+            matching this name will be passed the framework's request context.
         """
         super().__init__(*args, **kwargs)
         self.debug = debug
@@ -228,11 +234,6 @@ class AbstractAPI(AbstractMinimalAPI, metaclass=AbstractAPIMeta):
         :param pythonic_params: When True CamelCase parameters are converted to snake_case and an underscore is appended
             to any shadowed built-ins
         :type pythonic_params: bool
-        :param options: New style options dictionary.
-        :type options: dict | None
-        :param pass_context_arg_name: If not None URL request handling functions with an argument matching this name
-            will be passed the framework's request context.
-        :type pass_context_arg_name: str | None
         """
         self.validator_map = validator_map
 
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -14,7 +14,7 @@ from ..exceptions import ResolverError
 from ..http_facts import METHODS
 from ..jsonifier import Jsonifier
 from ..lifecycle import ConnexionResponse
-from ..operations import make_operation
+from ..operations import AbstractOperation, make_operation
 from ..options import ConnexionOptions
 from ..resolver import Resolver
 from ..spec import Specification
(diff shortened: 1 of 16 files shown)


Ruwann:
Some more context forpossible issues of subclassing `starlette.BaseHTTPMiddleware`: https://github.com/encode/starlette/issues/1029

RobbeSneyders:
Thanks for the additional context @Ruwann, I was indeed already running into some of these issues.
I was able to refactor and simplify the approach of the routing middleware, and I removed our own BaseHTTPMiddleware.

Note that there's some quick fixes in this PR that I still want to tackle later. Right now, I'm trying to extract all functionality into middleware with minimal effort via maximal reuse of existing connexion code (like the Api and Operation classes), and  maximal usage of Starlette tools.

tests/api/test_errors.py
```diff
@@ -27,7 +27,6 @@ def test_errors(problem_app):
     error405 = json.loads(get_greeting.data.decode('utf-8', 'replace'))
     assert error405['type'] == 'about:blank'
     assert error405['title'] == 'Method Not Allowed'
-    assert error405['detail'] == 'The method is not allowed for the requested URL.'
```
RobbeSneyders, line 30: This is Flask specific and not set by Starlette.

tests/fakeapi/auth.py
```diff
@@ -13,15 +13,3 @@ def fake_json_auth(token, required_scopes=None):
         return json.loads(token)
     except ValueError:
         return None
-
-
-async def async_basic_auth(username, password, required_scopes=None, request=None):
```
RobbeSneyders, line 18: Just some cleanup, these functions are not actually used in the tests.

RobbeSneyders (commented):
This one is ready for review @Ruwann

connexion/middleware/routing.py
```diff
@@ -0,0 +1,172 @@
...
+    async def default_fn(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """Callback to call next app as default when no matching route is found."""
+        original_scope, *_ = _scope_receive_send.get()
+
+        api_base_path = scope.get('root_path', '')[len(original_scope.get('root_path', '')):]
+
+        if CONNEXION_CONTEXT not in original_scope:
+            original_scope[CONNEXION_CONTEXT] = {}
+
+        original_scope[CONNEXION_CONTEXT].update({
```
Ruwann, line 89: FYI, the `dict.setdefault()` simplifies this a bit (although perhaps a bit less untuitive as it's less known):
```
original_scope.setdefault(CONNEXION_CONTEXT, {}).update({'api_base_path': api_base_path})
```

connexion/middleware/routing.py
```diff
@@ -0,0 +1,173 @@
...
+            raise NotFoundProblem
+
+    async def default_fn(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """Callback to call next app as default when no matching route is found."""
+        original_scope, *_ = _scope_receive_send.get()
+
+        api_base_path = scope.get('root_path', '')[len(original_scope.get('root_path', '')):]
+
+        if CONNEXION_CONTEXT not in original_scope:
+            original_scope[CONNEXION_CONTEXT] = {}
```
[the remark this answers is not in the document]
RobbeSneyders: Didn't know about this. Makes sense.

connexion/apps/abstract.py
```diff
@@ -54,8 +55,11 @@ def __init__(self, import_name, api_cls, port=None, specification_dir='',
 
         self.server = server
         self.server_args = dict() if server_args is None else server_args
+
         self.app = self.create_app()
-        self.middleware = self._apply_middleware()
+
+        middlewares = middlewares or ConnexionMiddleware.default_middlewares
```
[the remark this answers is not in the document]
RobbeSneyders: Don't really see a use case for it, but the current code can indeed at least lead to some unexpected behavior. I'll put it as default for the argument or handle the `[]` explicitly.

connexion/apis/abstract.py
```diff
@@ -109,7 +109,106 @@ def add_swagger_ui(self):
...
+    ) -> None:
+        """Minimal interface of an API, with only functionality related to routing.
+        :param specification: OpenAPI specification. Can be provided either as dict, or as path
+        to file.
+        :param base_path: Base path to host the API.
+        :param arguments: Jinja arguments to resolve in specification.
+        :param resolver: Callable that maps operationID to a function
+        :param resolver_error_handler: Callable that generates an Operation used for handling
+        ResolveErrors
+        :param debug: Flag to run in debug mode
```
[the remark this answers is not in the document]
RobbeSneyders: Indeed, wanted to describe parameters of both this class and superclass, but seems like I forgot to update some. Will clean it up.

connexion/middleware/routing.py
RobbeSneyders: Not a fan of the one-liner, but can indeed use `setdefault` for the first part.

RobbeSneyders:
Is this one good to go @Ruwann?

tests/test_middleware.py
```diff
@@ -0,0 +1,44 @@
...
+    return build_app_from_fixture('simple', request.param, middlewares=middlewares)
+
+
+def test_routing_middleware(middleware_app):
+    app_client = middleware_app.app.test_client()
+
+    response = app_client.post("/v1.0/greeting/robbe")
+
+    assert response.headers.get('operation_id') == 'fakeapi.hello.post_greeting', \
+        response.status_code
```
Ruwann, line 44: Why the additional `response.status_code` here? Should this be a separate `assert`?

Ruwann (approved):

tests/test_middleware.py
RobbeSneyders: It just raises the status code as info if the test fails.

PR #1499, opened 2022-03-21
Drop pkg_resources

### LINKED ISSUE #1479: Connexion doesn't declare setuptools dep

Ruwann:
Fixes #1479 

 - Use `importlib.resources` to retrieve schemas
   - `read_text()` doesn't support files in directories, so we have to use the package name, but the package name was invalid due to the `.` in the directories, and also needed to add `__init__.py` files for it to work. 
   We can also use the `files()` function, which does support subdirectories, but this would require the external package until Python 3.9
 - Use `importlib.metadata` and `packaging` to retrieve, parse and compare versions
   - This can be dropped once we bump the minimum version of `jsonschema` to `3`

### DIFF
--- a/connexion/decorators/validation.py
+++ b/connexion/decorators/validation.py
@@ -8,9 +8,14 @@ import functools
 import logging
 from typing import AnyStr, Union
 
-import pkg_resources
+try:
+    from importlib.metadata import version
+except ImportError:
+    from importlib_metadata import version
+
 from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
 from jsonschema.validators import extend
+from packaging.version import Version
 from werkzeug.datastructures import FileStorage
 
 from ..exceptions import (BadRequestProblem, ExtraParameterProblem,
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Another option for the resources would be `pkgutil`. [I think it does what we need](https://stackoverflow.com/a/58941536/4098821).

Ruwann:
Thanks, @RobbeSneyders , I didn't know this existed. It does seem to be the solution here :)

RobbeSneyders (approved):
Thanks @Ruwann, looks good to me!

PR #1501, opened 2022-03-24
Fix uri parsing for query parameter with empty brackets

RobbeSneyders:
Fixes an issue where query parameters with empty brackets are not sanitized correctly.

In the example below, the `ids[]` parameter was sanitized to `ids_` since [this change](https://github.com/spec-first/connexion/pull/1408/files#diff-fdef5b8ce70b2c603bff264825d3b8baad76326f32a4e94455224f680f65b780L68-R68).
```yaml
parameters:
  - name: ids[]
    in: query
    type: array
      items: string
    collectionFormat: multi
```
```Python
?ids[]=a&ids[]=b
```

This PR fixes the sanitization so it's `ids[]` becomes `ids`.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -48,6 +48,17 @@ def snake_and_shadow(name):
     return snake
 
 
+def sanitized(name):
+    return name and re.sub('^[^a-zA-Z_]+', '',
+                           re.sub('[^0-9a-zA-Z_]', '',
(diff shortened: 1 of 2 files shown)


Ruwann (approved):
LGTM, thanks!

FWIW, not sure what the *correct* way should be to handle it in general, but because this one is breaking old behaviour, it should be reverted.

PR #1511, opened 2022-04-11
Add pre-commit hook

RobbeSneyders:
This PR adds a pre-commit hook which runs the static analysis checks on the changed code before committing. The hook is reused by tox so the steps only need to be defined once.

This is a prerequisite for https://github.com/spec-first/connexion/issues/1498.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,10 +39,9 @@ flask_require = [
 ]
 
 tests_require = [
-    'decorator>=5,<6',
     'pytest>=6,<7',
+    'pre-commit>=2,<3',
     'pytest-cov>=2,<3',
-    'testfixtures>=6,<7',
     *flask_require,
     swagger_ui_require
 ]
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -39,10 +39,9 @@ def read_version(package):
 ]
 
 tests_require = [
-    'decorator>=5,<6',
```
RobbeSneyders, line 42: Doesn't seem to be used.

setup.py
```diff
@@ -39,10 +39,9 @@ def read_version(package):
...
     'pytest>=6,<7',
+    'pre-commit>=2,<3',
     'pytest-cov>=2,<3',
-    'testfixtures>=6,<7',
```
RobbeSneyders, line 45: Doesn't seem to be used.

setup.py
```diff
@@ -92,7 +91,6 @@ def readme():
     url='https://github.com/zalando/connexion',
     keywords='openapi oai swagger rest api oauth flask microservice framework',
     license='Apache License Version 2.0',
-    setup_requires=['flake8'],
```
RobbeSneyders, line 95: No longer needed here since we don't run `flake8` via `setup.py` anymore.

Ruwann (approved):
Nice, thanks for updating the README as well!

PR #1514, opened 2022-04-19
Extract security to middleware

### LINKED ISSUE #1120: Is it possible to disable security?
I've used `connexion` for a few years now, and back then you didn't have support for JWT, so I've implemented my own authenticator.

The problem now is that I was trying to upgrade `connexion` and it fails because by default you don't allow the request to go through if there are invalid parameters in the security definitions like missing the x-tokenInfoFunc, etc.

I can see a hacky way of doing this, just provide a function that returns True.

But I was wondering if there isn't a way to simply disable the auth done by connexion, as I'm doing it myself leveraging the `before_request` in flask.

RobbeSneyders:
Part of #1489.
Fixes #1120.

This PR adds the SecurityMiddleware, which checks the security for incoming requests.

 - Extract security from `AbstractAPI` and `FlaskApi` to `SecurityApi`
 - Replace `werkzeug` exceptions with `starlette.HTTPException` subclasses
 - Remove `SecureOperation` base class and extract security from all operations to separate `SecurityOperation` class only used in middleware
 - Collapse all security handler factory classes into a single async `SecurityHandlerFactory` class.
 - Split security and general operation tests.

It's quite a big one, but unfortunately it wasn't possible to split this into smaller PRs that would still pass the tests.


connexion/__init__.py
```diff
@@ -7,24 +7,16 @@
...
 from .decorators.produces import NoContent  # NOQA
 from .exceptions import ProblemException  # NOQA
-# add operation for backwards compatibility
-from .operations import compat
 from .problem import problem  # NOQA
 from .resolver import Resolution, Resolver, RestyResolver  # NOQA
 from .utils import not_installed_error  # NOQA
 
-full_name = f'{__package__}.operation'
-sys.modules[full_name] = sys.modules[compat.__name__]
```
RobbeSneyders, line 25: Removed the compat when reworking the operation.

connexion/operations/abstract.py
```diff
@@ -390,6 +385,17 @@ def function(self):
 
         return function
 
+    @property
+    def _request_response_decorator(self):
```
RobbeSneyders, line 389: Lifted from the removed `SecureOperation` as it's not actually related to security.

tests/api/test_secure_api.py
```diff
@@ -36,7 +36,6 @@ def test_security(oauth_requests, secure_endpoint_app):
     assert get_bye_no_auth.status_code == 401
     assert get_bye_no_auth.content_type == 'application/problem+json'
     get_bye_no_auth_reponse = json.loads(get_bye_no_auth.data.decode('utf-8', 'replace'))  # type: dict
-    assert get_bye_no_auth_reponse['title'] == 'Unauthorized'
```
RobbeSneyders, line 39: `starlette.HTTPException` does not have a `title` attribute.

tests/conftest.py
```diff
@@ -69,32 +72,36 @@ def f(*args, **kwargs):
...
-                return FakeResponse(200, '{"uid": "test-user", "scope": ["myscope", "otherscope"]}')
-            if token in ["300", "is_not_invalid"]:
-                return FakeResponse(404, '')
-            if token == "has_scopes_in_scopes_with_s":
-                return FakeResponse(200, '{"uid": "test-user", "scopes": ["myscope", "otherscope"]}')
-        return url
-
-    monkeypatch.setattr('connexion.security.flask_security_handler_factory.session.get', fake_get)
+
+    class FakeClient:
```
RobbeSneyders, line 76: Refactored to work with `httpx` instead of `requests`.

tests/test_operation2.py
```diff
@@ -292,29 +289,34 @@ def test_operation(api, security_handler_factory):
                                   operation=op_spec,
                                   app_produces=['application/json'],
                                   app_consumes=['application/json'],
-                                  app_security=[],
-                                  security_definitions=SECURITY_DEFINITIONS_REMOTE,
                                   definitions=DEFINITIONS,
                                   parameter_definitions=PARAMETER_DEFINITIONS,
                                   resolver=Resolver())
-    assert isinstance(operation.function, types.FunctionType)
```
RobbeSneyders, line 300: Removed intermediate security checks for tests not related to security.

tests/test_operation2.py
```diff
@@ -362,75 +360,32 @@ def test_operation_composed_definition(api):
...
-    security_decorator = operation.security_decorator
-    assert len(security_decorator.args[0]) == 1
-    assert security_decorator.args[0][0] == 'verify_oauth_result'
-    verify_oauth.assert_called_with(math.ceil, api.security_handler_factory.validate_scope, ['uid'])
+    SecurityOperation(security_handler_factory=security_handler_factory,
+                      security=[{'oauth': ['uid']}],
+                      security_schemes=SECURITY_DEFINITIONS_LOCAL)
 
-    assert operation.method == 'GET'
```
RobbeSneyders, line 394: Use `SecurityOperation` and remove intermediate checks not related to security for tests related to security.

connexion/middleware/security.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def __init__(
+            self,
+            specification: t.Union[pathlib.Path, str, dict],
+            auth_all_paths: bool = False,
+            *args,
+            **kwargs
+    ):
+        super().__init__(specification, *args, **kwargs)
+        self.security_handler_factory = SecurityHandlerFactory('context')
```
RobbeSneyders, line 64: Hardcoded `pass_context_arg_name` here. 
Would propose to get rid of the parameter and choose a fixed name, but to do so in a follow-up PR.

connexion/middleware/security.py
```diff
@@ -0,0 +1,225 @@
...
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        try:
+            connexion_context = scope['extensions'][CONNEXION_CONTEXT]
+        except KeyError:
+            raise  # TODO
```
RobbeSneyders, line 38: Still need to define a custom error to raise here because of missing routing middleware.

connexion/middleware/exceptions.py
```diff
@@ -6,13 +6,38 @@
...
+            status=exception.status,
+            title=exception.title,
+            detail=exception.detail,
+            type=exception.type,
+            instance=exception.instance,
+            headers=exception.headers,
+            ext=exception.ext
+        )
+
+        return Response(
```
Ruwann, line 34: Shouldn't we use `JSONResponse` here instead of a regular `Reponse`?

connexion/security/security_handler_factory.py
```diff
@@ -351,43 +351,48 @@ def wrapper(request, token, required_scopes):
             if need_to_add_context:
                 kwargs[self.pass_context_arg_name] = request.context
```
Ruwann, line 352: In the verify_security function, we are no longer using `request.context` and use `request.scope['context']` - why this is not the case here?

Ruwann (commented):
First review - some initial comments/questions, but still need to go over it again and have a better look at `test_operation2.py`

connexion/apis/flask_api.py
```diff
@@ -39,16 +31,6 @@ def _set_blueprint(self):
         self.blueprint = flask.Blueprint(endpoint, __name__, url_prefix=self.base_path,
                                          template_folder=str(self.options.openapi_console_ui_from_dir))
 
-    def add_auth_on_not_found(self, security, security_definitions):
```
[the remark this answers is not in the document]
RobbeSneyders: The same functionality is implemented for the middleware in these lines:
https://github.com/spec-first/connexion/pull/1514/files#diff-a1af7a3afc2ba04857107ee0179f4de7486723ef00531fa1a3685bcf491e2a78R69-R72

It makes sense to extract it into an `add_auth_on_not_found` method there as well though, will update.

connexion/apis/flask_api.py
```diff
@@ -156,9 +138,9 @@ def get_request(cls, *args, **params):
 
         :rtype: ConnexionRequest
         """
-        context_dict = {}
-        setattr(flask._request_ctx_stack.top, 'connexion_context', context_dict)
         flask_request = flask.request
+        context_dict = flask_request.environ['asgi.scope'].get('context', {})
```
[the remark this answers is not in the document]
RobbeSneyders: It's added by [`a2wsgi.WSGIMiddleware`](https://github.com/abersheeran/a2wsgi/blob/918d34300a5a136e32b3d65f970f095ae4f1b91f/a2wsgi/wsgi.py#L85).

connexion/security/security_handler_factory.py
```diff
@@ -351,43 +351,48 @@ def wrapper(request, token, required_scopes):
...
                 else:
                     logger.info("... No auth provided. Aborting with 401.")
-                    raise OAuthProblem(description='No authorization token provided')
+                    raise OAuthProblem(detail='No authorization token provided')
 
             # Fallback to 'uid' for backward compatibility
-            request.context['user'] = token_info.get('sub', token_info.get('uid'))
-            request.context['token_info'] = token_info
-            return function(request)
+            request.scope['context'] = {
```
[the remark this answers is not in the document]
RobbeSneyders: It's unused, but I'll wrap it a a bit more explicitly.

connexion/security/security_handler_factory.py
RobbeSneyders: I think it should be updated here as well indeed.

connexion/middleware/security.py
```diff
@@ -0,0 +1,229 @@
...
+                                    'you have a routing middleware registered upstream. ')
+
+        api_base_path = connexion_context.get('api_base_path')
+        if api_base_path:
+            api = self.apis[api_base_path]
+            operation_id = connexion_context.get('operation_id')
+            try:
+                operation = api.operations[operation_id]
+            except KeyError:
+                pass
```
nielsbox, line 47: What is the reason why we silently skip the Security check when the operation_id was not found in the registered APIs? Would still put a warning/info log here, wdyt?

connexion/middleware/security.py
```diff
@@ -0,0 +1,229 @@
...
+        if auth_all_paths:
+            self.add_auth_on_not_found()
+        else:
+            self.operations: t.Dict[str, SecurityOperation] = {}
+
+        self.add_paths()
+
+    def add_auth_on_not_found(self):
+        """Register a default SecurityOperation for routes that are not found."""
+        self.operations = defaultdict()
```
nielsbox, line 78: Would define the type here too, something like this?
```suggestion
        self.operations: t.Dict[str, SecurityOperation] = defaultdict()
```

connexion/middleware/security.py
```diff
@@ -0,0 +1,229 @@
...
+        default_operation = self.make_operation()
+        self.operations.default_factory = lambda: default_operation
+
+    def add_paths(self):
+        paths = self.specification.get('paths', {})
+        for path, methods in paths.items():
+            for method, operation in methods.items():
+                if method not in METHODS:
+                    continue
```
nielsbox, line 87: Wouldn't it make more sense to handle this more upstream when parsing the spec? If not, we should warn here that a method is being skipped.

nielsbox (commented):
Reviewed half of the PR, left some questions and nitpicks. Will do the rest this weekend.

connexion/middleware/security.py
```diff
@@ -0,0 +1,229 @@
...
+        api = SecurityAPI(specification, **kwargs)
+        self.apis[api.base_path] = api
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        try:
+            connexion_context = scope['extensions']['connexion_routing']
```
nielsbox, line 35: Would like to see this extension/context name defined in the RoutingMiddleware as I see this is directly  linked with that. For example RoutingMiddleware.CONTEXT_NAME. Means less prone to typos and auto complete.

connexion/middleware/security.py
RobbeSneyders: This is validated when parsing the spec, but there are other valid keys at this level. For instance for [common parameters](https://swagger.io/docs/specification/describing-parameters/) for a path. We want to skip those.

connexion/middleware/security.py
RobbeSneyders: The `operation_id` can be `None` for operations not specified in the spec. A default `SecurityOperation` is only registered when using `auth_all_paths`, so this can throw a `KeyError`.

Refactored this and included debug log and a check for unknown `operation_id`s.

connexion/middleware/exceptions.py
```diff
@@ -1,19 +1,42 @@
...
+            headers=exception.headers,
+            ext=exception.ext
+        )
+
+        return JSONResponse(
+            content=connexion_response.body,
+            status_code=connexion_response.status_code,
+            media_type=connexion_response.mimetype,
+            headers=connexion_response.headers
+        )
```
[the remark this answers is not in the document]
RobbeSneyders: Makes sense. Reverting to regular `Response`.
CC: @Ruwann

connexion/middleware/security.py
RobbeSneyders: Refactored this a bit into something more clear and where types should be able to be deducted automatically. Would also only focus on typing details when we start implementing actual `mypy` checks.

connexion/middleware/security.py
RobbeSneyders: Good point, lost this while refactoring.

RobbeSneyders:
Thanks @Ruwann @nielsbox, all comments should be addressed.

nielsbox (approved):

PR #1521, opened 2022-05-04
Only log x-body-name deprecation warning when used incorrectly

### LINKED ISSUE #1520: x-body-name deprecation warning though it's not present
### Description

connexion is used in Apache Airflow API module. The spec file is present here : https://github.com/apache/airflow/blob/673f8940786b33663a5eb47047af2d7c2a14df99/airflow/api_connexion/openapi/v1.yaml . The tests are generating below warning though `x-body-name` is not used in the spec file.

```
tests/api_connexion/endpoints/test_dag_run_endpoint.py: 12 warnings
  /usr/local/lib/python3.7/site-packages/connexion/operations/openapi.py:286: DeprecationWarning: x-body-name within the requestBody schema will be deprecated in the next major version. It should be provided directly under the requestBody instead.
    'the requestBody instead.', DeprecationWarning)
```

Here the condition is `if not x_body_name` so I assume this raises DeprecationWarning even when the value is `None`.

https://github.com/spec-first/connexion/blob/972d1a9f9c113a2deedec7b2e345a4983293a63c/connexion/operations/openapi.py#L274-L282

### Additional info:

Output of the commands:

- `python --version` Python 3.7.13
- `pip show connexion | grep "^Version\:"` Version: 2.13.0

RobbeSneyders:
Fixes #1520

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -282,10 +282,12 @@ class OpenAPIOperation(AbstractOperation):
 
         if not x_body_name:
             # x-body-name also accepted in the schema field for legacy connexion compat
-            warnings.warn('x-body-name within the requestBody schema will be deprecated in the '
-                          'next major version. It should be provided directly under '
-                          'the requestBody instead.', DeprecationWarning)
             x_body_name = sanitize(self.body_schema.get('x-body-name', 'body'))
+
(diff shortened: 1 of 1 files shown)


Ruwann (approved):
Thanks, LGTM!

PR #1540, opened 2022-05-17
Removed jsonschema version check

### LINKED ISSUE #1537: Remove jsonschema version check
Now that we have bumped the `jsonschema` version to `4.x` in #1536 , we can also remove the code that checks whether `jsonschema` is version 3 or higher.

https://github.com/spec-first/connexion/blob/156bf79ee68a82feaf02840f5ba6d1f3ee92291c/connexion/decorators/validation.py#L28

and 

https://github.com/spec-first/connexion/blob/156bf79ee68a82feaf02840f5ba6d1f3ee92291c/connexion/decorators/validation.py#L283-L295

As a result, I believe we can also drop the dependency on `importlib_metadata`.

jonasboecquaert:
Fixes #1537

 - Remove jsonschema version checks
 - Remove dependency on `importlib_metadata`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,6 @@ install_requires = [
     'requests>=2.27,<3',
     'inflection>=0.3.1,<0.6',
     'werkzeug>=2,<3',
-    'packaging>=20',
     'starlette>=0.15,<1',
     'httpx>=0.15,<1',
 ]
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,6 @@ install_requires = [
     'requests>=2.27,<3',
     'inflection>=0.3.1,<0.6',
     'werkzeug>=2,<3',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,7 +26,6 @@ def read_version(package):
     'requests>=2.27,<3',
     'inflection>=0.3.1,<0.6',
     'werkzeug>=2,<3',
-    'importlib-metadata>=1 ; python_version<"3.8"',
     'packaging>=20',
```
RobbeSneyders, line 29: I think we can drop `packaging` as a dependency as well.

RobbeSneyders (commented):
Thanks @jonasboecquaert

setup.py
jonasboecquaert: I'll remove it 👍

RobbeSneyders (approved):

Issue #1114, opened 2019-12-30
bug of the function is_json_mimetype() in the file connexion/utils.py


adf2007:
### Description
json mimetype may like these:
   application/json
   application/vnd.com.myEntreprise.v6+json
   application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0
   application/vnd.com.myEntreprise.v6+json; charset=UTF-8
   ...

  def is_json_mimetype(mimetype):
    """
    :type mimetype: str
    :rtype: bool
    """
    maintype, subtype = mimetype.split('/')  # type: str, str
    return maintype == 'application' and (subtype == 'json' or subtype.endswith('+json'))

   The function can only recognize the 1st and 2nd case, but can't recognize the 3rd and 4th case.

Output of the commands:
-  python3 --version
Python 3.6.8
- `pip show connexion | grep "^Version\:"`
Version: 2.5.1

PR #1541, opened 2022-05-19
Fix for bug of the function is_json_mimetype() (#1114)

jacobstanly89:
Fixes # .

- Modified check for '+json' in subtype

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -132,7 +132,7 @@ def is_json_mimetype(mimetype):
     :rtype: bool
     """
     maintype, subtype = mimetype.split('/')  # type: str, str
-    return maintype == 'application' and (subtype == 'json' or subtype.endswith('+json'))
+    return maintype == 'application' and (subtype == 'json' or '+json' in subtype)
 
 
 def all_json(mimetypes):


connexion/utils.py
```diff
@@ -137,7 +137,7 @@ def is_json_mimetype(mimetype):
     :rtype: bool
     """
     maintype, subtype = mimetype.split('/')  # type: str, str
```
Ruwann, line 139: A MIME type has the following structure:
```
type/subtype;parameter=value
```
So instead of changing the check itself, we should extract the subtype in the correct way by accounting for the `;` and splitting on it

Ruwann (commented):
Thanks for the PR!

Could you also add a test for this? It can be added in `tests/test_utils.py`

connexion/utils.py
jacobstanly89: Thanks for the feedback. I will make the update later this week.

jacobstanly89: @Ruwann tests are now added and the proposed changes are done.

connexion/utils.py
```diff
@@ -136,7 +136,10 @@ def is_json_mimetype(mimetype):
     :type mimetype: str
     :rtype: bool
     """
+
     maintype, subtype = mimetype.split('/')  # type: str, str
+    if ';' in subtype:
+        subtype, parameter = subtype.split(';')
```
Ruwann, line 142:
```suggestion
        subtype, parameter = subtype.split(';', maxsplit=1)
```

Ruwann (approved):
Thanks a lot, @jacobstanly89 !

PR #1550, opened 2022-06-09
Avoid false positive deprecation warning

lanzkron:
Fixes # .

Do not warn about the schema's `x-body-name` being deprecated unless it is being used.

 - Code behaviour is unchanged, if body name is set use it, otherwise fallback to the schema's (deprecated) name or `"body"` if nothing is set 
 - If `x-body-name` is set in the schema, warn about deprecation and use (unless overridden by the request body).


RobbeSneyders:
Hi @lanzkron.
Thanks for the PR, but this was already fixed on the v2 branch (https://github.com/spec-first/connexion/pull/1521) and we'll cherry pick this to main. It hasn't been released yet, but I'll prepare a new release now.

RobbeSneyders:
Actually, the fix I mentioned just moves the issue.

RobbeSneyders:
@lanzkron I changed the target branch to v2. Could you remove the commits from `main` and rebase on `v2`?

lanzkron:
I seem to have messed up the rebase. I'll close this PR and create a clean one on v2.
BTW the reason #1550 didn't work was because the second `get` gave a default `"body"` so it was never `None`.

PR #1552, opened 2022-06-11
call as_view in methodresolver

### LINKED ISSUE #1549: support method view decorator
Hi, I am trying to use `MethodView` decorator. 

I found out that the regular decorator list from the `MethodView` is not called, since .as_view is never called. 

I can see in one of the example, there is a field called method_decorators  but I cannot find any reference in the code base. It appears to be unused. https://github.com/spec-first/connexion/blob/b561ecfdaacd7606d1398512762b946e2d778c74/examples/openapi3/methodresolver/api/petsview.py#L11

Is this possible?

If its not possible, we could maybe add it to the `MethodResolver`. I managed to kind of get it working by modifying the code.

```python
if view is None:
    view = view_cls()
    self.initialized_views.append(view)
func = getattr(view, meth_name)

# I have added this:
# if the view has decorators, apply them to the function
# since .as_view() is never called
if view.decorators:
    for decorator in view.decorators:
        func = decorator(func)

# Return the method function of the class
return func
```

I have added this to the code here https://github.com/spec-first/connexion/blob/b561ecfdaacd7606d1398512762b946e2d778c74/connexion/resolver.py#L241

bluebrown:
Signed-off-by: Nico Braun <rainbowstack@gmail.com>

Fixes #1549 

>**Warning** 
>this is a breaking change
> users will be required to change any search method to a get method, merging _get by id_ and _get collection_ into a single handler

The primary reason for this change is to be able to use method decorators as described here: https://flask.palletsprojects.com/en/2.1.x/views/#decorating-views

 - change the `MethodResolver` to call `as_view` on the classes

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -192,15 +192,8 @@ class MethodViewResolver(RestyResolver):
                     return ...
     """
 
-    def __init__(self, *args,  class_arguments={}, **kwargs):
-        """
-        :param class_arguments: args and kwargs to pass to the class constructor
-        :type class_arguments: dict[str, dict[str, tuple[Any] | dict[str, Any]]]
-        example:
-        >>> class_arguments = {'MyView': 'args': ('foo',), 'kwargs': {'bar': 'baz'}}
-        """
-
-        self.class_arguments = class_arguments
+    def __init__(self, *args,  class_arguments=None, **kwargs):
+        self.class_arguments = class_arguments or {}
         super(MethodViewResolver, self).__init__(*args, **kwargs)
         self.initialized_views = []
 
@@ -263,7 +256,7 @@ class MethodViewResolver(RestyResolver):
             # get the args and kwargs for this view
             cls_arguments = self.class_arguments.get(view_name, {})
             cls_args = cls_arguments.get('args', ())
-            cls_kwargs= cls_arguments.get('kwargs', {})
+            cls_kwargs = cls_arguments.get('kwargs', {})
             # call as_view to get a view function
             # that is decorated with the classes
             # decorator list, if any
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -192,7 +192,15 @@ class MethodViewResolver(RestyResolver):
                     return ...
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args,  class_arguments={}, **kwargs):
+        """
+        :param class_arguments: args and kwargs to pass to the class constructor
+        :type class_arguments: dict[str, dict[str, tuple[Any] | dict[str, Any]]]
+        example:
+        >>> class_arguments = {'MyView': 'args': ('foo',), 'kwargs': {'bar': 'baz'}}
+        """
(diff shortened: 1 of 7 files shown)


bluebrown:
I did not really find out how to write a test for this. All tests have passed. I think this kind of functionality is currently not covered. Like there is no app created to see what the methods return. Or perhaps, I overlooked it.

bluebrown:
We, should probably also update the documentation on this.

RobbeSneyders (commented):
Thanks @bluebrown.

- The changes look good to me in general. I left one small comment on the example.
- We should indeed update the docs on this. Could you have a look?
- As mentioned in the issue, it would be good to still offer a resolver with the old behavior, but we can mark it deprecated.
- The example currently doesn't run on main since we're in the middle of a large refactoring (https://github.com/spec-first/connexion/issues/1489), but the resolver itself does seem to work correctly. I submitted a PR which fixes the blocking issue (https://github.com/spec-first/connexion/pull/1553).
- The current resolver tests indeed only check the generated `operation_id`, which is unchanged. I'll have a deeper look at the tests later to see how we can best add a test for the changed behavior.

examples/openapi3/methodresolver/api/petsview.py
```diff
@@ -5,10 +5,17 @@
...
+
+
 class PetsView(MethodView):
     """ Create Pets service
     """
-    method_decorators = []
+    decorators = [example_decorator]
     pets = {}
 
     def post(self):
```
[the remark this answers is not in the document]
bluebrown: I have updated the example to use the body param in post and put

bluebrown:
- added a way to use the old resolver as `DeprecatedMethodViewResolver`
- added a `class_arguments` parameter to the new `MethodViewResolver` to allow dependency injection
- updated the tests to test both resolvers with the existing tests via parameterized fixture
- updated the example app
- updated the docs

connexion/resolver.py
```diff
@@ -192,7 +192,8 @@ def post(self):
                     return ...
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args,  class_arguments=None, **kwargs):
+        self.class_arguments = class_arguments or {}
```
RobbeSneyders, line 196: Could we name this a bit more specific eg. `view_arguments`?
It would also be good to document it here as an argument on top of the documentation you already added.

RobbeSneyders (commented):
Thanks for the changes @bluebrown.
I left one comment, but overall changes look great.

I still need to find the time to have a look at how we can add an additional test for the resulting behavior,

connexion/resolver.py
bluebrown: I think class_arguments is more fitting because the arguments for the view function are also called `class_args` and `class_kwargs` that's because these arguments are passed to the class when instantiating it.

bluebrown: I will try to add a comment to the init function. I had actually one already with example usage, but pre-commit hook didnt like it.

PR #1553, opened 2022-06-11
Use resolver in security middleware

RobbeSneyders:
Continues on https://github.com/spec-first/connexion/pull/1514.

This PR continues the implementation of the security middleware, which was started in https://github.com/spec-first/connexion/pull/1514. I took some shortcuts in the original PR, one of them the naive 'resolution' of the `operation_id` by fetching it from the spec instead of relying on a resolver.

https://github.com/spec-first/connexion/blob/b561ecfdaacd7606d1398512762b946e2d778c74/connexion/middleware/security.py#L93

This PR replaces this with resolution via resolver.

You'll see that I rely on the original Connexion `Operation` classes for this, instead of on the `Resolver` directly. I propose to further hollow out the `Operation` classes during the refactoring into a middleware stack, until it becomes a 'data class' providing an interface for the specification of the operation across openapi versions. This is similar to the `Specification` class, but on an operation level.

Specific middleware operations such as `SecurityOperation` and `RoutingOperation` can then use this `Operation` object to initialize themselves. I added this for the `RoutingOperation` as a separate commit.

### DIFF
--- a/connexion/apis/abstract.py
+++ b/connexion/apis/abstract.py
@@ -38,6 +38,8 @@ class AbstractSpecAPI(metaclass=AbstractAPIMeta):
             self,
             specification: t.Union[pathlib.Path, str, dict],
             base_path: t.Optional[str] = None,
+            resolver: t.Optional[Resolver] = None,
+            resolver_error_handler: t.Optional[t.Callable] = None,
             arguments: t.Optional[dict] = None,
             options: t.Optional[dict] = None,
             *args,
@@ -48,6 +50,9 @@ class AbstractSpecAPI(metaclass=AbstractAPIMeta):
         :param specification: OpenAPI specification. Can be provided either as dict, or as path
             to file.
         :param base_path: Base path to host the API.
+        :param resolver: Callable that maps operationID to a function
+        :param resolver_error_handler: Callable that generates an Operation used for handling
+            ResolveErrors
         :param arguments: Jinja arguments to resolve in specification.
         :param options: New style options dictionary.
         """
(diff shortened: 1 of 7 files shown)


connexion/apis/abstract.py
```diff
@@ -48,6 +49,9 @@ def __init__(
         :param specification: OpenAPI specification. Can be provided either as dict, or as path
             to file.
         :param base_path: Base path to host the API.
+        :param resolver: Callable that maps operationID to a function
+        :param resolver_error_handler: Callable that generates an Operation used for handling
+            ResolveErrors
```
Ruwann, line 54:
```suggestion
            ResolverErrors
```
😛

connexion/middleware/security.py
```diff
@@ -81,30 +82,36 @@ def __init__(
...
-            security = operation_spec.get('security', self.app_security)
-
-        return SecurityOperation(
-            self.security_handler_factory,
-            security=security,
-            security_schemes=self.specification.security_definitions
+                try:
+                    self.add_operation(path, method)
+                except ResolverError:
+                    # ResolverErrors are handled in routing middleware
```
Ruwann, line 97: Why don't we propagate the error in this case?

If the resolving fails, it is normally already raised in the routing middleware, or am I mistaken?
But if it is already raised there, then we don't need to account for it here?

So, not sure what the flow is here :)

Ruwann (approved):
Thanks, LGTM! Had 1 question though.

I really like the idea of making the operation class a simple class that only contains functionality to the operation object as specified in the spec (e.g. fetch parameters from the common definitions, or in case of security, get the applicable security for this operation in case it's defined both globally and on the operation).

connexion/middleware/security.py
RobbeSneyders: Resolution is done at start up. So 
- if the `ResolverError` is not handled in the routing middleware, start up will fail and this code will indeed never be reached.
- If the `ResolverError` is handled in the routing middleware, it will still be raised here during start-up, but a request will never reach the operation since it will already be handled by the `resolver_error_handler` in the routing middleware.

I can add some more explanation to the comment, but open to suggestions on how to handle it more clearly in code.

connexion/apis/abstract.py
RobbeSneyders: I copy pasted this :stuck_out_tongue:

PR #1556, opened 2022-06-17
debug mode interaction for non-flask server

thvu11:
Fixes the interaction of the debug mode with available server options.

When having `debug=True` and specify `server` argument, the application currently has the interaction as below:
- `server='flask'`: this works correctly and the application is updated when users edit a file
- `server='tornado'`: no update when users edit a file
- `server='gevent'`: no update when users edit a file

 - for tornado, `tornado.autoreload` is enabled if debug is true
 - for gevent, because it doesn't support automatic detection for code changes, the application'll display a log message suggesting users to try other options

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -140,6 +140,7 @@ class FlaskApp(AbstractApp):
                          extra_files=self.extra_files, **options)
         elif self.server == 'tornado':
             try:
+                import tornado.autoreload
                 import tornado.httpserver
                 import tornado.ioloop
                 import tornado.wsgi
@@ -148,6 +149,8 @@ class FlaskApp(AbstractApp):
             wsgi_container = tornado.wsgi.WSGIContainer(self.app)
             http_server = tornado.httpserver.HTTPServer(wsgi_container, **options)
             http_server.listen(self.port, address=self.host)
+            if self.debug:
+                tornado.autoreload.start()
             logger.info('Listening on %s:%s..', self.host, self.port)
             tornado.ioloop.IOLoop.instance().start()
         elif self.server == 'gevent':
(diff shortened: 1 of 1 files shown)


connexion/apps/flask_app.py
```diff
@@ -148,13 +149,17 @@ def run(self,
...
+                tornado.autoreload.start()
             logger.info('Listening on %s:%s..', self.host, self.port)
             tornado.ioloop.IOLoop.instance().start()
         elif self.server == 'gevent':
             try:
                 import gevent.pywsgi
             except ImportError:
                 raise Exception('gevent library not installed')
+            if self.debug:
+                logger.warning("gevent server doesn't support debug mode. Please switch to flask/tornado server.")
```
Ruwann, line 162: Haven't tested this myself yet, but is it possible to enable debug mode by setting it on `self.app` that is getting passed to the gevent server?

```
self.app.debug = True
```

Ruwann (approved):
Thanks for the PR!

Left one comment, maybe we can make it work on gevent. Do you have time to take a look at that?

Just so you know, this might get changed/deleted for the version `3.x` rework anyway - don't have a good view on that atm

connexion/apps/flask_app.py
thvu11: Unfortunately, this didn't work. `Gevent` module doesn't handle `self.app.debug` flag 😞

Ruwann: Alright, thanks for checking!

thvu11:
> Just so you know, this might get changed/deleted for the version `3.x` rework anyway - don't have a good view on that atm

No worries, I am exciting to see how things will be shaped up in version `3.x` and make some contributions to the future development. Even though this might get change, it's still a good idea to keep an eye on how everything interacts in the debug option.

PR #1558, opened 2022-06-22
Fix async tests

Ruwann:
The async tests were getting skipped because we didn't have an appropriate pytest setup for it. This PR aims to fix this and adjusts the tests accordingly.

### DIFF
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 from unittest.mock import MagicMock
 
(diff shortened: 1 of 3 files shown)


tests/decorators/test_security.py
```diff
@@ -63,16 +64,22 @@ async def get_tokeninfo_response(*args, **kwargs):
     client.get = get_tokeninfo_response
     monkeypatch.setattr(SecurityHandlerFactory, 'client', client)
 
-    with pytest.raises(OAuthScopeProblem, match="Provided token doesn't have the required scope"):
+    with pytest.raises(OAuthScopeProblem) as exc_info:
         await wrapped_func(request)
 
+    assert exc_info.value.status_code == 403
+    assert exc_info.value.detail == "Provided token doesn't have the required scope"
```
Ruwann, line 71: I believe this is because we are now subclassing connexion exceptions from starlette now, which have a different `str` representation

RobbeSneyders (approved):
Thanks @Ruwann. LGTM.

PR #1539, opened 2022-05-16
Add black formatting

RobbeSneyders:
Fixes #1498 

I've added the changes as 2 separate commits for easy review:
- Add the configuration changes to run black as part of pre-commit
- Run black formatting

If you have a big PR in progress I can rerun the formatting after merging @Ruwann.

PR #1559, opened 2022-06-24
Add test for MethodViewResolver

RobbeSneyders:
Follow up on #1552 

 - Add test for `MethodViewResolver`. I had to rename some stuff to clean this up a bit.
 - Add warning about ignoring `collection_endpoint_name` in `MethodViewResolver`
 - Add format of `class_arguments` in docstring

CC: @bluebrown

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -194,6 +194,11 @@ class MethodViewResolver(RestyResolver):
 
     def __init__(self, *args,  class_arguments=None, **kwargs):
         self.class_arguments = class_arguments or {}
+        if "collection_endpoint_name" in kwargs:
+            del kwargs["collection_endpoint_name"]
+            # Dispatch of request is done by Flask
+            logger.warning("collection_endpoint_name is ignored by the MethodViewResolver. "
+                           "Requests to a collection endpoint will be routed to .get()")
         super(MethodViewResolver, self).__init__(*args, **kwargs)
         self.initialized_views = []
 
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -6,6 +6,7 @@ from the operations defined in the OpenAPI spec.
 import inspect
 import logging
 import sys
+import typing as t
 
 from inflection import camelize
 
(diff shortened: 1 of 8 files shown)


bluebrown:
Perhaps the `DeprecatedMethodViewResolver` could be named `MethodResolver`. 

Because, it actually works without the class implementing the methods, inheriting from a `MethodView`. Since the resolver only instantiates the class via standard initializer.
 
The new `MethodViewResolver` on the other hand expects the `as_view` method to be present.

So both could be actually useful, in certain scenarios. 

Perhaps it could also make sense to pass the `class_arguments` to the class inside the old `MethodResolver`, it would be backwards compatible but add dependency injection capabilities. 

On that note, it might be worth documentation, that for the new `MethodViewResolver`, on each request a new instance of the view class is created, while for the old `MethodView`, the same instance is used across all requests. This might help people avoid pitfall regarding thread safety.

RobbeSneyders:
> Perhaps the `DeprecatedMethodViewResolver` could be named `MethodResolver`.
> 
> Because, it actually works without, the class implementing the methods, inheriting from a `MethodView`. Since the resolver only instantiates the class via standard initializer.
> 
> The new `MethodViewResolver` on the other hand expects the `as_view` method to be present.
> 
> So both could be actually useful, in certain scenarios.

Sounds good. 
Let's then split them up with a shared base class, so we can also throw the warning I added only for the MethodViewResolver.
 
> Perhaps it could also make sense to pass the `class_arguments` to the class inside the old `MethodResolver`, it would be backwards compatible but add dependency injection capabilities.

Good point.

> On that note, it might be worth documentation, that for the new `MethodViewResolver`, on each request a new instance of the view class is created, while for the old `MethodView`, the same instance is used across all requests. This might help people avoid pitfall regarding thread safety.

Flask just merged a flag to configure this behavior https://github.com/pallets/flask/pull/4624. Still unreleased, so would probably be good to refer to their documentation.

RobbeSneyders:
Would prefer to do it in a separate PR though. Feel free to take this up if you'd like.

bluebrown:
ok, should I wait until this one is merged then?

RobbeSneyders:
Have you reviewed this one? Let me know if it looks good to you and I'll merge.

bluebrown:
Yes, LGTM.

RobbeSneyders:
Just a heads up @bluebrown, we just merged black formatting (#1539).
So good to start from latest main.

PR #1560, opened 2022-06-24
Activate mypy check in pre-commit

### LINKED ISSUE #1111: Add type annotations and mypy validation

RobbeSneyders:
Fixes #1111

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,3 +41,15 @@ repos:
         name: black tests
         files: "^tests/"
         args: ["tests"]
+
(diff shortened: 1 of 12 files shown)


nielsbox (approved):
LGTM 👍🏻 there seem to be some merge conflicts though

PR #1561, opened 2022-06-27
Add .git-blame-ignore-revs file with black commit

RobbeSneyders:
Ignore the black formatting commit in git blame.
See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view.


Ruwann (approved):
Thanks!

PR #1564, opened 2022-07-03
Feature/method views args

### LINKED ISSUE #1563: allow MethodResolver and MethodViewResolver
It would be nice to have both:

- a `MethodResolver` that works with any class
- a `MethodViewResolver` that works specifically with flasks method views

In both cases, the class `args` and `kwargs` should be passed inside the resolver to allow for dependency injection.

Perhaps somewhere it should be documented that for the `MethodResolver` a single instance is used across all requests while with the `MethodViewResolver`, as the result of the as_view behaviour, a new instance is created for each request. This could help to avoid pitfall regarding thread safety.

bluebrown:
Fixes #1563 

 - rename DeprecatedMethodViewResolver to MethodResolver
 - use a base class for both MethodResolver and MethodViewResolver
 - pass class args / kwargs to the class instance/as_view function in both resolvers

### DIFF
--- a/connexion/resolver.py
+++ b/connexion/resolver.py
@@ -185,7 +185,7 @@ class RestyResolver(Resolver):
         return f"{get_controller_name()}.{get_function_name()}"
 
 
-class MethodViewResolver(RestyResolver):
+class MethodResolverBase(RestyResolver):
     """
     Resolves endpoint functions based on Flask's MethodView semantics, e.g. ::
 
@@ -225,7 +225,7 @@ class MethodViewResolver(RestyResolver):
                 "collection_endpoint_name is ignored by the MethodViewResolver. "
                 "Requests to a collection endpoint will be routed to .get()"
             )
-        super(MethodViewResolver, self).__init__(*args, **kwargs)
+        super(MethodResolverBase, self).__init__(*args, **kwargs)
         self.initialized_views = []
 
     def resolve_operation_id(self, operation):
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Thanks @bluebrown, looks good to me.

You do have an orphan commit because I rewrote history on main (to unsquash the black formatting so we can exclude it from history, https://github.com/spec-first/connexion/pull/1561). It's not an issue, but it will add me as a co-author to this PR. You can rebase on main to remove it.

bluebrown:
Hi @RobbeSneyders, what do you mean by rebase on main? 

Do you mean I should squash these 3 commits into 1, in my branch?

RobbeSneyders:
The first commit of your PR (9be473d) is outdated. The same changes are now added as separate commits on main. If you rebase on main, you should be able skip the first commit, only keeping the 2 relevant commits.

bluebrown:
Sorry, I don't know how to do that *rebase on main*. 

The only rebasing I know would be like below, on my feature branch.

```
git rebase -i HEAD~3
```

RobbeSneyders:
`git rebase main` should do it. Just make sure to pull `main` first so it's up to date.

bluebrown:
I have pulled the latest changes before branching.  Right now, I cannot fetch upsream because there are no new changes. 

image

Pulling my own main, locally also says no changes, and `git rebase main` in the feature branch as well.

```bash
connexion on  feature/method-views-args via 🐍 v3.9.2 on ☁️  (eu-central-1) ❯ git rebase main
Current branch feature/method-views-args is up to date
```

RobbeSneyders (approved):

PR #1568, opened 2022-07-21
Removed internal variable pass_context_arg_name

### LINKED ISSUE #1532: Remove `pass_context_arg_name` argument
I propose to remove the `pass_context_arg_name` argument and fix the name under which the context is passed. This parameter is passed often between internal functions and classes, which is not worth the benefit of having it configurable.

leonardofesta:
Fixes #1532 .

 - I removed the parameter _pass_context_arg_name_ passed internally as asked from the various constructors, and left in the _SecurityHandlerFactory_ a constant called **_context_kw_**, following the same pattern of _required_scopes_kw_. 

I don't know if it is better to put in uppercase both values (_context_kw and required_scopes_kw_), now they are constants in the end. 

I left as the other one to have consistency, but if you prefer I can change them both.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -62,9 +62,7 @@ def pythonic(name):
     return sanitized(name)
 
 
-def parameter_to_arg(
-    operation, function, pythonic_params=False, pass_context_arg_name=None
-):
+def parameter_to_arg(operation, function, pythonic_params=False):
     """
     Pass query and body parameters as keyword arguments to handler function.
 
@@ -74,9 +72,6 @@ def parameter_to_arg(
     :param pythonic_params: When True CamelCase parameters are converted to snake_case and an underscore is appended to
     any shadowed built-ins
     :type pythonic_params: bool
-    :param pass_context_arg_name: If not None URL and function has an argument matching this name, the framework's
-    request context will be passed as that argument.
-    :type pass_context_arg_name: str|None
     """
     consumes = operation.consumes
 
@@ -123,10 +118,6 @@ def parameter_to_arg(
                 kwargs[key] = value
             else:
                 logger.debug("Context parameter '%s' not in function arguments", key)
-
-        # attempt to provide the request context to the function
-        if pass_context_arg_name and (has_kwargs or pass_context_arg_name in arguments):
-            kwargs[pass_context_arg_name] = request.context
 
         return function(**kwargs)
 
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -62,7 +62,9 @@ def pythonic(name):
     return sanitized(name)
 
 
-def parameter_to_arg(operation, function, pythonic_params=False):
+def parameter_to_arg(
+    operation, function, pythonic_params=False, pass_context_arg_name=None
+):
     """
     Pass query and body parameters as keyword arguments to handler function.
 
@@ -118,6 +120,9 @@ def parameter_to_arg(operation, function, pythonic_params=False):
                 kwargs[key] = value
             else:
                 logger.debug("Context parameter '%s' not in function arguments", key)
+        # attempt to provide the request context to the function
+        if pass_context_arg_name and (has_kwargs or pass_context_arg_name in arguments):
+            kwargs[pass_context_arg_name] = request.context
 
         return function(**kwargs)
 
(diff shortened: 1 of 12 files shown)


RobbeSneyders:
Thanks @leonardofesta!
Could you update the failing tests to make them pass?

RobbeSneyders:
Hi @leonardofesta, did you have a chance to have another look at this? Let me know if you need support.

leonardofesta:
Sorry @RobbeSneyders I was on vacation and I didn't saw the results. Now it should pass all tests.

nielsbox (approved):
LGTM apart from the thing that Ruwann said about collisions, could you maybe add a test where there is a parameter called `context`? Also converting the constants to all uppercase seems like a refactor we should do in a separate PR. Let's see what Ruwann and Robbe say.

connexion/decorators/parameter.py
```diff
@@ -124,10 +119,6 @@ def wrapper(request):
             else:
                 logger.debug("Context parameter '%s' not in function arguments", key)
 
-        # attempt to provide the request context to the function
```
RobbeSneyders, line 127: We still want to pass the context as an argument to the view function, so we can't just remove this. We need to include the context in `kwargs` using `context_kw`.

tests/decorators/test_parameter.py
```diff
@@ -23,11 +23,6 @@ def get_arguments(self, *args, **kwargs):
     parameter_to_arg(Op(), handler)(request)
     func.assert_called_with(p1="123")
 
-    parameter_to_arg(Op(), handler, pass_context_arg_name="framework_request_ctx")(
```
RobbeSneyders, line 26: We want to keep this behaviour, so should keep testing for it

tests/fakeapi/hello/__init__.py
```diff
@@ -87,8 +87,8 @@ def get_bye_secure_from_flask():
     return "Goodbye {user} (Secure!)".format(user=context["user"])
 
 
-def get_bye_secure_from_connexion(req_context):
-    return "Goodbye {user} (Secure!)".format(user=req_context["user"])
+def get_bye_secure_from_connexion():
+    return "Goodbye {user} (Secure!)".format(user=context["user"])
```
RobbeSneyders, line 91: We want to check that the context is passed to the function, so we shouldn't use the global context here.

RobbeSneyders (changes requested):
Thanks @LeenaBhegade

You've removed some functionality that we need to keep. I added some comments to indicate where.

This means that we need the `context_kw` both in the parameter decorator and the security handler factory. We can probably just define it on both separately, as they are quite decoupled from each other.

I would also propose to use `context_` as the default `context_kw`. The underscore lowers the chance of collision and indicates that it's a special variable.

RobbeSneyders:
Any update @leonardofesta?

leonardofesta:
> Any update @leonardofesta?
I just pushed the update.
I'm using context_ as suggested as value, in both cases used as constant.

- I basically restored the decorator functionality by using a boolean with default == False to maintain the same logic
- In the security handler part, please recheck  but now it should be correct

connexion/security/security_handler_factory.py
```diff
@@ -335,9 +333,7 @@ def wrapper(request):
 
     def _need_to_add_context_or_scopes(self, func):
         arguments, has_kwargs = inspect_function_arguments(func)
-        need_context = self.pass_context_arg_name and (
-            has_kwargs or self.pass_context_arg_name in arguments
-        )
+        need_context = has_kwargs or self.context_kw in arguments
```
RobbeSneyders, line 336: We could do this by removing the `has_kwargs` here.

RobbeSneyders (changes requested):
> * I basically restored the decorator functionality by using a boolean with default == False to maintain the same logic

The reason to remove the `pass_context_arg_name` variable, was that it needs to be passed around a lot for a small benefit. Replacing it with a boolean doesn't really solve this.

Instead it would be nice if the context is passed automatically if the `context_` parameter is added to the view function.

connexion/security/security_handler_factory.py
leonardofesta: Ok i'll fix the two things as suggested

RobbeSneyders:
Thanks @leonardofesta, looks good to me.

I pushed a change to the test, as it wasn't actually testing the new behavior.

RobbeSneyders (approved):

PR #1570, opened 2022-08-02
fix: don't interpret simple parameter as deepObject

### LINKED ISSUE #1565: Parameters are treated as deep object even if not configured
### Description

We define array parameters which names contain brackets, like this:

```yaml
    SurnameFilterInQuery:
      name: filter[surname]
      description: ...
      in: query
      required: false
      explode: false
      schema:
        type: array
        items:
          $ref: "#/components/schemas/Surname"

    # ...

    Surname:
      type: string
      description: ...
```

Note that we do not define `deepObject` style and simply want to pass comma-separated array items.

Now we perform slightly malformed queries to our service, like this:

```shell
curl -H "https://api.example.com/?filter[surname]_bogus=greinacher" # url-decoded for readability
``` 

### Actual behaviour

Connexion crashes with this message:

```
ERROR:asyncio:'dict' object has no attribute 'split'
```

Which is caused by this line: https://github.com/spec-first/connexion/blob/v2/connexion/decorators/uri_parsing.py#L259.

### Expected behaviour

Connexion does not crash but fail because the parameter `filter[surname]_bogus` is not declared.

### Steps to reproduce

See above

### Additional info:

Output of the commands:

- `poetry run python --version`: `Python 3.9.13`
- `poetry show connexion | grep "^version" `: `version      : 2.14.0`

fgreinacher:
Fixes #1565

### DIFF
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -201,6 +201,9 @@ class OpenAPIURIParser(AbstractURIParser):
             if k == root_key:
                 return k, v, False
 
+        if not self._is_deep_object_style_param(root_key):
+            return k, v, False
+
         key_path = re.findall(r'\[([^\[\]]*)\]', k)
         root = prev = node = {}
         for k in key_path:
@@ -209,6 +212,11 @@ class OpenAPIURIParser(AbstractURIParser):
             node = node[k]
         prev[k] = v[0]
         return root_key, [root], True
+    
+    def _is_deep_object_style_param(self, param_name):
+        default_style = self.style_defaults["query"] 
+        style = self.param_defns.get(param_name, {}).get('style', default_style)
+        return style == "deepObject"
 
     def _preprocess_deep_objects(self, query_data):
         """ deep objects provide a way of rendering nested objects using query
--- a/connexion/decorators/uri_parsing.py
+++ b/connexion/decorators/uri_parsing.py
@@ -212,9 +212,9 @@ class OpenAPIURIParser(AbstractURIParser):
             node = node[k]
         prev[k] = v[0]
         return root_key, [root], True
-    
+
     def _is_deep_object_style_param(self, param_name):
-        default_style = self.style_defaults["query"] 
+        default_style = self.style_defaults["query"]
         style = self.param_defns.get(param_name, {}).get('style', default_style)
         return style == "deepObject"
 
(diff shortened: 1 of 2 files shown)


fgreinacher:
@RobbeSneyders Here is my take on a fix. It works fine in our application:

```shell
$ http :8080/api "filter[surname]_bogus"=="greinacher"
HTTP/1.1 400 Bad Request
Content-Length: 235
Content-Type: application/vnd.api+json; charset=utf-8
Date: Tue, 02 Aug 2022 14:56:31 GMT

{
    "errors": [
        {
            "code": "xrn:err:core:invalidRequestParameter",
            "detail": "Extra query parameter(s) filter[surname]_bogus not in spec",
            "id": "a6a0b17e-e52f-37e4-a7d9-6c2e926ed4f8",
            "status": "400",
            "title": "Invalid Request Parameter"
        }
    ]
}
```

connexion/decorators/uri_parsing.py
```diff
@@ -201,6 +201,9 @@ def _make_deep_object(self, k, v):
             if k == root_key:
                 return k, v, False
 
+        if not self._is_deep_object_style_param(root_key):
+            return k, v, False
+
```
fgreinacher, line 206: This is the first place where we know the root parameter name and have the schema information available. We already have a lot of special casing here, so that follow-up code can assume everything is fine. Let me know if you prefer to move that checker somewhere else.

RobbeSneyders:
Thanks @fgreinacher!
I've submitted a PR to fix the failing tests due to a new Flask version: https://github.com/spec-first/connexion/pull/1572. I'd like to merge that one first. Feel free to review it.

fgreinacher:
> Thanks @fgreinacher! I've submitted a PR to fix the failing tests due to a new Flask version: #1572. I'd like to merge that one first. Feel free to review it.

Thanks @RobbeSneyders, I was wondering how this failure could be related to my change :)

The other MR looks good to me!

fgreinacher:
@RobbeSneyders I'm planning the last days before my summer vacation. Do you think we could get this in early next week so I can verify it with our services? 🙇

RobbeSneyders:
Sorry for the delay, I was on vacation myself :)
Should be good to go now. I'll prep a release.

RobbeSneyders (approved):

D3mong13 (approved):

PR #1571, opened 2022-08-06
Allow enums with nullable defaults

hauntsaninja:
The following two PRs added better support for nullable defaults:
https://github.com/spec-first/connexion/pull/1463
https://github.com/spec-first/connexion/pull/1464

However, it looks like it missed adding `NullableEnumValidator` to `create_spec_validator`.

Also see https://github.com/OAI/OpenAPI-Specification/issues/1900 , which establishes that you can't just put `null` into the enum.

I'd never used openapi or jsonschema or whatever before this week, so maybe I'm totally misunderstanding something. Apologies if so.

### DIFF
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -17,7 +17,7 @@ from jsonschema import Draft4Validator
 from jsonschema.validators import extend as extend_validator
 
 from .exceptions import InvalidSpecification
-from .json_schema import NullableTypeValidator, resolve_refs
+from .json_schema import NullableEnumValidator, NullableTypeValidator, resolve_refs
 from .operations import OpenAPIOperation, Swagger2Operation
 from .utils import deep_get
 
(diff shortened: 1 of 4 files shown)


hauntsaninja:
cc @Ruwann  :-)

hauntsaninja:
Thanks RobbeSneyders for triggering CI on this :-)

If this is a valid fix, I'm also curious about whether a release could be made... I wasn't able to find any docs about connexion's release schedule. Thank you again!

RobbeSneyders:
Hi @hauntsaninja, thanks for the PR. 

It seems like there was a [clarification](https://github.com/tedepstein/OpenAPI-Specification/blob/325db7f50610aed0af967d5907f9e5f6a32168a1/proposals/2019-10-31-Clarify-Nullable.md#if-a-schema-specifies-nullable-true-and-enum-1-2-3-does-that-schema-allow-null-values-see-1900) to the OpenAPI spec last year that specifies that this behavior is not allowed though:
> **If a schema specifies nullable: true and enum: [1, 2, 3], does that schema allow null values? (See https://github.com/OAI/OpenAPI-Specification/issues/1900.)**  
>No. The nullable: true assertion folds into the type assertion, which presumably specifies integer or number.
> While the modified type now allows null, the enum does not. Consistent with JSON schema, a value conforms to the schema only if it is valid against all constraints. Any constraint, in this case enum, can cause a value to fail validation, even if that value meets all of the other constraints.

We should probably remove the NullableEnumValidator instead of expanding its usage.

hauntsaninja:
Okie dokie, seems clear enough! Thank you

PR #1582, opened 2022-09-02
Update json for Flask 2.3

### LINKED ISSUE #1576: Pytest generates 2,464 warnings that should be investigated
### Description

Our GitHub Actions run pytest which generates a lot of warnings that should be investigated and remedied.

From the GitHub Action output:
> ================ 467 passed, 2464 warnings in 111.63s (0:01:51) ================

### Additional info:

The output of the commands:

- `python --version`

RobbeSneyders:
Partially fixes #1576 by updating the jsonifier and related code to be compatible with Flask 2.3.
Since this new json interface in Flask was only introduced in Flask 2.2, we'll have to pin it as the lower bound version.

I'm doubting if we should backport this to v2. It introduces some breaking changes for users who use custom json encoders, and it would force us to bump the lower bound version of Flask to 2.2 there as well.

If we don't do a backport, we should pin the upper bound Flask version on v2 to 2.2 instead.

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -2,16 +2,16 @@
 This module defines a FlaskApp, a Connexion application to wrap a Flask application.
 """
 
-import datetime
 import logging
 import pathlib
-from decimal import Decimal
 from types import FunctionType  # NOQA
 
 import a2wsgi
 import flask
 import werkzeug.exceptions
-from flask import json, signals
+from flask import signals
+
+from connexion import jsonifier
 
 from ..apis.flask_api import FlaskApi
 from ..exceptions import ProblemException
@@ -36,7 +36,7 @@ class FlaskApp(AbstractApp):
 
     def create_app(self):
         app = flask.Flask(self.import_name, **self.server_args)
-        app.json_encoder = FlaskJSONEncoder
+        app.json = FlaskJSONProvider(app)
         app.url_map.converters["float"] = NumberConverter
         app.url_map.converters["int"] = IntegerConverter
         return app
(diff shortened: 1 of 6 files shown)


nielsbox (approved):
LGTM 
```we should pin the upper bound Flask version on v2 to 2.2 instead.```
I agree with not backporting and pinning the version instead. Seems like an unnecessary breaking change we do not wish to introduce in a minor version increase from a Connexion perspective.

On a side note: Are there features from 2.2> that some users cannot wait for? Maybe an unofficial release could be the answer for that?

RobbeSneyders:
> On a side note: Are there features from 2.2> that some users cannot wait for? 

2.2 is currently the latest version, so hard to say. Biggest risk is that it will still take a while before we can release 3.X, and we'll be lagging multiple versions behind Flask. Of course we can still choose to backport this later on.

> Maybe an unofficial release could be the answer for that?

What do you mean with an unofficial release? I wouldn't like to branch our releases even further. We now have 2 branches to maintain (3.X and 2.X) which already requires some additional effort. I wouldn't want to have to support a 3rd :)

nielsbox:
> > On a side note: Are there features from 2.2> that some users cannot wait for?
> 
> 2.2 is currently the latest version, so hard to say. Biggest risk is that it will still take a while before we can release 3.X, and we'll be lagging multiple versions behind Flask. Of course we can still choose to backport this later on.

Exactly my thoughts. Since we are currently developing 3.0 we cannot issue a new breaking release (we can from a separate branch, but I don't think it is worth it IMO like you said)

> > Maybe an unofficial release could be the answer for that?
> 
> What do you mean with an unofficial release? I wouldn't like to branch our releases even further. We now have 2 branches to maintain (3.X and 2.X) which already requires some additional effort. I wouldn't want to have to support a 3rd :)
Yeah, I meant exactly that but looking back on my suggestion it seems a bit too much, users will be expecting support and maintenance. So let's just pin it and keep it at that. We can always backport this later on, indeed.

RobbeSneyders:
Ok, I'll submit a PR to the v2 branch to limit the upper version of Flask.

PR #1584, opened 2022-09-04
Resolve $ref referring to another $ref

### LINKED ISSUE #1581: Nested reference cannot be resolved
### Description

I am not too sure if this is expected behavior or not but simple nested reference cannot be resolved.

```python
>>> from connexion.json_schema import resolve_refs
>>> spec = {
...     "A": {
...         "$ref": "#/B",
...     },
...     "B": {
...         "$ref": "#/C",
...     },
...     "C": {
...         "type": "string",
...     },
... }
>>> resolve_refs(spec)
{'A': {}, 'B': {'type': 'string'}, 'C': {'type': 'string'}}
```

I expect the following result.

```python
{'A': {'type': 'string'}, 'B': {'type': 'string'}, 'C': {'type': 'string'}}
```

### Expected behaviour

Nested reference can be resolved or an exception should be raised if nested reference is not supported.

```python
>>> from connexion.json_schema import resolve_refs
>>> spec = {
...     "A": {
...         "$ref": "#/B",
...     },
...     "B": {
...         "$ref": "#/C",
...     },
...     "C": {
...         "type": "string",
...     },
... }
>>> resolve_refs(spec)
{'A': {'type': 'string'}, 'B': {'type': 'string'}, 'C': {'type': 'string'}}
```

### Actual behaviour

See the description above. [connexion/json_schema.py#L86-L90](https://github.com/spec-first/connexion/blob/2.14.1/connexion/json_schema.py#L86-L90) might be the root cause of this behavior. This may be changed as follows:

```python
            try:
                # resolve known references
                retrieved = deep_get(spec, path)
                node.update(retrieved)
                if isinstance(retrieved, Mapping) and '$ref' in retrieved:
                    node = _do_resolve(node)
                node.pop('$ref', None)
                return node
```

### Steps to reproduce

See the description above.

### Additional info:

Output of the commands:

- `python --version`: `Python 3.9.13`
- `pip show connexion | grep "^Version\:"`: `Version: 2.14.1`

chibacchie:
Fixes #1581.

 - Support to resolve `$ref` referring to another `$ref`.

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -85,8 +85,11 @@ def resolve_refs(spec, store=None, handlers=None):
             path = node["$ref"][2:].split("/")
             try:
                 # resolve known references
-                node.update(deep_get(spec, path))
-                del node["$ref"]
+                retrieved = deep_get(spec, path)
+                node.update(retrieved)
+                if isinstance(retrieved, Mapping) and '$ref' in retrieved:
+                    node = _do_resolve(node)
+                node.pop('$ref', None)
                 return node
             except KeyError:
                 # resolve external references
--- a/tests/test_references.py
+++ b/tests/test_references.py
@@ -97,3 +97,24 @@ def test_resolve_web_reference(api):
 
     spec = resolve_refs(op_spec, store=store)
     assert spec["parameters"][0]["name"] == "test"
+
+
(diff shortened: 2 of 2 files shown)


RobbeSneyders (approved):
Thanks @chibacchie, LGTM.

PR #1585, opened 2022-09-07
Fix Flask upper limit to 2.3

RobbeSneyders:
Follow up on https://github.com/spec-first/connexion/pull/1582

This PR fixes the upper limit version for Flask and Werkzeug to 2.3, since supporting 2.3 would require us to make breaking changes. We should release this as a new minor version.


Ruwann (approved):
👍

Issue #1525, opened 2022-05-06
Extract validation into middleware


RobbeSneyders:
This issue is part of #1489 and has the goal of moving the validation into the middleware stack.

Validation contains:
- Request body validation
- Request parameter validation (path & query)
- Response body validation

This depends on #1524

PR #1588, opened 2022-09-13
Extract JSON request body validation to middleware

RobbeSneyders:
First step towards #1525 

This PR extracts the request body validation for json payloads into middleware. The body validator is pluggable by content type and implements the ASGI interface. This way the stream can be consumed and replayed as needed for each content type.

A couple of notes / high level ideas
 - We should transform the existing `Operation` classes into `OperationSpec` classes which provides an interface for an operation specification across Swagger 2 & OpenAPI 3. I don't think we should subclass the middleware operation classes for Swagger 2 & OpenAPI separately.
 - I think middleware operation classes should implement the ASGI interface and wrap the next app. This is already the case in the routing middleware, but not the security middleware. I'll fix this in a separate PR.
 - We have a lot of boilerplate across the middlewares. We should check if we can abstract this into some base classes.


connexion/middleware/validation.py
```diff
@@ -0,0 +1,232 @@
...
+from connexion.validators import JSONBodyValidator
+
+from ..decorators.response import ResponseValidator
+from ..decorators.validation import ParameterValidator
+
+logger = logging.getLogger("connexion.middleware.validation")
+
+VALIDATOR_MAP = {
+    "parameter": ParameterValidator,
+    "body": {"application/json": JSONBodyValidator},
```
Ruwann, line 28: I have been struggling here with the distinction of `parameter` and `body` as there currently isn't a clear split: `form` parameters are validated by the `ParameterValidator` but also by the `RequestBodyValidator`. The request body also contains parameters, at least in Swagger 2 ([link](https://swagger.io/docs/specification/2-0/describing-parameters/#types)). In OpenAPI there is a clearer distinction between "parameters" and the "body" ([link](https://swagger.io/docs/specification/describing-parameters/#types)).

Seeing this, I think a way forward could be to move everything of formdata to a separate body validator as well. This also simplifies things wrt accessing the ASGI scope only vs also needing the `send` and `receive` callables. Wdyt?

RobbeSneyders: Yes, I think that makes sense. Just let me rephrase to make sure I understand:

- Move form validation into a separate body validator so it has the full ASGI interface (scope, send, receive) available and can decide if / how to consume the stream
- Parameter validators will only have the scope available, possibly via a Request object which does not provide access to the stream.

Correct?

tests/fixtures/simple/openapi.yaml
```diff
@@ -980,7 +980,7 @@ paths:
                 type: object
       requestBody:
         content:
-          multipart/form-data:
+          application/x-www-form-urlencoded:
```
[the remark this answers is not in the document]
RobbeSneyders: The [test](https://github.com/spec-first/connexion/blob/main/tests/api/test_parameters.py#L456) that uses this endpoint uses the Flask test client, which automatically adds the `application/x-www-form-urlencoded` content type header when passing data ([docs](https://werkzeug.palletsprojects.com/en/2.2.x/test/#request-body))

tests/api/test_responses.py
```diff
@@ -368,31 +372,6 @@ def test_post_wrong_content_type(simple_app):
     )
     assert resp.status_code == 415
 
-    # this test checks exactly what the test directly above is supposed to check,
```
RobbeSneyders, line 371: Test is outdated as explained in comment.

tests/api/test_responses.py
```diff
@@ -356,10 +356,14 @@ def test_post_wrong_content_type(simple_app):
     )
     assert resp.status_code == 415
 
-    resp = app_client.post(
-        "/v1.0/post_wrong_content_type", data=json.dumps({"some": "data"})
-    )
-    assert resp.status_code == 415
+    # According to the spec, a content-type header is not required
```
RobbeSneyders, line 359: Can you validate my thinking here @Ruwann?

nielsbox: According to [this](https://portswigger.net/kb/issues/00800500_content-type-is-not-specified) Probably the only way to do this, and to follow the RFC spec, is to either try to determine the bytes using something like the `magic` library or to try to parse the bytes using all available payload parsers. If either fails we should send a 415.

RobbeSneyders: We're actually doing some [best guessing](https://github.com/spec-first/connexion/pull/1588/files#diff-0f386dec8fb7dd9135bbf5c0a680ab61c9c9a9e07988be627c43b0aaf0e6692dR189) now, which is why we're no longer returning a 415 error here.

This test previously ensured that a request without a `Content-Type` header would return a 415. But as you mention as well, a `Content-Type` header is not required. So I think this test was incorrect.

Ruwann: I agree, I don't think that just the absence of a `Content-Type` header warrants a 415 response. So the test is incorrect as far as I'm concerned.

However, to nitpick, if the endpoint specifies a certain `Content-Type` in the API contract, and if the `Content-Type` header is missing and the payload cannot be interpreted as that content type, it is still *valid* to return a 415. For example, [here](https://github.com/spec-first/connexion/pull/1588/files#diff-f3b50008022f534c8f329ff026c99b5bf9d12b27da6aaeae91ba2d32583f7c37R74) it would be possible to return a 415 in case of failure?

connexion/middleware/validation.py
Ruwann: Yes, indeed. This to limit the amount of places in which the body stream could be accessed.

tests/api/test_responses.py
Ruwann: Yeah, true. Didn't consider that a malformed payload is also an option.

Agreed, it was a bit of a sidetrack here as the original question/comment on the test is clear now.

Ruwann:
- Agreed that there should be a `Operation`-like class that only serves to provide access to the Operation Spec. It would be nice to have a single interface for swagger 2 and openapi 3, but don't have a good view on whether that's actually feasible (some changes from 2 to 3 might be too big to unify again)
- I *think* it makes sense indeed? At first sight, this would make it easier to reason about the `send` and `receive` callables, especially for the body validators. But I'll have to try it out myself to fully understand the impact.

RobbeSneyders:
@Ruwann @nielsbox The PR is cleaned up and all tests are passing, so if you approve, I'd propose to merge this already.

nielsbox (approved):

connexion/middleware/validation.py
LewisCowlesMotive: 👋  does this code mean that the `JSONBodyValidator` is string-literally matched? I ask because I was thinking about a case where an American business is insisting that a partner use their `mm/dd/yyyy` format for dates. They already use Connexion, and I'd sooner specify a separate accept & content-type for their weird expected output, and have everything else use, nice readable, standard-compliant ISO-8601 as per https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 and https://swagger.io/docs/specification/data-models/data-types/#string.

PR #1589, opened 2022-09-15
Remove AbstractSwaggerUIAPI class

RobbeSneyders:
This PR removes the `AbstractSwaggerUIAPI` class, which was a temporary base class used to support the swagger UI both in the Flask App & Middleware. Since the swagger UI has been moved completely into the middleware, we don't need it anymore.


Ruwann (approved):
👍

PR #1590, opened 2022-09-16
Extract boilerplate code into Routed base classes

RobbeSneyders:
As mentioned in #1588, we're repeating a lot of boilerplate across the middlewares. This PR extracts this boilerplate into shared base classes. Feel free to propose better naming for them :slightly_smiling_face: 

I had to revert the order of Middleware - Api - Operation to Operation - Api - Middleware due to dependencies in the typing.

### DIFF
--- a/connexion/middleware/abstract.py
+++ b/connexion/middleware/abstract.py
@@ -93,11 +93,8 @@ class RoutedMiddleware(AppMiddleware, t.Generic[API]):
     operation_id lookup.
     """
 
-    @property
-    @abc.abstractmethod
-    def api_cls(self) -> t.Type[API]:
-        """The subclass of RoutedAPI this middleware uses."""
-        raise NotImplementedError
+    api_cls: t.Type[API]
+    """The subclass of RoutedAPI this middleware uses."""
 
     def __init__(self, app: ASGIApp) -> None:
         self.app = app
--- a/connexion/middleware/abstract.py
+++ b/connexion/middleware/abstract.py
@@ -3,6 +3,7 @@ import logging
 import pathlib
 import typing as t
 
+import typing_extensions as te
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from connexion.apis.abstract import AbstractSpecAPI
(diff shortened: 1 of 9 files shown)


connexion/middleware/routing.py
```diff
@@ -6,61 +6,36 @@
 from starlette.types import ASGIApp, Receive, Scope, Send
```
RobbeSneyders, line 6: Only changed the order in this file.

connexion/middleware/swagger_ui.py
```diff
@@ -21,56 +21,6 @@
 _original_scope: ContextVar[Scope] = ContextVar("SCOPE")
```
RobbeSneyders, line 21: Only changed the order in this file.

connexion/middleware/routing.py
RobbeSneyders, line 6: Only changed the order in this file and moved the `ROUTING_CONTEXT`.

connexion/middleware/abstract.py
```diff
@@ -12,3 +26,120 @@ def add_api(
...
+    its operations.
+
+    The RoutingMiddleware adds the operation_id to the ASGI scope. This middleware registers its
+    operations by operation_id at startup. At request time, the operation is fetched by an
+    operation_id lookup.
+    """
+
+    @property
+    @abc.abstractmethod
+    def api_cls(self) -> t.Type[API]:
```
Ruwann, line 98: Possible to do similar as in RouteAPI to avoid property?

connexion/middleware/security.py
```diff
@@ -304,8 +198,47 @@ def _get_verification_fn(self):
...
+            next_app=self.next_app,
+            security_handler_factory=self.security_handler_factory,
+        )
+
+
+class SecurityMiddleware(RoutedMiddleware[SecurityAPI]):
+    """Middleware to check if operation is accessible on scope."""
+
+    @property
+    def api_cls(self) -> t.Type[SecurityAPI]:
```
Ruwann, line 240: see base class

Ruwann (approved):

PR #1591, opened 2022-09-18
Move JSON response body validation to middleware

RobbeSneyders:
This PR moves the JSON response body validation to middleware as another step towards #1525.

I moved away from having the body validators implement the complete ASGI interface and calling the next app as mentioned in #1588, as this is not possible for the response validators. We only know the response content type once the response has started, so we can't wrap the validator around the next app.

Instead, the validator now provides a wrapped receive / send callable depending on if it validates either request or response bodies. I updated the request body validator to match this.


tests/api/test_headers.py
```diff
@@ -34,7 +34,7 @@ def test_header_not_returned(simple_openapi_app):
     assert data["title"] == "Response headers do not conform to specification"
     assert (
         data["detail"]
-        == "Keys in header don't match response specification. Difference: Location"
+        == "Keys in header don't match response specification. Difference: location"
```
RobbeSneyders, line 37: Headers are case insensitive. We cast them to lowercase to check the diff here.

tests/api/test_schema.py
```diff
@@ -80,59 +80,59 @@ def test_schema_response(schema_app):
     request = app_client.get(
         "/v1.0/test_schema/response/object/valid", headers={}, data=None
     )  # type: flask.Response
-    assert request.status_code == 200
+    assert request.status_code == 200, request.text
```
RobbeSneyders, line 83: These tests were failing during development and don't provide a lot of insight into what causes the error. Logging the request text if the assert fails makes it easier to debug.

tests/fakeapi/example_method_view.py
```diff
@@ -6,19 +6,22 @@ class PetsView(MethodView):
     mycontent = "demonstrate return from MethodView class"
 
     def get(self, **kwargs):
-        kwargs.update({"method": "get"})
```
RobbeSneyders, line 9: These responses were not conforming to the spec, but were not checked correctly before.

tests/test_resolver_methodview.py
```diff
@@ -192,13 +192,13 @@ def test_method_view_resolver_integration(method_view_app):
     client = method_view_app.app.test_client()
 
     r = client.get("/v1.0/pets")
-    assert r.json == {"method": "get"}
```
RobbeSneyders, line 195: Reflect changes in the fake api above.

nielsbox (approved):

connexion/validators.py
```diff
@@ -77,11 +75,92 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+
+        while self._messages:
+            await self._send(self._messages.pop(0))
+
+
+class TextResponseBodyValidator(JSONResponseBodyValidator):
+    @staticmethod
+    def parse(body: str) -> str:  # type: ignore
+        try:
+            return json.loads(body)
```
Ruwann, line 154: TBD: Is this still necessary or can we change this behaviour in the new major version?

RobbeSneyders: See my comment above

connexion/validators.py
```diff
@@ -77,11 +75,92 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+        except json.decoder.JSONDecodeError:
+            return body
+
 
-        await self.next_app(scope, wrapped_receive, send)
+VALIDATOR_MAP = {
+    "parameter": ParameterValidator,
+    "body": {"application/json": JSONRequestBodyValidator},
+    "response": {
```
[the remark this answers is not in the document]
RobbeSneyders: Good catch. Question is what kind of behavior we want here. I do think we want to update at the top level, as it would otherwise be impossible to remove validators. That just means the user needs to include all validators they want to use in their custom validator_map.

PR #1595, opened 2022-09-24
Add form data validator for validation middleware

RobbeSneyders:
This PR moves form data validation to the middleware, which is quite complex due to differences between Swagger 2 and OpenAPI 3 in how form data is represented.

For OpenAPI 3, this is quite simple. The form data is represented as an object in the request body, which means we can just use a request body validator.

For Swagger 2, this is more complicated. The form data is represented as parameters instead. I see two options here:
- We validate the form data as parameters, which is what Connexion is currently doing. The advantage is that you can easily use the current parameter validation with the spec. The downside is that it's different from OpenAPI 3 and you need to consume the stream in the parameter validation.
- We transform the FormData parameters in the Swagger2Operation to a body schema spec as in OpenAPI 3. Advantage is that we can then use the same request body validator and only consume the stream there. The downside is additional complexity in the Swagger2Operation.

This PR implements option 2.

Some open points still to address with the request body validator in this PR:
- I implemented `strict_validation`, which makes less sense for OpenAPI 3 form validation, as this can now be controlled by `additionalProperties` in the schema.instead.

### DIFF
--- a/connexion/validators.py
+++ b/connexion/validators.py
@@ -6,10 +6,20 @@ import logging
 import typing as t
 
 from jsonschema import Draft4Validator, ValidationError, draft4_format_checker
+from starlette.datastructures import FormData, Headers
+from starlette.formparsers import FormParser, MultiPartParser
 from starlette.types import Receive, Scope, Send
 
-from connexion.decorators.validation import ParameterValidator
-from connexion.exceptions import BadRequestProblem, NonConformingResponseBody
+from connexion.decorators.validation import (
+    ParameterValidator,
+    TypeValidationError,
+    coerce_type,
+)
+from connexion.exceptions import (
+    BadRequestProblem,
+    ExtraParameterProblem,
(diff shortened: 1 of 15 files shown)


connexion/validators.py
```diff
@@ -56,7 +67,7 @@ def validate(self, body: dict):
             )
             raise BadRequestProblem(detail=f"{exception.message}{error_path_msg}")
 
-    async def receive(self) -> t.Optional[t.MutableMapping[str, t.Any]]:
+    async def wrapped_receive(self) -> Receive:
```
RobbeSneyders, line 70: I changed this so validation is not delayed until the stream is consumed by the application. I think it makes sense for connexion to always validate the body (for certain content types), independent on if it's read in the application or not. This also makes it easier to follow / debug validation, and prevents further middlewares / application logic from being executed for an invalid request.

PR #1596, opened 2022-09-27
Add form parsing via URIParser

Ruwann:
Quick and dirty solution to combine starlette form parsing with the URIParser (as we have on v2 with the werkzeug one)


RobbeSneyders:
Nice, seems like I misassumed how the uri parsing works.

PR #1598, opened 2022-09-27
Feature/multiple yaml same base path

leonardofesta:
Fixes https://github.com/spec-first/connexion/issues/1542 .

It is possible to pass a name parameter to the connexion.app_api(..) method
the name parameter will be assigned to the Flask blueprint
in case you don't pass the name, it will rely on previous/default behaviour (base path as name in Flask case)
Having multiple specification in the same path, opens to new conflicting case.

In case the same endpoint is defined from 2 specs, the first one added and covering the case will be accepted, the second one ignored.
A more niche case, in case you define 2 operation on the same path but different method, the second one won't work, even if they are complementary.

Eg:
spec1 get foo/
spec2 post foo/
if you call post foo/ --> response will be 405, method not allowed (it ignores 2nd definition of the same path)

I think these cases can be fine as long they are documented (let me know where do you think I should amend the documentation).

Let me know if you want more unit tests

Ps: sorry for the mess on the other pr, somehow a work account. ended up as commentor

### DIFF
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -4,8 +4,7 @@ This module defines a FlaskApp, a Connexion application to wrap a Flask applicat
 
 import logging
 import pathlib
-from types import FunctionType
-from typing import Optional  # NOQA
+from types import FunctionType  # NOQA
 
 import a2wsgi
 import flask
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -4,7 +4,8 @@ This module defines a FlaskApp, a Connexion application to wrap a Flask applicat
 
 import logging
 import pathlib
-from types import FunctionType  # NOQA
+from types import FunctionType
+from typing import Optional  # NOQA
 
 import a2wsgi
 import flask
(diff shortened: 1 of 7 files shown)


RobbeSneyders:
Thanks @leonardofesta!

> In case the same endpoint is defined from 2 specs, the first one added and covering the case will be accepted, the second one ignored.
> A more niche case, in case you define 2 operation on the same path but different method, the second one won't work, even if they are complementary.

I don't really like this behavior as it is not the behavior I would expect. I think ideally we would:
- Raise an error if multiple operations are registered for the same path and method
- Allow different operations on the same path as long as they are complementary

I think this behavior is generated by how operations are registered in our RoutingMiddleware (see #1489 for background):
https://github.com/spec-first/connexion/blob/181c61bfb68eb13afe7707b362da9e2245448ef3/connexion/middleware/routing.py#L77-L80

We can probably add logic here to check if an operation is already registered for the same path and methods, and either add a new route, update the existing one, or raise an error.

RobbeSneyders:
Updated my comment above after having a second look.

leonardofesta:
Yes, it doesn't seem straightforward.

As far as I understood for each _add_api()_ we create a blueprint and a starlette router. 
This make difficult to check issues between the 2 yaml configurations. 

Also as far as I see from the starlette router, from their documentation [on routing](https://www.starlette.io/routing/) they specify they use a policy of priority on routes, in case of conflicts first win. 

The second issue where you get 405 for same path and different method is caused also from starlette implementation for this piece of code [here](https://github.com/encode/starlette/blob/fa974b6a0c6f1199357f946080dde9b96ef5ec02/starlette/routing.py#L265-L276)

They have this concept of partial match. When the path is registered but the method is not present, it causes an exception and returns 405 meethod not allowed.

RobbeSneyders:
Closing in favor of #1736. I cherry-picked your commits there.

PR #1599, opened 2022-10-04
Move Swagger 2 form validation to middleware

RobbeSneyders:
This PR moves Swagger 2 form validation to the middleware by translating the Swagger 2 form parameter definition into an OpenAPI 3 RequestBody JsonSchema spec. This allows us to use the `FormDataValidator` introduced in #1595 for Swagger 2 as well instead of the old `ParameterValidator`.

### DIFF
--- a/connexion/operations/swagger2.py
+++ b/connexion/operations/swagger2.py
@@ -3,15 +3,25 @@ This module defines a Swagger2Operation class, a Connexion operation specific fo
 """
 
 import logging
+import typing as t
 from copy import deepcopy
 
 from connexion.operations.abstract import AbstractOperation
 
 from ..decorators.uri_parsing import Swagger2URIParser
 from ..exceptions import InvalidSpecification
+from ..http_facts import FORM_CONTENT_TYPES
 from ..utils import deep_get, is_null, is_nullable, make_type
 
 logger = logging.getLogger("connexion.operations.swagger2")
+
+
+COLLECTION_FORMAT_MAPPING = {
+    "multi": {"style": "form", "explode": True},
+    "csv": {"style": "form", "explode": False},
+    "ssv": {"style": "spaceDelimited", "explode": False},
+    "pipes": {"style": "pipeDelimited", "explode": False},
+}
 
 
 class Swagger2Operation(AbstractOperation):
@@ -135,7 +145,7 @@ class Swagger2Operation(AbstractOperation):
             security_schemes=spec.security_schemes,
             definitions=spec.definitions,
             *args,
-            **kwargs
+            **kwargs,
         )
 
     @property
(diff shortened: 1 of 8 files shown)


tests/decorators/test_validation.py
```diff
@@ -193,22 +193,3 @@ def test_writeonly_required_error():
     }
     with pytest.raises(ValidationError):
         Draft4RequestValidator(schema).validate({"bar": "baz"})
-
-
-def test_formdata_extra_parameter_strict():
```
RobbeSneyders, line 198: Formdata is no longer validated with the `ParameterValidator`. We still have API tests testing this behavior.

connexion/operations/swagger2.py
```diff
@@ -224,30 +234,95 @@ def _nested_example(self, schema):
...
+                    "{method} {path} There can be one 'body' parameter at most".format(
+                        method=self.method, path=self.path
+                    )
                 )
-            )
-        return body_parameters[0] if body_parameters else {}
+            body_definition = body_parameters[0] if body_parameters else {}
+        return body_definition
+
+    def _transform_form(self, form_parameters: t.List[dict]) -> dict:
```
Ruwann, line 266: Might be handy to add a unit test for this? Or do you think the other tests are sufficient?

tests/decorators/test_validation.py
Ruwann: We should still check for this behaviour, no? If there are any extra parameters when `strict_validation=True`?

RobbeSneyders: We still do here in the API tests:
https://github.com/spec-first/connexion/blob/1ab5400c0b0abe0483181dbbe972d00fa1d6e077/tests/api/test_parameters.py#L267-L276

connexion/operations/swagger2.py
RobbeSneyders: Yes a test would be good, I'll add one.

connexion/operations/swagger2.py
```diff
@@ -224,30 +234,95 @@ def _nested_example(self, schema):
...
+                "type": "object",
+                "properties": properties,
+                "required": required,
+                "additionalProperties": not self.strict_validation,
+            }
+        }
+
+        if collection_formats:
+            if len(collection_formats) > 1:
+                raise InvalidSpecification(
```
RobbeSneyders, line 317: As far as I'm aware, this and the `tsv` collection format are the only two aspects of the Swagger 2 parameter spec that cannot be translated to the OpenAPI 3 style request body spec.

I think these cases are rare enough that we can decide not to support them so we can unify form validation across versions.

RobbeSneyders: I'll add an explicit error for the `tsv` collection format as well though.

Ruwann: Indeed, I believe you're right. There are just a few exceptions where Swagger 2 and OpenAPI 3:

> This instance [tsv], and the case where a parameter has a type of array and a collectionFormat of x (e.g. pipes) and an items.collectionFormat of y (e.g. csv) (giving: a,b|c,d) seem to be the only parts of the 2.0 specification which can't be losslessly converted to 3.0.x

https://github.com/OAI/OpenAPI-Specification/issues/1046#issuecomment-297436141

RobbeSneyders: Ok, so multiple collection formats for a single operation does work, just not multiple collection formats for the same parameter / property. That makes it even more rare. Updated the code to reflect this.

Ruwann (approved):
I think this one is good to go - anything you want to still get double checked?

RobbeSneyders:
No, nothing specific. Some general polishing of the form validation still to do, but I'll do that separately.

PR #1600, opened 2022-10-04
Remove support for deprecated x-body-name position

RobbeSneyders:
This deprecated behavior should be removed for version 3.0.


Ruwann (approved):
Thanks!

PR #1604, opened 2022-10-16
Polish form validation

RobbeSneyders:
This PR cleans up some details of the form validation. With these changes, I think it's in a good state to merge.

I didn't pass the custom jsonifier to the validators, as this would require us to pass it along a lot, and add a JSON specific parameter to the validator interface. Instead, I further extracted the parsing so it can easily be overwritten by subclassing the validator. 
We might want to switch to a more performant json library and handle it as a singleton so it can easily be overwritten globally.

### DIFF
--- a/connexion/datastructures.py
+++ b/connexion/datastructures.py
@@ -1,0 +1,31 @@
+from fnmatch import fnmatch
+
+
+class MediaTypeDict(dict):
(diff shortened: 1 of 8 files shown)


nielsbox (approved):
LGTM

PR #1609, opened 2022-10-28
Add regex pattern routing for flask apis

nickvalin1:

 - Enable request routing using pattern regex in a path parameter

This feature is supported by Flask and OpenAPI spec and is a common feature for web frameworks like Django, Phalcon, .NET, etc.

The update to the docs show an example of how it works.

Has to be enabled by a feature flag to maintain full backwards compatibility with error codes for non-matching pattern requests. Let me know if there's a better place to add a feature flag than the options object, but it seemed like a good spot to me.

### DIFF
--- a/connexion/apis/flask_utils.py
+++ b/connexion/apis/flask_utils.py
@@ -44,7 +44,10 @@ def flaskify_endpoint(identifier, randomize=None):
 def convert_path_parameter(match, types):
     name = match.group(1)
     swagger_type = types.get(name)
-    converter = PATH_PARAMETER_CONVERTERS.get(swagger_type)
+    if swagger_type and swagger_type.startswith("regex"):
+        converter = swagger_type
+    else:
+        converter = PATH_PARAMETER_CONVERTERS.get(swagger_type)
     return "<{}{}{}>".format(
         converter or "", ":" if converter else "", name.replace("-", "_")
     )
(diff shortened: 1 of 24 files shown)


RobbeSneyders:
Thanks @nickvalin1!

We're working towards version 3.0 on the main branch, so we can make this a breaking change and remove the feature flag.

nickvalin1:
@RobbeSneyders thanks for running the workflows! If you think the 404 error code for non-matching patterns is the best default behavior for version 3, I have no qualms and will update the PR to remove the flag.

nickvalin1:
@RobbeSneyders I have removed the feature flag, let me know what you think!

RobbeSneyders (approved):
Thanks @nickvalin1, LGTM.

I believe a 400 error would be better than a 404, as the issue is that the resource identifier is malformed, not that the resource cannot be found. However I prefer to have a consistent 400 error over yet another config parameter to change this :slightly_smiling_face:

RobbeSneyders:
I'm unfortunately going to have to revert merging this PR for now. 

We're reworking Connexion into an ASGI middleware stack based on Starlette routing, which doesn't support regex as nicely out of the box. This means that the ASGI middleware stack might be validating the request against the wrong route. 

This is clear when combining this PR with PR #1610, as the `test_matched_path_param_pattern` test breaks because the parameters will be validated against the 'shard 1' route independent of the provided path argument. But the wrong route will be used for other middleware as well, eg. when defining different security on both routes.

Before merging this PR again, we need to add regex support to the `RoutingMiddleware` as well.

PR #1610, opened 2022-11-14
Move parameter validation to middleware

RobbeSneyders:
Fixes https://github.com/spec-first/connexion/issues/1525

This PR follows up on #1588 and #1591 and moves the last part of validation, the parameter validation, to the middleware,


connexion/decorators/uri_parsing.py
```diff
@@ -119,6 +120,15 @@ def resolve_params(self, params, _in):
             else:
                 resolved_param[k] = values[-1]
 
+            if not (is_nullable(param_defn) and is_null(resolved_param[k])):
+                try:
+                    # TODO: coerce types in a single place
```
RobbeSneyders, line 125: We now coerce the types in three locations
- Here, in the URI parser
- In the parameter validation
- In the form validation

I feel like we should be able to only do this in the URI parsing, but then we need some more work to still propagate the validation errors properly and make sure the URI parsing is used in all cases.

connexion/exceptions.py
```diff
@@ -206,3 +206,22 @@ def __init__(
                 )
 
         super().__init__(title=title, detail=detail, **kwargs)
+
+
+class TypeValidationError(Exception):
```
RobbeSneyders, line 211: Only moved

connexion/middleware/routing.py
```diff
@@ -25,6 +25,10 @@ def from_operation(cls, operation: AbstractOperation, next_app: ASGIApp):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         """Attach operation to scope and pass it to the next app"""
         original_scope = _scope.get()
+        # Pass resolved path params along
```
RobbeSneyders, line 28: Starlette `Request` relies on this.

connexion/utils.py
```diff
@@ -296,3 +298,53 @@ def extract_content_type(
                 mime_type = content_type
             break
     return mime_type, encoding
+
+
+def coerce_type(param, value, parameter_type, parameter_name=None):
```
RobbeSneyders, line 303: Only moved

connexion/validators/form_data.py
```diff
@@ -0,0 +1,150 @@
+import logging
```
RobbeSneyders, line 1: Only moved

connexion/validators/json.py
```diff
@@ -0,0 +1,155 @@
+import json
```
RobbeSneyders, line 1: Only moved

connexion/validators/parameter.py
```diff
@@ -0,0 +1,148 @@
+import collections
```
RobbeSneyders, line 1: Mostly moved with some small changes.

tests/api/test_parameters.py
```diff
@@ -172,7 +172,7 @@ def test_path_parameter_someint__bad(simple_app):
     # non-integer values will not match Flask route
     app_client = simple_app.app.test_client()
     resp = app_client.get("/v1.0/test-int-path/foo")  # type: flask.Response
-    assert resp.status_code == 404
+    assert resp.status_code == 400, resp.text
```
RobbeSneyders, line 175: I believe 400 is more correct than 404 when a wrong type of path parameter is passed.

tests/test_operation2.py
```diff
@@ -706,7 +706,7 @@ def test_oauth_scopes_in_or(security_handler_factory):
 
 
 def test_form_transformation(api):
-    mock_self = mock.Mock(strict_validation=True)
+    mock_self = mock.Mock()
```
RobbeSneyders, line 709: `strict_validation` for form data is checked by the `FormDataValidator`, we don't need to include it in the transformation here.

PR #1613, opened 2022-12-23
Add async app

### LINKED ISSUE #1534: Add 'native' async app
Next to the synchronous Flask app, we also want to provide an asynchronous app. Instead of implementing an existing async framework, I propose to implement a simple 'native' app. We can implement this app using a combination of low level starlette components and the routing lookup mechanism we already implemented in the middleware.

Instead of using the native app, the connexion middleware can also be applied to existing async frameworks to add connexion functionality.

#1395 for context.

RobbeSneyders:
Fixes #1534

 - Adds a "native" async app based on lower level Starlette tools. It leverages the routing already done in the middleware.
 - Extracts code related to the parameter decorator out of the Operation objects

There's some more work to be done, but this PR is already quite huge, so I wanted to stop at a point where I got all the tests and a simple example working.
- Still some general refactoring. I left some TODOs, and the Operation classes are still relied on in too many places.
- Sync routes on an Async app should receive async decorators, so they can access the body using `await`. The decorators should then run the sync route in a threadpool to prevent blocking.
- We should be able to reuse the parameter decorator for all ASGI apps since we're working with a Starlette Request which we can get directly from the scope and receive channel, which we can make available as context. This means we'll pass Starlette Datastructures to the view functions (eg. FormData, UploadFiles, ...), but if this is opt-in, I don't see any issue with this.
- We should be able to reuse the response parsing for Starlette apps as well since it returns a StarletteResponse.
- We should test the AsyncApp as well. I'm hoping we can do this quite quickly by generating a test client for it in the fixtures as well, but in the long term some more work will be needed.


connexion/apis/abstract.py
```diff
@@ -231,62 +234,44 @@ def add_operation(self, path, method):
...
     @abc.abstractmethod
-    def get_request(self, *args, **kwargs):
+    def get_request(cls, **kwargs) -> Request:
         """
         This method converts the user framework request to a ConnexionRequest.
         """
 
     @classmethod
-    @abc.abstractmethod
-    def get_response(self, response, mimetype=None, request=None):
```
RobbeSneyders, line 246: Should include this again without the request parameter.

connexion/apis/flask_api.py
```diff
@@ -130,75 +120,13 @@ def _build_response(
         return flask.current_app.response_class(**kwargs)
 
     @classmethod
-    def _serialize_data(cls, data, mimetype):
-        if isinstance(mimetype, str) and is_json_mimetype(mimetype):
-            body = cls.jsonifier.dumps(data)
-        elif not (isinstance(data, bytes) or isinstance(data, str)):
-            warnings.warn(
-                "Implicit (flask) JSON serialization will change in the next major version. "
```
RobbeSneyders, line 138: Removed this deprecated behavior and lifted the method to the base class.

connexion/apis/flask_api.py
```diff
@@ -130,75 +120,13 @@ def _build_response(
...
-            extra={
-                "data": request.body,
-                "data_type": type(request.body),
-                "url": request.url,
-            },
-        )
-        return request
+    def get_request(cls, **kwargs) -> ConnexionRequest:
+        uri_parser = kwargs.pop("uri_parser")
+        return ConnexionRequest(flask.request, uri_parser=uri_parser)
```
RobbeSneyders, line 125: I simplified the ConnexionRequest since I don't believe we should still provide such a high abstraction layer here anymore. This should only be implemented for Flask and Starlette. Other frameworks are supported via the ASGI interface.

connexion/apis/flask_api.py
```diff
@@ -130,75 +120,13 @@ def _build_response(
...
         Use Flask specific JSON loader
         """
         cls.jsonifier = Jsonifier(flask.json, indent=2)
-
-
-def _get_context():
-    return getattr(flask.globals.request_ctx, "connexion_context")
-
-
-context = LocalProxy(_get_context)
```
RobbeSneyders, line 204: Connexion context is now provided by the connexion.context module and the Context Middleware.

connexion/apps/async_app.py
```diff
@@ -0,0 +1,268 @@
...
+        try:
+            connexion_context = scope["extensions"][ROUTING_CONTEXT]
+        except KeyError:
+            raise MissingMiddleware(
+                "Could not find routing information in scope. Please make sure "
+                "you have a routing middleware registered upstream. "
+            )
+
+        api_base_path = connexion_context.get("api_base_path")
+        if api_base_path and not api_base_path == self.base_path:
```
RobbeSneyders, line 66: We might be able to leverage the RoutedMiddleware base classes here, will try to rework.

connexion/apps/async_app.py
```diff
@@ -0,0 +1,268 @@
...
+logger = logging.getLogger("Connexion.app")
+
+
+class AsyncAsgiApp:
+    """Mixin for usage with Abstract Apps & Apis"""
+
+    def __init__(self, *args, base_path="", **kwargs):
+        self.apis: t.Dict[str, AsyncApi] = {}
+        self.operations: t.Dict[str, AsyncOperation] = {}
+        self.router = Router()
```
RobbeSneyders, line 37: Add some documentation on this class about how the routing works (RoutingMiddleware + Router)

connexion/apps/flask_app.py
```diff
@@ -100,6 +108,29 @@ def add_error_handler(self, error_code, function):
         # type: (int, FunctionType) -> None
         self.app.register_error_handler(error_code, function)
 
+    def route(self, rule, **options):
```
RobbeSneyders, line 111: Decorator to allow for registration of additional routes on the Connexion App instead of having to use the underlying Flask app. The same behavior is offered by the async app.

TODO: ccheck if this can be lifted to the AbstractApp, or copy documentation to Async App as well.

connexion/decorators/parameter.py
```diff
@@ -39,91 +172,320 @@ def inspect_function_arguments(function):  # pragma: no cover
...
 
 
-def pythonic(name):
+def pythonic(name: str) -> str:
     name = name and snake_and_shadow(name)
     return sanitized(name)
 
 
-def parameter_to_arg(operation, function, pythonic_params=False):
+def get_arguments(
```
RobbeSneyders, line 198: Lots of this was moved from the Operation classes with minimal changes.

examples/openapi3/restyresolver/resty-api.yaml
```diff
@@ -127,6 +127,7 @@ components:
           format: int64
         name:
           type: string
+          format: date-time
```
RobbeSneyders, line 130: Don't know where this comes from.
TODO: remove again.

connexion/middleware/routing.py
```diff
@@ -129,3 +136,46 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+
+
+def convert_path_parameter(match, types):
+    name = match.group(1)
+    swagger_type = types.get(name)
+    converter = PATH_PARAMETER_CONVERTERS.get(swagger_type)
+    return f'{{{name.replace("-", "_")}{":" if converter else ""}{converter or ""}}}'
+
+
+def starlettify_path(swagger_path, types=None):
```
RobbeSneyders, line 152: This fixes routing for conflicting paths with path parameters with different types.

tests/api/test_parameters.py
```diff
@@ -172,7 +172,7 @@ def test_path_parameter_someint__bad(simple_app):
     # non-integer values will not match Flask route
     app_client = simple_app.app.test_client()
     resp = app_client.get("/v1.0/test-int-path/foo")  # type: flask.Response
-    assert resp.status_code == 400, resp.text
+    assert resp.status_code == 404, resp.text
```
RobbeSneyders, line 175: Due to the fix for the typed path params, we now return a 404 again instead of a 400 when sending the wrong type of path parameter. This is because the request is no longer matched by the router, so it doesn't hit the parameter validation. It is either rejected by the Middleware router, or directly passed to the App router, where it is rejected.

tests/fixtures/different_schemas/openapi.yaml
```diff
@@ -59,7 +59,7 @@ paths:
         '200':
           description: Requested new_stack data model
           content:
-            text/plain:
+            application/json:
```
RobbeSneyders, line 62: We no longer parse the response with json for non json response types.

connexion/apis/abstract.py
```diff
@@ -435,20 +406,22 @@ def _prepare_body_and_status_code(
...
+            extra={"body": body},
         )
 
         return body, status_code, mimetype
 
     @classmethod
-    @abc.abstractmethod
     def _serialize_data(cls, data, mimetype):
-        pass
+        if isinstance(mimetype, str) and is_json_mimetype(mimetype):
```
RobbeSneyders, line 419: Would be nice if we could make this pluggable.

tests/test_datastructures.py
```diff
@@ -0,0 +1,22 @@
+from connexion.datastructures import MediaTypeDict
+
+
+def test_media_type_dict():
```
RobbeSneyders, line 4: This file wasn't added to git yet in the PR where I introduced the MediaTypeDict type, so including it here.

tests/fixtures/simple/swagger.yaml
```diff
@@ -45,6 +29,22 @@ paths:
           required: true
           type: string
           format: path
+  /greeting/{name}:
```
RobbeSneyders, line 32: TODO: move this back to its original position.

RobbeSneyders (commented):
Self review containing some explanations and TODOs still for this PR.

tests/fixtures/simple/swagger.yaml
RobbeSneyders: The new order is actually needed since the routing behavior in the middleware was fixed and now correctly leverages Starlette routing. [Starlette route priority](https://www.starlette.io/routing/#route-priority) differs from Flask route priority:

> In cases where more that one route could match an incoming path, you should take care to ensure that more specific routes are listed before general cases.

RobbeSneyders:
Merging this so I can continue the work on fresh PRs.

PR #1615, opened 2022-12-28
Update examples for Connexion 3.0

RobbeSneyders:
This PR updates the examples for Connexion 3.0 and merges them for OpenAPI and Swagger.

2 examples required some changes to make them work:
- The reverse proxy example required some fixes to the SwaggerUIMiddleware to leverage the `root_path` correctly. This is included in the PR.
- The enforced defaults example requires the json validator to adapt the body and pass it on. We currently pass on the original body after validation, and I'm not sure if we should change this. I'll submit a separate PR to discuss this.


connexion/apps/abstract.py
```diff
@@ -20,9 +20,7 @@ def __init__(
         self,
         import_name,
         api_cls,
-        port=None,
```
RobbeSneyders, line 23: I see no reason to pass these to the app instead of to `app.run` directly.

connexion/apps/async_app.py
```diff
@@ -83,7 +83,11 @@ async def asgi_app(self, scope: Scope, receive: Receive, send: Send) -> None:
             )
 
         api_base_path = connexion_context.get("api_base_path")
-        if api_base_path and not api_base_path == self.base_path:
+        if (
```
RobbeSneyders, line 86: Handle case when base path is "", which is the case both when endpoint is not part of an API, or when endpoint is part of the API registered on the root path.

connexion/apps/flask_app.py
```diff
@@ -124,89 +118,11 @@ def index():
         logger.debug("Adding %s with decorator", rule, extra=kwargs)
         return self.app.route(rule, **kwargs)
 
-    def run(
```
RobbeSneyders, line 127: Lifted to AbstractApp.

connexion/decorators/parameter.py
```diff
@@ -38,6 +38,7 @@ def parameter_to_arg(
     sanitize = pythonic if pythonic_params else sanitized
     arguments, has_kwargs = inspect_function_arguments(function)
 
+    # TODO: should always be used for AsyncApp
```
RobbeSneyders, line 41: This needs some work. Will submit a separate PR for this.

connexion/middleware/main.py
```diff
@@ -55,7 +55,7 @@ def _apply_middlewares(
         for middleware in reversed(middlewares):
             app = middleware(app)  # type: ignore
             apps.append(app)
-        return app, reversed(apps)
+        return app, list(reversed(apps))
```
RobbeSneyders, line 58: Make sure iterator doesn't run out when registering multiple apis.

connexion/middleware/swagger_ui.py
```diff
@@ -42,16 +43,11 @@ def __init__(self, *args, default: ASGIApp, **kwargs):
     def normalize_string(string):
         return re.sub(r"[^a-zA-Z0-9]", "_", string.strip("/"))
 
-    def _base_path_for_prefix(self, request):
+    def _base_path_for_prefix(self, request: StarletteRequest) -> str:
```
RobbeSneyders, line 46: Fixed for proper usage of root_path behind reverse proxy.

examples/apikey/spec/openapi.yaml
```diff
@@ -2,6 +2,8 @@ openapi: 3.0.0
 info:
   title: API Key Example
   version: '1.0'
+servers:
+  - url: /openapi
```
RobbeSneyders, line 6: We need a base path now that we're registering two apis on the same app.

examples/basicauth/spec/swagger.yaml
```diff
@@ -1,9 +1,8 @@
 swagger: "2.0"
-
 info:
   title: Basic Auth Example
   version: "1.0"
-
+basePath: /swagger
```
RobbeSneyders, line 5: We need a base path now that we're registering two apis on the same app.

examples/enforcedefaults/app.py
```diff
@@ -1,12 +1,13 @@
...
 
 import connexion
 import jsonschema
-from connexion.decorators.validation import RequestBodyValidator
 from connexion.json_schema import Draft4RequestValidator
+from connexion.validators import JSONRequestBodyValidator
 
 
-def echo(data):
+# TODO: should work as sync endpoint when parameter decorator is fixed
```
RobbeSneyders, line 9: Currently the parameter decorator uses a sync decorator when the view function is sync, even when the framework is async. This doesn't work as the async body cannot be read by a sync decorator.

setup.py
```diff
@@ -100,7 +104,8 @@ def readme():
         'tests': tests_require,
         'flask': flask_require,
         'swagger-ui': swagger_ui_require,
-        'docs': docs_require
+        'docs': docs_require,
+        'uvicorn': uvicorn_requires,
```
RobbeSneyders, line 108: Install with `connexion[uvicorn]` to enable `app.run`. The `standard` extra includes the dependencies needed to watch files for reloading the app.

RobbeSneyders (commented):
Self review.

Subin-Qreative:
https://github.com/spec-first/connexion/blob/main/examples/jwt/app.py

the following example is targeted at FlaskApp  would the same work out of the box by just replacing with AsyncApp ?

RobbeSneyders:
Yes it should @Subin-Qreative!

PR #1616, opened 2022-12-28
Enable enforcing defaults

RobbeSneyders:
This PR is a proposal to enforce defaults in json bodies.

I'm not sure which way to go here
- I think connexion should offer the functionality to fill in the defaults, as this helps enforce the contract. If a value is not provided, the application gets the default. We should probably offer this using a simple `default` flag instead of the custom validator now needed.
- I don't like that validation adapts the body. Validation is a read-only operation and it would be nice to keep it that way.
- Splitting enforcing defaults from validation would be less efficient since we'd need to traverse the body multiple times.

If we do want to go this way, we should probably implement a similar fix for form data, as we also use json validators there.

### DIFF
--- a/connexion/apps/asynchronous.py
+++ b/connexion/apps/asynchronous.py
@@ -65,7 +65,8 @@ class AsyncOperation:
     async def __call__(
         self, scope: Scope, receive: Receive, send: Send
     ) -> StarletteResponse:
-        return await self.fn(scope=scope, receive=receive, send=send)
+        response = await self.fn()
+        return await response(scope, receive, send)
 
 
 class AsyncApi(RoutedAPI[AsyncOperation]):
@@ -80,6 +81,9 @@ class AsyncApi(RoutedAPI[AsyncOperation]):
 
 
 class AsyncMiddlewareApp(RoutedMiddleware[AsyncApi]):
+
+    api_cls = AsyncApi
+
     def __init__(self) -> None:
         self.apis: t.Dict[str, AsyncApi] = {}
         self.operations: t.Dict[str, AsyncOperation] = {}
@@ -87,8 +91,7 @@ class AsyncMiddlewareApp(RoutedMiddleware[AsyncApi]):
         super().__init__(self.router)
 
     def add_api(self, *args, **kwargs):
-        api = AsyncApi(*args, **kwargs)
-        self.apis[api.base_path] = api
+        api = super().add_api(*args, **kwargs)
         self.router.mount(api.base_path, api.router)
         return api
 
--- a/connexion/decorators/main.py
+++ b/connexion/decorators/main.py
@@ -166,10 +166,7 @@ class ASGIDecorator(BaseDecorator):
                 uri_parser=self.uri_parser, scope=scope, receive=receive
             )
             decorated_function = self.decorate(function)
-            response = decorated_function(request)
-            while asyncio.iscoroutine(response):
-                response = await response
-            return response
+            return await decorated_function(request)
 
         return wrapper
 
--- a/connexion/middleware/abstract.py
+++ b/connexion/middleware/abstract.py
@@ -25,7 +25,7 @@ class SpecMiddleware(abc.ABC):
     @abc.abstractmethod
     def add_api(
         self, specification: t.Union[pathlib.Path, str, dict], **kwargs
-    ) -> None:
+    ) -> t.Any:
         """
         Register an API represented by a single OpenAPI specification on this middleware.
         Multiple APIs can be registered on a single middleware.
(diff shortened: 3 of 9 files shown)


RobbeSneyders:
I updated the PR with a solution that I think is a good middle ground.

The standard `JsonRequestBodyValidator` does not adapt the body. Instead, I added a `DefaultsJsonRequestBodyValidator` which does. The user can easily activate enforcing default by passing this Validator in a `validator_map`.

I would have liked connexion to have an `enforce_defaults` flag, but it would be hard to make this compatible with the `validator_map` argument.

connexion/apps/asynchronous.py
```diff
@@ -63,7 +63,8 @@ def fn(self) -> t.Callable:
     async def __call__(
         self, scope: Scope, receive: Receive, send: Send
     ) -> StarletteResponse:
-        return await self.fn(scope=scope, receive=receive, send=send)
```
RobbeSneyders, line 66: Just a fix to get the `AsyncApp` working correctly. Not related to enforcing defaults, but necessary to make the example work.

connexion/apps/asynchronous.py
```diff
@@ -78,15 +79,17 @@ def make_operation(self, operation: AbstractOperation) -> AsyncOperation:
...
     def __init__(self) -> None:
         self.apis: t.Dict[str, AsyncApi] = {}
         self.operations: t.Dict[str, AsyncOperation] = {}
         self.router = Router()
         super().__init__(self.router)
 
     def add_api(self, *args, **kwargs):
-        api = AsyncApi(*args, **kwargs)
-        self.apis[api.base_path] = api
+        api = super().add_api(*args, **kwargs)
```
RobbeSneyders, line 92: Just a fix to get the `AsyncApp` working correctly. Not related to enforcing defaults, but necessary to make the example work.

connexion/decorators/main.py
```diff
@@ -166,10 +166,7 @@ async def wrapper(*args, **kwargs):
                 uri_parser=self.uri_parser, scope=scope, receive=receive
             )
             decorated_function = self.decorate(function)
-            response = decorated_function(request)
-            while asyncio.iscoroutine(response):
-                response = await response
-            return response
+            return await decorated_function(request)
```
RobbeSneyders, line 169: Just a fix to get the `AsyncApp` working correctly. Not related to enforcing defaults, but necessary to make the example work.

connexion/middleware/abstract.py
```diff
@@ -244,11 +244,10 @@ def __init__(self, app: ASGIApp) -> None:
         self.app = app
         self.apis: t.Dict[str, API] = {}
 
-    def add_api(
-        self, specification: t.Union[pathlib.Path, str, dict], **kwargs
-    ) -> None:
+    def add_api(self, specification: t.Union[pathlib.Path, str, dict], **kwargs) -> API:
         api = self.api_cls(specification, next_app=self.app, **kwargs)
         self.apis[api.base_path] = api
+        return api
```
RobbeSneyders, line 250: Just a fix to get the `AsyncApp` working correctly. Not related to enforcing defaults, but necessary to make the example work.

examples/sqlalchemy/spec/openapi.yaml
```diff
@@ -62,10 +62,10 @@ paths:
         '201':
           description: New pet created
       requestBody:
+        x-body-name: pet
```
RobbeSneyders, line 65: Unrelated fix

connexion/validators/json.py
```diff
@@ -52,30 +52,107 @@ def validate(self, body: dict):
...
+            validator_class, {"properties": set_defaults}
+        )
+
+    async def read_body(self) -> t.Tuple[str, int]:
+        """Read the body from the receive channel.
+
+        :return: A tuple (body, max_length) where max_length is the length of the larges message.
+        """
+        more_body = True
+        max_length = 256000
```
Ruwann, line 114: Where does this number come from? I'm not aware of an ASGI limit

connexion/validators/json.py
```diff
@@ -52,30 +52,107 @@ def validate(self, body: dict):
...
+            yield from validate_properties(validator, properties, instance, schema)
+
+        return jsonschema.validators.extend(
+            validator_class, {"properties": set_defaults}
+        )
+
+    async def read_body(self) -> t.Tuple[str, int]:
+        """Read the body from the receive channel.
+
+        :return: A tuple (body, max_length) where max_length is the length of the larges message.
```
Ruwann, line 111: typo: largest

Ruwann (approved):
I agree that the proposed validator is a good middleground :)

Perhaps also good to put this in the updated documentation, as it might otherwise be hard to find

connexion/validators/json.py
RobbeSneyders: This is not constrained by ASGI, but by the server. I ran some tests with Uvicorn and this was the upper limit, although I couldn't find where this limit is defined.

PR #1618, opened 2023-01-03
Refactor decorators

RobbeSneyders:
This PR refactors the decorators so they can be used independently of an API, which will allow them to be used by other wrapped frameworks as well.


connexion/apis/abstract.py
```diff
@@ -6,35 +6,25 @@
...
 from connexion.spec import Specification
-from connexion.utils import is_json_mimetype
 
 MODULE_PATH = pathlib.Path(__file__).absolute().parent.parent
 SWAGGER_UI_URL = "ui"
 
 logger = logging.getLogger("connexion.apis.abstract")
 
 
-class AbstractAPIMeta(abc.ABCMeta):
```
RobbeSneyders, line 31: Don't see a reason for this to be a Metaclass. I would still like to rework how the jsonifier is defined and passed around.

connexion/jsonifier.py
```diff
@@ -41,7 +41,7 @@ def wrapped_default(self, o):
         if isinstance(o, uuid.UUID):
             return str(o)
 
-        return default_fn(o)
+        return default_fn(self, o)
```
RobbeSneyders, line 44: We're wrapping a method.

connexion/operations/abstract.py
```diff
@@ -243,45 +228,20 @@ def get_mimetype(self):
             return DEFAULT_MIMETYPE
 
     @property
-    def _uri_parsing_decorator(self):
+    def uri_parser_class(self):
```
RobbeSneyders, line 231: Would prefer not to pass this via the operation. We can probably handle it similarly to the Jsonifier.

connexion/decorators/response.py
```diff
@@ -0,0 +1,190 @@
...
+
+        :return: Deducted content type
+
+        :raises: NonConformingResponseHeaders if content type cannot be deducted.
+        """
+        content_type = headers.get("Content-Type")
+
+        # TODO: don't default
+        produces = list(set(self.operation.produces))
+        if data is not None and not produces:
```
RobbeSneyders, line 56: This keeps backward compatibility, although I don't think we should. We should expect the spec to define a content type when a response is returned.

https://swagger.io/specification/v2/#responseObject
- A missing response schema defines an empty response
- If a response schema is present, produces should be defined

In OpenAPI 3, a response cannot be defined without its content type.

tests/api/test_responses.py
```diff
@@ -139,7 +139,12 @@ def test_pass_through(simple_app):
     app_client = simple_app.app.test_client()
 
     response = app_client.get("/v1.0/multimime", data={})  # type: flask.Response
-    assert response.status_code == 200
+    assert response.status_code == 500
+    assert (
```
RobbeSneyders, line 143: This is a change of behavior. If the spec defines multiple content types, we should not guess.

tests/decorators/test_parameter.py
```diff
@@ -1,67 +1,116 @@
 from unittest.mock import MagicMock
 
-from connexion.decorators.parameter import parameter_to_arg, pythonic
+from connexion.decorators.parameter import (
```
RobbeSneyders, line 3: Test both Sync & Async decorator. Refactored to work with new decorators.

RobbeSneyders (commented):
Self-review

PR #1621, opened 2023-01-15
Implement user facing interface for ConnexionMiddleware

RobbeSneyders:
This PR adds an interface for the ConnexionMiddleware, similar to the interface of the Connexion Apps.

The Connexion Apps are now a simple wrapper around the ConnexionMiddleware and framework app, delegating the work to the middleware. This enables a similar interface and behavior for users when using either the middleware or apps.

The arguments are repeated everywhere there is a user interface, but are parsed in a central place. Repeating the arguments is not DRY, but needed to provide users with IDE autocomplete, typing, etc. They are parsed in a single `_Options` class, which also provides a mechanism to set default options on an App level, and override them on the more granular API level.

This makes the long list of provided parameters a lot more manageable, so I would like to use it for the `Jsonifier` as well, and re-add the `debug` and `extra_files` arguments which I have dropped in previous PRs. I'll submit a separate PR for this.

I renamed the `options` parameter to `swagger_ui_options` since it only contains swagger UI options. This is a breaking change though, and we'll need to highlight this upon release.

We still have quite a lot of `App`, `MiddlewareApp`, and abstract classes. It would be great if we could find a way to reduce those further, or at least find better naming to make it more clear what each one does :slightly_smiling_face: .

Finally, I added examples on how the middleware can be used with third party frameworks under `examples/frameworks`. Currently there's an example for Starlette and Quart, but this should be easy to extend. They also show how the `ASGIDecorator` and `StarletteDecorator` from my previous PR can be used.


connexion/middleware/abstract.py
```diff
@@ -1,42 +1,185 @@
...
+
+    def __init__(
+        self,
+        *args,
+        pythonic_params=False,
+        resolver_error_handler: t.Optional[t.Callable] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.pythonic_params = pythonic_params
```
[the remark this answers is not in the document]
RobbeSneyders: It is used in subclasses.

connexion/apps/asynchronous.py
```diff
@@ -0,0 +1,123 @@
...
+        **options
+    ):
+        self.router.add_route(rule, endpoint=view_func, name=endpoint, methods=methods)
+
+
+class AsyncApp(AbstractApp):
+    """Connexion Application based on ConnexionMiddleware wrapping a async Connexion application
+    based on starlette tools."""
+
+    middleware_app = AsyncMiddlewareApp()
```
Ruwann, line 108: Is it intentional that this is a class attribute and not an instance attribute?

connexion/apps/asynchronous.py
```diff
@@ -0,0 +1,129 @@
...
+        super().__init__(*args, **kwargs)
+        self.pythonic_params = pythonic_params
+        self.router = Router()
+        self.add_paths()
+
+    def make_operation(self, operation: AbstractOperation) -> AsyncOperation:
+        return AsyncOperation.from_operation(operation, self.pythonic_params)
+
+
+class AsyncMiddlewareApp(RoutedMiddleware[AsyncApi]):
```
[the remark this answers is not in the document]
Ruwann: Well...

> There are only two hard things in Computer Science: cache invalidation and naming things.

This class is also an app, but a different app than the `AsyncApp`?

connexion/middleware/abstract.py
```diff
@@ -1,42 +1,185 @@
...
+from connexion.exceptions import MissingMiddleware, ResolverError
 from connexion.http_facts import METHODS
+from connexion.jsonifier import Jsonifier
 from connexion.operations import AbstractOperation
-from connexion.resolver import ResolverError
+from connexion.resolver import Resolver
+from connexion.spec import Specification
 
-logger = logging.getLogger("connexion.middleware.abstract")
+logger = logging.getLogger(__name__)
```
Ruwann, line 16: small nit: There is a mix of `__name__` and strings across files, also in this PR.

connexion/middleware/abstract.py
```diff
@@ -1,42 +1,185 @@
...
+class SpecMiddleware(abc.ABC):
+    """Middlewares that need the specification(s) to be registered on them should inherit from this
+    base class"""
 
     @abc.abstractmethod
     def add_api(
         self, specification: t.Union[pathlib.Path, str, dict], **kwargs
     ) -> None:
+        """
+        Register een API represented by a single OpenAPI specification on this middleware.
```
Ruwann, line 30: Typo: Register *een* API

connexion/middleware/abstract.py
```diff
@@ -1,42 +1,185 @@
...
+    def _set_base_path(self, base_path: t.Optional[str] = None) -> None:
+        if base_path is not None:
+            # update spec to include user-provided base_path
+            self.specification.base_path = base_path
+            self.base_path = base_path
+        else:
+            self.base_path = self.specification.base_path
+
+
+OP = t.TypeVar("OP")
```
Ruwann, line 70: small nit but comment to explain the TypeVar?

connexion/middleware/main.py
```diff
@@ -1,19 +1,86 @@
...
+        """Returns a factory to create a ResolverErrorHandler."""
+        if self.resolver_error is not None:
+
+            def resolver_error_handler(*args, **kwargs) -> ResolverErrorHandler:
+                return ResolverErrorHandler(self.resolver_error, *args, **kwargs)
+
+            return resolver_error_handler
+        return None
+
+    def replace(self, **changes) -> "_Options":
```
Ruwann, line 70: Not ideal that this is a separate function that needs to be called (if I understand correctly) - is there a way to make it part of the initialization of the class?

connexion/utils.py
```diff
@@ -354,3 +357,51 @@ def cast_leaves(d, schema):
             raise TypeValidationError(param_type, parameter_type, parameter_name)
         except TypeError:
             return value
+
+
+def get_root_path(import_name: str) -> str:
+    """Copied from Flask:
```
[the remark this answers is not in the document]
Ruwann: not a lawyer but there might be some implications for the license?

examples/frameworks/hello_quart.py
```diff
@@ -0,0 +1,28 @@
...
+from connexion.resolver import RelativeResolver
+from quart import Quart, json, request
+
+app = Quart(__name__)
+
+
+@app.route("/openapi/greeting/<name>", methods=["POST"])
+@app.route("/swagger/greeting/<name>", methods=["POST"])
+@ASGIDecorator()
+def post_greeting(name: str, number: int = None) -> str:
```
Ruwann, line 14: nit but the type for number is `Optional[int]` 😅

Ruwann (commented):
Had a first look, haven't fully grokked it yet, so while I left some smaller comments, didn't really have substantive comments

connexion/middleware/abstract.py
RobbeSneyders: Yes, I think `__name__` is a better choice than hardcoding. So I'm changing it when I touch the code, but haven't applied this change in a structured way.

connexion/middleware/main.py
RobbeSneyders: This method is called after the initialization to update the instance with new options.

The instance is created in the Middleware `__init__` method and updated in the `add_api` method. This way, we can set default options for the middleware and override them on an API level.

connexion/apps/asynchronous.py
RobbeSneyders: Not really, I just didn't want to repeat the whole `__init__` method with all its parameters here. But I guess this could prove to be a problem with parallelism / concurrency. Will change it.

connexion/utils.py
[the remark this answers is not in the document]
RobbeSneyders: As far as I understand, BSD3 (Flask's license) is compatible with Apache 2 as long as you copy the notice, which can be found with the provided link.

connexion/middleware/abstract.py
RobbeSneyders: What would you want it to say?
It's a trivial use of a type variable, so there's not really much to document about it beyond explaining what a type variable is, which I think is better left to the Python docs.

Ruwann: True, the TypeVar use itself shouldn't be explained.
It just took me a while this indicate an Operation, and not an abbreviation "OP" 😅

connexion/middleware/abstract.py
```diff
@@ -1,42 +1,186 @@
...
+""""""
```
Ruwann, line 71: A random string appeared? 😁

connexion/middleware/main.py
Ruwann: Got it now, the split is intentional :) I thought it was always something like
```python
options = _Options(...)
options.replace(...)
```

Ruwann (approved):

connexion/middleware/abstract.py
RobbeSneyders: Woops, seems like I forgot to push my latest fixes.

PR #1626, opened 2023-01-26
Cache operation body definition

RobbeSneyders:
Fixes a small todo I left in the code before to cache the body definition on the operation.


Ruwann (approved):
LGTM!

PR #1627, opened 2023-01-26
Coerce types only in uri parser

RobbeSneyders:
This PR moves all type coercing into the URI parsers and makes sure it's only done once for each code path.

### DIFF
--- a/connexion/validators/form_data.py
+++ b/connexion/validators/form_data.py
@@ -6,14 +6,10 @@ from starlette.datastructures import FormData, Headers, UploadFile
 from starlette.formparsers import FormParser, MultiPartParser
 from starlette.types import Receive, Scope
 
-from connexion.exceptions import (
-    BadRequestProblem,
-    ExtraParameterProblem,
-    TypeValidationError,
-)
+from connexion.exceptions import BadRequestProblem, ExtraParameterProblem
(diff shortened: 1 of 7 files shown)


Ruwann (approved):

connexion/validators/form_data.py
```diff
@@ -94,30 +81,29 @@ def validate(self, data: FormData) -> None:
...
 
-                try:
-                    data[k] = coerce_type(param_defn, data[k], "requestBody", k)
-                except TypeValidationError as e:
-                    logger.exception(e)
-                    errs += [str(e)]
+    def _validate_strictly(self, data: FormData) -> None:
+        form_params = data.keys()
+        spec_params = self.schema.get("properties", {}).keys()
+        errors = set(form_params).difference(set(spec_params))
```
Ruwann, line 98: Is the set conversion actually needed here? Doesn't the keyview support set operations directly?

RobbeSneyders: Only via bitwise operations. So `form_params - set(spec_params)` would work and I guess that's easier to understand. I'll do it in a separate PR though, since I'm afraid the pipeline on this PR would start failing due to [an issue](https://github.com/PyCQA/isort/issues/2083) for which I already included a fix in #1628.

PR #1634, opened 2023-02-13
Make tests framework agnostic

RobbeSneyders:
Currently, our tests only run against the `FlaskApp`. This PR parametrizes the relevant tests to run against both the `AsyncApp` and `FlaskApp`. 

Unfortunately, the test clients provided by Flask and Starlette are not entirely aligned. This PR addresses this by refactoring the tests so they are framework agnostic.

This increased the number of tests from 483 to 709, of which 66 still fail due to issues with the `AsyncApp`. I will submit separate PRs targeting this branch to fix those.

### DIFF
--- a/connexion/__init__.py
+++ b/connexion/__init__.py
@@ -18,8 +18,6 @@ from .resolver import Resolution, Resolver, RestyResolver  # NOQA
 from .utils import not_installed_error  # NOQA
 
 try:
-    from flask import request  # NOQA
-
     from connexion.apps.flask import FlaskApi, FlaskApp
 except ImportError as e:  # pragma: no cover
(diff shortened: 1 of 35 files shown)


RobbeSneyders:
> Unfortunately, the test clients provided by Flask and Starlette are not entirely aligned. This PR addresses this by refactoring the tests so they are framework agnostic.

I added a new commit which uses the Starlette `TestClient` for the `FlaskApp` as well. Since both apps are wrapped by ASGI middleware, this works, and we no longer need to write our tests to match both test client interfaces.

62 tests still failing.

Ruwann (approved):

PR #1641, opened 2023-02-17
Bump starlette version

Ruwann:
Addresses https://github.com/encode/starlette/security/advisories/GHSA-74m5-2c7w-9w3x


RobbeSneyders (approved):
Thanks!

PR #1646, opened 2023-02-19
Add high-level documentation for 3.0

RobbeSneyders:
I'd like to cut an alpha release for 3.0 while we further polish the code base and update the documentation. This way users can already test 3.0 and provide feedback.

This PR adds some very high-level documentation to serve as a starting point for users to test the alpha version.


docs/v3.rst
```diff
@@ -0,0 +1,72 @@
+.. warning::
+
+    Do not use Connexion 3.0 in production yet!
+
+    Connexion 3.0 is currently available in alpha to give users the ability to test it and provide
+    feedback while we further polish the codebase and update the documentation.
```
Ruwann, line 6: Small one: Let's make it explicit what the best way is for users to provide feedback? 
--> by creation an separate issue
--> as comment on a general feedback 3.x issue

Ruwann (approved):

docs/v3.rst
RobbeSneyders: I'll add this in the announcement on Github. We can then later refer to that announcement in this doc, but currently it's a bit chicken and egg :)

PR #1653, opened 2023-02-27
Create abstract validator classes

RobbeSneyders:
As promised, an `AbstractRequestBodyValidator` class. This has 2 advantages:
- Groups all common code
- Provides a base for custom validators

Looking at the remaining logic in the subclasses, I'm quite happy with the current abstraction. A lot of the complexity comes from substituting the default body when no body is received.

I'll leave some additional comments in-line.

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -107,6 +107,13 @@ def resolve_refs(spec, store=None, base_uri=""):
     return res
 
 
+def format_error_with_path(exception: ValidationError) -> str:
+    """Format a `ValidationError` with path to error."""
+    error_path = ".".join(str(item) for item in exception.path)
+    error_path_msg = f" - '{error_path}'" if error_path else ""
+    return error_path_msg
+
+
 def allow_nullable(validation_fn: t.Callable) -> t.Callable:
     """Extend an existing validation function, so it allows nullable values to be null."""
 
--- a/connexion/middleware/exceptions.py
+++ b/connexion/middleware/exceptions.py
@@ -65,5 +65,5 @@ class ExceptionMiddleware(StarletteExceptionMiddleware):
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         # Needs to be set so starlette router throws exceptions instead of returning error responses
-        scope["app"] = self
+        scope["app"] = "connexion"
         await super().__call__(scope, receive, send)
--- a/connexion/middleware/request_validation.py
+++ b/connexion/middleware/request_validation.py
@@ -68,8 +68,6 @@ class RequestValidationOperation:
             )
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send):
-        receive_fn = receive
-
         # Validate parameters & headers
         uri_parser_class = self._operation._uri_parser_class
         uri_parser = uri_parser_class(
@@ -100,8 +98,6 @@ class RequestValidationOperation:
                 )
             else:
                 validator = body_validator(
-                    scope,
-                    receive,
                     schema=schema,
                     required=self._operation.request_body.get("required", False),
                     nullable=utils.is_nullable(
@@ -113,9 +109,9 @@ class RequestValidationOperation:
                         self._operation.parameters, self._operation.body_definition()
                     ),
                 )
-                receive_fn = await validator.wrapped_receive()
+                receive = await validator.wrap_receive(receive, scope=scope)
 
-        await self.next_app(scope, receive_fn, send)
+        await self.next_app(scope, receive, send)
 
 
 class RequestValidationAPI(RoutedAPI[RequestValidationOperation]):
--- a/connexion/operations/swagger2.py
+++ b/connexion/operations/swagger2.py
@@ -291,6 +291,7 @@ class Swagger2Operation(AbstractOperation):
 
             default = param.get("default")
             if default is not None:
+                prop["default"] = default
                 defaults[param["name"]] = default
 
             nullable = param.get("x-nullable")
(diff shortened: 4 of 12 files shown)


connexion/json_schema.py
```diff
@@ -107,6 +107,13 @@ def _do_resolve(node):
     return res
 
 
+def format_error_with_path(exception: ValidationError) -> str:
```
RobbeSneyders, line 110: Extracted from the validator class.

connexion/middleware/exceptions.py
```diff
@@ -65,5 +65,5 @@ def common_error_handler(_request: StarletteRequest, exc: Exception) -> Response
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         # Needs to be set so starlette router throws exceptions instead of returning error responses
-        scope["app"] = self
+        scope["app"] = "connexion"
```
RobbeSneyders, line 68: Including a string here is more lightweight and easier to handle than a module. I believe I changed this when I ran into a pickle error, although I forgot where it originated.

connexion/middleware/request_validation.py
```diff
@@ -113,9 +112,9 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send):
                         self._operation.parameters, self._operation.body_definition()
                     ),
                 )
-                receive_fn = await validator.wrapped_receive()
+                next_app = validator
```
RobbeSneyders, line 115: Validators are now ASGI apps. Turtles all the way down :)

connexion/operations/swagger2.py
```diff
@@ -291,7 +290,7 @@ def _transform_form(self, form_parameters: t.List[dict]) -> dict:
 
             default = param.get("default")
             if default is not None:
-                defaults[param["name"]] = default
+                prop["default"] = default
```
RobbeSneyders, line 294: Also needs to be added for each property separately, so if a body is sent without the property, the default value is added.

connexion/operations/swagger2.py
```diff
@@ -320,11 +321,11 @@ def _transform_form(self, form_parameters: t.List[dict]) -> dict:
             "schema": {
                 "type": "object",
                 "properties": properties,
-                "default": defaults,
                 "required": required,
             }
         }
-
+        if defaults:
+            definition["schema"]["default"] = defaults
```
RobbeSneyders, line 328: No longer set an empty dict as default if no defaults are defined.

connexion/validators/form_data.py
```diff
@@ -1,85 +1,59 @@
...
-        self.check_empty()
+        super().__init__(
+            next_app,
+            schema=schema,
+            required=required,
+            nullable=nullable,
+            encoding=encoding,
+            strict_validation=strict_validation,
+        )
+        self._uri_parser = uri_parser
```
RobbeSneyders, line 39: Not completely happy with this. The `__init__` arguments for the validators are fixed in the `request_validation` , so it doesn't really make sense that individual validators define additional parameters.
On the other hand, we do support form data specifically, so I can live with it. Open for better solutions though.

RobbeSneyders:
We should provide a similar base class for the `ResponseBodyValidator`s.

connexion/middleware/request_validation.py
RobbeSneyders: Actually, creating the `AbstractResponseBodyValidator` class, I remembered that the response validators cannot be ASGI Apps, as they can only be created at response time. So reverted this to keep the different validators aligned.

RobbeSneyders:
Added an `AbstractResponseBodyValidator` class and aligned the `Request` one. Commits are atomic, which is useful to review.

connexion/validators/form_data.py
Ruwann: Don't have a better idea at the moment :)

tests/api/test_parameters.py
```diff
@@ -443,11 +446,8 @@ def test_nullable_parameter(simple_app):
     resp = app_client.put("/v1.0/nullable-parameters", content="null", headers=headers)
     assert resp.json() == "it was None"
 
-    resp = app_client.put("/v1.0/nullable-parameters", content="None", headers=headers)
```
Ruwann, line 446: Why can we delete this one now?

Ruwann (approved):

tests/api/test_parameters.py
RobbeSneyders: In query parameters it's impossible to represent `null` there:
- `?key=` could be interpreted as an empty string
- `?key` is not clear as discussed in this issue: 
https://github.com/spec-first/connexion/issues/1538

So connexion supports `"null"` and `"None"` as null values, so you can send:
- `?key=null`
- `?key=None`
(I'm actually not a fan of supporting `"None"` here, since it's Python specific, but if we still need to support `"null"`, the benefit of dropping it is probably too small.)

In json, null can be represented:
```
{
    "key": null
}
```
or for a `null` body (eg. try `json.dumps(None)`):
```
"null"
```
So we don't need to support `"None"` here. 
We actually also don't need to support `"null"` here, since we should always parse it before checking, in which case it's parsed to `None`. This is why I replaced all `is_null(body)` checks in the body validators with `body is None`.

connexion/validators/form_data.py
```diff
@@ -1,85 +1,57 @@
...
-        receive: Receive,
         *,
         schema: dict,
-        validator: t.Type[Draft4Validator] = None,
         required=False,
         nullable=False,
         encoding: str,
-        uri_parser: t.Optional[AbstractURIParser] = None,
         strict_validation: bool,
+        uri_parser: t.Optional[AbstractURIParser] = None,
```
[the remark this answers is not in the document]
Ruwann: Yes, true for the case of `application/x-www-form-urlencoded` but for `multipart/form-data` the parameters are sent in different parts instead of query string.
But indeed, not necessarily as surprising as I thought (URI <> body) :)

PR #1664, opened 2023-03-04
Fix and extend v3 documentation

RobbeSneyders:


RobbeSneyders:
https://connexion--1664.org.readthedocs.build/en/1664/

PR #1667, opened 2023-03-05
Add explicit asgiref dependency

RobbeSneyders:
We use `asgiref` as a direct dependency
https://github.com/spec-first/connexion/blob/ab94e2aaa67d30ab2bb97e4d8d27746a5ed43d20/connexion/decorators/main.py#L7

So should add it explicitly as well.

This issue is hidden in our tests since we install all extras. To prevent this, we need to address https://github.com/spec-first/connexion/issues/1389.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ connexion = 'connexion.cli:main'
 
 [tool.poetry.dependencies]
 python = '^3.7'
+asgiref = "^3.4"
 clickclick = ">= 1.2, < 21"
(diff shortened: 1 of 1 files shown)


Ruwann (approved):
Is this the lowest version possible? Is it worth it to lower this if possible?

RobbeSneyders:
I lowered it to match the lower bound that Flask defines.

RobbeSneyders:
That didn't work. Bumped it back up to match the lower bound defined by uvicorn.

PR #1671, opened 2023-03-07
Make security pluggable

Ruwann:
Make security pluggable

- [x] Solution for standard security handlers: `security_deny`, `security_passthrough`, `verify_none` 
- [x] HTTP security handlers & overlap with basic from swagger 2
- [x] Do we need a separate handler for each `oauth2` flow?

### DIFF
--- a/connexion/middleware/security.py
+++ b/connexion/middleware/security.py
@@ -97,7 +97,6 @@ class SecurityOperation:
 
         request = ASGIRequest(scope)
         await self.verification_fn(request)
-        # TODO: Adjust scope? Or adjust downstream validation to account for api keys in query?
         await self.next_app(scope, receive, send)
 
 
--- a/connexion/middleware/security.py
+++ b/connexion/middleware/security.py
@@ -51,7 +51,7 @@ class SecurityOperation:
         auth_funcs = []
         for security_req in self.security:
             if not security_req:
-                auth_funcs.append(self.security_handler_factory.verify_none())
+                auth_funcs.append(self.security_handler_factory.verify_none)
                 continue
 
             sec_req_funcs = {}
@@ -67,123 +67,15 @@ class SecurityOperation:
                         )
                         break
                     oauth = True
-                    token_info_func = self.security_handler_factory.get_tokeninfo_func(
-                        security_scheme
-                    )
-                    scope_validate_func = (
-                        self.security_handler_factory.get_scope_validate_func(
-                            security_scheme
-                        )
-                    )
-                    if not token_info_func:
-                        logger.warning("... x-tokenInfoFunc missing", extra=vars(self))
-                        break
 
-                    sec_req_funcs[
-                        scheme_name
-                    ] = self.security_handler_factory.verify_oauth(
-                        token_info_func, scope_validate_func, required_scopes
-                    )
-
-                # Swagger 2.0
-                elif security_scheme["type"] == "basic":
-                    basic_info_func = self.security_handler_factory.get_basicinfo_func(
-                        security_scheme
-                    )
-                    if not basic_info_func:
-                        logger.warning("... x-basicInfoFunc missing", extra=vars(self))
-                        break
-
-                    sec_req_funcs[
-                        scheme_name
-                    ] = self.security_handler_factory.verify_basic(basic_info_func)
-
-                # OpenAPI 3.0.0
-                elif security_scheme["type"] == "http":
-                    scheme = security_scheme["scheme"].lower()
-                    if scheme == "basic":
-                        basic_info_func = (
-                            self.security_handler_factory.get_basicinfo_func(
-                                security_scheme
-                            )
-                        )
-                        if not basic_info_func:
-                            logger.warning(
-                                "... x-basicInfoFunc missing", extra=vars(self)
-                            )
-                            break
-
-                        sec_req_funcs[
-                            scheme_name
-                        ] = self.security_handler_factory.verify_basic(basic_info_func)
-                    elif scheme == "bearer":
-                        bearer_info_func = (
-                            self.security_handler_factory.get_bearerinfo_func(
-                                security_scheme
-                            )
-                        )
-                        if not bearer_info_func:
-                            logger.warning(
-                                "... x-bearerInfoFunc missing", extra=vars(self)
-                            )
-                            break
-                        sec_req_funcs[
-                            scheme_name
-                        ] = self.security_handler_factory.verify_bearer(
-                            bearer_info_func
-                        )
-                    else:
-                        logger.warning(
-                            "... Unsupported http authorization scheme %s" % scheme,
-                            extra=vars(self),
-                        )
-                        break
-
-                elif security_scheme["type"] == "apiKey":
-                    scheme = security_scheme.get("x-authentication-scheme", "").lower()
-                    if scheme == "bearer":
-                        bearer_info_func = (
-                            self.security_handler_factory.get_bearerinfo_func(
-                                security_scheme
-                            )
-                        )
-                        if not bearer_info_func:
-                            logger.warning(
-                                "... x-bearerInfoFunc missing", extra=vars(self)
-                            )
-                            break
-                        sec_req_funcs[
-                            scheme_name
-                        ] = self.security_handler_factory.verify_bearer(
-                            bearer_info_func
-                        )
-                    else:
-                        apikey_info_func = (
-                            self.security_handler_factory.get_apikeyinfo_func(
-                                security_scheme
-                            )
-                        )
-                        if not apikey_info_func:
-                            logger.warning(
-                                "... x-apikeyInfoFunc missing", extra=vars(self)
-                            )
-                            break
-
-                        sec_req_funcs[
-                            scheme_name
-                        ] = self.security_handler_factory.verify_api_key(
-                            apikey_info_func,
-                            security_scheme["in"],
-                            security_scheme["name"],
-                        )
-
-                else:
-                    logger.warning(
-                        "... Unsupported security scheme type %s"
-                        % security_scheme["type"],
-                        extra=vars(self),
-                    )
+                sec_req_func = self.security_handler_factory.parse_security_scheme(
+                    security_scheme, required_scopes
+                )
+                if sec_req_func is None:
                     break
+
+                sec_req_funcs[scheme_name] = sec_req_func
+
             else:
                 # No break encountered: no missing funcs
                 if len(sec_req_funcs) == 1:
@@ -199,16 +91,23 @@ class SecurityOperation:
         return self.security_handler_factory.verify_security(auth_funcs)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if not self.security:
+            await self.next_app(scope, receive, send)
+            return
+
         request = ASGIRequest(scope)
         await self.verification_fn(request)
+        # TODO: Adjust scope? Or adjust downstream validation to account for api keys in query?
         await self.next_app(scope, receive, send)
 
 
 class SecurityAPI(RoutedAPI[SecurityOperation]):
-    def __init__(self, *args, auth_all_paths: bool = False, **kwargs):
+    def __init__(
+        self, *args, auth_all_paths: bool = False, security_map: dict = None, **kwargs
+    ):
         super().__init__(*args, **kwargs)
 
-        self.security_handler_factory = SecurityHandlerFactory()
+        self.security_handler_factory = SecurityHandlerFactory(security_map)
 
         if auth_all_paths:
             self.add_auth_on_not_found()
--- a/connexion/security.py
+++ b/connexion/security.py
@@ -1,6 +1,46 @@
 """
-This module defines an abstract SecurityHandlerFactory which supports the creation of security
-handlers for operations.
+This module defines a SecurityHandlerFactory which supports the creation of
+SecurityHandler instances for different security schemes.
+
+It also exposes a `SECURITY_HANDLERS` dictionary which maps security scheme
+types to SecurityHandler classes. This dictionary can be used to register
+custom SecurityHandler classes for custom security schemes, or to overwrite
+existing SecurityHandler classes.
+This can be done by supplying a value for `security_map` argument of the
+SecurityHandlerFactory.
+
+Swagger 2.0 lets you define the following authentication types for an API:
+
+- Basic authentication
+- API key (as a header or a query string parameter)
+- OAuth 2 common flows (authorization code, implicit, resource owner password credentials, client credentials)
+
+
+Changes from OpenAPI 2.0 to OpenAPI 3.0
+If you used OpenAPI 2.0 before, here is a summary of changes to help you get started with OpenAPI 3.0:
+- securityDefinitions were renamed to securitySchemes and moved inside components.
+- type: basic was replaced with type: http and scheme: basic.
+- The new type: http is an umbrella type for all HTTP security schemes, including Basic, Bearer and other,
+and the scheme keyword indicates the scheme type.
+- API keys can now be sent in: cookie.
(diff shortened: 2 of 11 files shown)


connexion/security.py
```diff
@@ -25,6 +59,21 @@
 logger = logging.getLogger("connexion.api.security")
 
 
+@dataclass
```
Ruwann, line 62: Not used currently, was exploring the option to create classes for the security schemes instead of regular dicts.

connexion/security.py
```diff
@@ -45,6 +94,87 @@ class SecurityHandlerFactory:
     context_kw = "context_"
     client = None
 
+    # TODO: Extract into separate class? (That has a security_handler_factory attribute?)
+    def parse_security_scheme(
```
Ruwann, line 98: Idea is that the SecurityHandlerFactory can get subclasses and this function overwritten:
```python
def parse_security_scheme(...):
    if security_scheme["type"] == "oidc":
        # oidc implementation
    return super().parse_security_scheme(...)
```

Ruwann (commented):
The image should also be deleted

connexion/security.py
RobbeSneyders: I was thinking more in the lines of having a separate class for each type like we have with the validators. Something like:

```
security_handler_dict = {
    "oauth2": Oauth2Handler,
    "basic": BasicHandler,
    ...
}

def parse_security_scheme(...):
    handler = security_handler_dict.get("security_scheme["type"])
    return handler.get_fn(security_scheme, required_scopes)
```

and then move the `verify_<type>` functions from this class into those handlers as well.

This splits the functionality nicely per type and allows people to contribute their handlers back to connexion.

Wdyt?

connexion/security.py
```diff
@@ -22,229 +55,269 @@
...
+        )
 
-    def verify_api_key(self, api_key_info_func, loc, name):
+    # TODO: Move logic from here to `get_fn`?
+    #   that has the entire security definition, so should determine which function?
+    # All API key checking funcs have signature func(api_key, required_scopes) -> result
+    def _get_verify_func(self, api_key_info_func, loc, name):
         check_api_key_func = self.check_api_key(api_key_info_func)
 
         def wrapper(request):
```
RobbeSneyders, line 323: Can't add a comment on the specific line below, but the API key used to be removed from the query string before passing it to the view function. I think this is necessary to prevent an unexpected keyword argument error. If we want to achieve the same behaviour, we should ensure it is removed from the scope.

connexion/security.py
```diff
@@ -403,6 +523,126 @@ async def wrapper(request, token, required_scopes):
 
         return wrapper
 
+
+class OIDCSecurityHandlerFactory(AbstractSecurityHandlerFactory):
```
RobbeSneyders, line 527: Let's leave this one out of this PR. We can do it separately.

connexion/security.py
```diff
@@ -403,6 +523,126 @@ async def wrapper(request, token, required_scopes):
...
+    """
+    SecurityHandlerFactory for the OIDC security scheme.
+    """
+
+    # TODO
+
+
+SECURITY_HANDLERS = {
```
RobbeSneyders, line 535: Just to confirm if I understand this. If I want to add a custom `SecurityHandlerFactory` for an http scheme with "digest" type, I just add `"digest": DigestSecurityHandlerFactory` as a top level key here, right?

connexion/security.py
```diff
@@ -403,6 +523,126 @@ async def wrapper(request, token, required_scopes):
...
+    "openIdConnect": OIDCSecurityHandlerFactory,
+    # OpenAPI 3: http schemes
+    "bearer": BearerSecurityHandlerFactory,
+}
+
+
+# TODO: Split up
+#   along security schemes
+#   along 'functionality' -> get_*_func (resolver), security_{deny,passthrough}, and verify_*
+class SecurityHandlerFactory(StandardHandlerFactory):
```
RobbeSneyders, line 552: Seems like we're now calling two distinct things `SecurityHandlerFactory`, this class, and the classes it calls. We should probably rename one of them for clarity.

connexion/security.py
```diff
@@ -403,6 +523,126 @@ async def wrapper(request, token, required_scopes):
...
+
+    no_value = AbstractSecurityHandlerFactory.no_value  # TODO
+    required_scopes_kw = "required_scopes"
+    context_kw = "context_"
+    client = None
+
+    def __init__(self, security_handlers: t.Optional[dict] = None) -> None:
+        super().__init__()
+        self.security_handlers = SECURITY_HANDLERS.copy()
+        if security_handlers is not None:
```
RobbeSneyders, line 573: We still need to expose this to the user. Currently the `SecurityHandlerFactory` is instantiated without passing any options at https://github.com/spec-first/connexion/pull/1671/files#diff-a1af7a3afc2ba04857107ee0179f4de7486723ef00531fa1a3685bcf491e2a78R103.

RobbeSneyders (commented):
This is starting to look great @Ruwann!

I left some comments already.

connexion/security.py
```diff
@@ -22,229 +55,269 @@
...
-            check helpers (used outside wrappers): _need_to_add_context_or_scopes
-            the security function
+    @staticmethod
+    async def security_passthrough(request):
+        return request
 
-        verify helpers (used inside wrappers): get_auth_header_value, get_cookie_value
-    """
+    @staticmethod
+    def security_deny(function):
```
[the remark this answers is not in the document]
Ruwann: Good question, I actually noticed today that this function isn't used anymore 😓 Seems like this was used when connexion didn't support multiple schemes in AND fashion anymore. We can delete this function now.

connexion/security.py
```diff
@@ -22,229 +55,269 @@
...
+        """
+
+        def deny(*args, **kwargs):
+            raise ConnexionException("Error in security definitions")
+
+        return deny
 
-    no_value = object()
+    @staticmethod
+    def verify_none():
```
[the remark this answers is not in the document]
Ruwann: This is used when optional security is defined, such as:

```yaml
security:
  - {}
  - myapikey: []
```

https://github.com/OAI/OpenAPI-Specification/issues/14#issuecomment-297457320
https://github.com/microsoft/OpenAPI.NET/issues/521#issuecomment-753003830

connexion/security.py
```diff
@@ -22,229 +55,269 @@
...
-        if token_info_url:
-            return self.get_token_info_remote(token_info_url)
+        try:
+            auth_type, value = authorization.split(None, 1)
+        except ValueError:
+            raise OAuthProblem(detail="Invalid authorization header")
+        return auth_type.lower(), value
 
-        return None
+    def _need_to_add_context_or_scopes(self, func):
```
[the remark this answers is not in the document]
Ruwann: For the context, it would make sense to keep only one way of accessing it instead of 2 ways to do the same thing.

Not sure about the required scopes:
in the past, only oauth and oidc were allowed to use scopes:
> For other security scheme types, the array MUST be empty.

but in OpenAPI 3.1, all security schemes can now use the scopes:
> For other security scheme types, the array MAY contain a list of role names which are required for the execution, but are not otherwise defined or exchanged in-band.

So it would depend on the signature of the security function again whether scopes need to be passed.

connexion/security.py
```diff
@@ -22,229 +55,269 @@
...
+        arguments, has_kwargs = inspect_function_arguments(func)
+        need_context = self.context_kw in arguments
+        need_required_scopes = has_kwargs or self.required_scopes_kw in arguments
+        return need_context, need_required_scopes
 
-    @classmethod
-    def get_scope_validate_func(cls, security_definition):
+    def _resolve_func(self, security_scheme):
```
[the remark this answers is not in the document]
Ruwann: The idea was that there are 2 main steps in each class (similar to the previous implementation):
1. from security scheme to a Python function
2. transform/wrap Python function such that it gets the right information is supplied when it is called

connexion/security.py
Ruwann: Good point, let's keep this one as `SecurityHandlerFactory` and rename the other ones.

connexion/security.py
Ruwann: Yes, indeed. Not entirely convinced that this is the best approach, but seemed to be the best starting point.

Happy to hear if you have other ideas :)

connexion/security.py
Ruwann: Noticed that it is still subclassing from the `StandardHandlerFactory` - have to rework this as well. Not sure if these standard functions need to be able to be overwritten/customized.

connexion/security.py
Ruwann: Very good remark, didn't think of that. I suspect this mean that we don't have a test for this now then, otherwise this test should be failing, no?

connexion/security.py
RobbeSneyders: As long as we don't expect any conflicts, I think this is fine. I had a quick look, and as far as I can tell there are no conflicts within the standards officially supported by OpenAPI.

connexion/security.py
RobbeSneyders: I assume a test is missing indeed. Something to check before merging the PR.

Currently the scope is not available in the `SecurityHandler`s though, which complicates modifying the scope. We could solve this by making the `SecurityHandler`s ASGI callables, which could have some additional nice benefits. 
- It would fit with the "everything is an ASGI callable" design
- It might remove some of the functional complexity we currently have, as the `SecurityHandler` would no longer need to return a function, but can just be called itself

connexion/security.py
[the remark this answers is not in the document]
RobbeSneyders: Yes, the split definitely makes sense, but without the `get_fn` in the base class, it just imposes a structure for the internal functions which seems a bit arbitrary.

connexion/lifecycle.py
```diff
@@ -28,6 +28,10 @@ def content_type(self) -> str:
     def mimetype(self) -> str:
         raise NotImplementedError
 
+    @property
+    def headers(self) -> t.Any:
```
Ruwann, line 32: Only noticed the `__getattr__` function now, seems like the `spec_set` argument of `MagicMock` isn't able to properly detect this, leading to an error in the tests

connexion/security.py
```diff
@@ -1,182 +1,123 @@
 """
 This module defines an abstract SecurityHandlerFactory which supports the creation of security
```
RobbeSneyders, line 2: Also defines SecurityHandlers now

connexion/security.py
```diff
@@ -403,6 +431,148 @@ async def wrapper(request, token, required_scopes):
 
         return wrapper
 
+
+class OIDCSecurityHandler(AbstractSecurityHandler):
```
RobbeSneyders, line 435: Let's remove this since it's not implemented.

connexion/security.py
```diff
@@ -403,6 +431,148 @@ async def wrapper(request, token, required_scopes):
...
+
+    verify_* -> returns a security wrapper around the security function
+        check_* -> returns a function tasked with doing auth for use inside the verify wrapper
+            check helpers (used outside wrappers): _need_to_add_context_or_scopes
+            the security function
+
+        verify helpers (used inside wrappers): get_auth_header_value, get_cookie_value
+    """
+
+    no_value = AbstractSecurityHandler.no_value  # TODO
```
RobbeSneyders, line 473: Maybe better to define this on a module level since we're accessing it from different classes?

connexion/security.py
```diff
@@ -403,6 +431,148 @@ async def wrapper(request, token, required_scopes):
...
+        required_scopes: t.List[str],
+    ) -> t.Optional[t.Callable]:
+        """Parses the security scheme and returns the function for verifying it.
+
+        :param security_scheme: The security scheme from the spec.
+        :param required_scopes: List of scopes for this security scheme.
+        """
+        security_type = security_scheme["type"]
+        if security_type in ("basic", "oauth2"):
+            security_factory = self.security_handlers[security_type]
```
RobbeSneyders, line 495: Some inconsistent naming here.

connexion/security.py
```diff
@@ -403,6 +431,148 @@ async def wrapper(request, token, required_scopes):
...
+    "apiKey": ApiKeySecurityHandler,
+    "oauth2": OAuthSecurityHandler,
+    # OpenAPI 3 only
+    "openIdConnect": OIDCSecurityHandler,
+    # OpenAPI 3: http schemes
+    "bearer": BearerSecurityHandler,
+}
+
+
+# TODO: Split up
```
RobbeSneyders, line 457: This Todo is outdated.

connexion/security.py
```diff
@@ -403,6 +431,148 @@ async def wrapper(request, token, required_scopes):
...
+#   along security schemes
+#   along 'functionality' -> get_*_func (resolver), security_{deny,passthrough}, and verify_*
+class SecurityHandlerFactory:
+    """
+    get_*_func -> _get_function -> get_function_from_name (name=security function defined in spec)
```
RobbeSneyders, line 462: I think this documentation is outdated.

connexion/security.py
```diff
@@ -228,44 +188,68 @@ def wrapper(request):
...
-                and a new immutable dict without the popped key.
-                """
-                cls = type(_dict)
-                try:
-                    _dict = _dict.to_dict(flat=False)
-                    return _dict.pop(key)[0], cls(_dict)
-                except AttributeError:
-                    _dict = dict(_dict.items())
-                    return _dict.pop(key), cls(_dict)
+    # TODO: Move logic from here to `get_fn`?
```
RobbeSneyders, line 244: Is this TODO stil valid?

connexion/security.py
```diff
@@ -192,24 +133,43 @@ def get_auth_header_value(request):
...
+        Wraps the security handler function
+        """
+        return self._generic_check(function, "Provided authorization is not valid")
 
-        return wrapper
 
-    def verify_basic(self, basic_info_func):
+class BasicSecurityHandler(AbstractSecurityHandler):
+    """
+    Factory for security for
```
RobbeSneyders, line 164: Docstrings mention factory a couple of times while the name has been changed.

RobbeSneyders (commented):
Looks great! Seems like you have cracked the nut :)
I left some smaller comments in-line.

On a higher level,
- I think the code would still benefit from making the handlers callable, even if they are not ASGI callable. There's currently still a lot of functions being passed around, which makes it harder to follow.
- It would be good to add a test for the `security_map` keyword argument.

security.excalidraw.png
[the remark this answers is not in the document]
Ruwann: This can be deleted indeed, don't think it adds a lot of value.

connexion/middleware/security.py
```diff
@@ -199,16 +91,26 @@ def _get_verification_fn(self):
         return self.security_handler_factory.verify_security(auth_funcs)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
```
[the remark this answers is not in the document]
Ruwann: Yes, good point, it's not necessary here!

connexion/validators/parameter.py
```diff
@@ -102,6 +112,7 @@ def validate_request(self, request):
 
         if self.strict_validation:
             query_errors = self.validate_query_parameter_list(request)
+            query_errors.difference_update(self.security_query_params)
```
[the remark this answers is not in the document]
Ruwann: That makes it cleaner indeed, thanks for the input!

RobbeSneyders (approved):
Thanks!

This looks good to me except for
- Extending the interface of the Apps and APIs
- Adding a test for the validator_map
- Possibly making the security handlers callable

For me it's fine to merge this PR and tackle those in separate PRs though, which will make it easier to review.

Ruwann:
I added the `security_map` argument for Apps and APIs in this PRs. The other ones can indeed be done in separate PRs, then we can merge this one (finally :))

PR #1673, opened 2023-03-08
Pass through lifespan events

### LINKED ISSUE #1672: Connexion 3.0.0a2, Starlette, on_startup and on_shutdown
### Description
Using a Starlette app with Connexion 3, startup and shutdown handlers are not called. 

I understand that Connexion uses/wraps Starlette Router, but the question remains; how to implement startup and shutdown handlers. As well, if we want to use the vanilla `connexion.AsyncApp`, what's the recommended way of adding handlers for lifecycle events?

### Expected behaviour

```
def on_startup():
    logging.info("Starting up")
def on_shutdown():
    logging.info("Shutting down")
app = Starlette(on_startup=[on_startup], on_shutdown=[on_shutdown])
app = connexion.ConnexionMiddleware(app)
app.run(f"{Path(__file__).stem}:app", host="0.0.0.0", port=8082, workers=1, reload=False)

```
The on_startup and on_shutdown handlers should be called as part of server lifecycle.

### Actual behaviour

The on_startup and on_shutdown are ignored/never called.

### Steps to reproduce
Shown above.

### Additional info:

Output of the commands:

Python 3.11.2
Connexion 3.0.0a2

RobbeSneyders:
Fixes #1672

Lifespan events were being intercepted by the swagger UI middleware. We should let them pass through.

### DIFF
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 import pytest
 from connexion.middleware import ConnexionMiddleware
 from starlette.datastructures import MutableHeaders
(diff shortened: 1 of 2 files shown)


tests/test_middleware.py
```diff
@@ -46,3 +48,13 @@ def test_routing_middleware(middleware_app):
     assert (
         response.headers.get("operation_id") == "fakeapi.hello.post_greeting"
     ), response.status_code
+
+
+async def test_lifecycle():
```
Ruwann, line 53: This test fails on python 3.7 because `AsyncMock` is only availabe from Python 3.8.
We skip other tests relying on AsyncMock for Python 3.7

PR #1678, opened 2023-03-28
Fix stream replay in validators

RobbeSneyders:
The current implementation of replaying the stream will always replay the first message. This PR fixes this by progressing through the messages with each call.

### DIFF
--- a/connexion/validators/abstract.py
+++ b/connexion/validators/abstract.py
@@ -99,11 +99,14 @@ class AbstractRequestBodyValidator:
         receive: Receive, *, messages: t.Iterable[t.MutableMapping[str, t.Any]]
     ) -> Receive:
         """Insert messages at the start of the `receive` channel."""
+        # Ensure that messages in an iterator
+        messages = iter(messages)
 
         async def receive_() -> t.MutableMapping[str, t.Any]:
-            for message in messages:
-                return message
-            return await receive()
+            try:
+                return next(iter(messages))
+            except StopIteration:
+                return await receive()
 
         return receive_
 


RobbeSneyders:
cc: @julienschuermans

connexion/validators/abstract.py
```diff
@@ -99,11 +99,14 @@ def _insert_messages(
         receive: Receive, *, messages: t.Iterable[t.MutableMapping[str, t.Any]]
     ) -> Receive:
         """Insert messages at the start of the `receive` channel."""
+        # Ensure that messages in an iterator
```
Ruwann, line 102:
```suggestion
        # Ensure that messages is an iterator
```
😇

connexion/validators/abstract.py
```diff
@@ -99,11 +99,14 @@ def _insert_messages(
...
+        messages = iter(messages)
 
         async def receive_() -> t.MutableMapping[str, t.Any]:
-            for message in messages:
-                return message
-            return await receive()
+            try:
+                return next(iter(messages))
```
Ruwann, line 107: There's double `iter()` call (also done on line 103), here we can just do `next(messages)` I believe

tests/test_validation.py
```diff
@@ -140,3 +140,29 @@ def test_parameter_validator(monkeypatch):
...
+    )
+
+    replay = []
+    more_body = True
+    while more_body:
+        message = await wrapped_receive()
+        replay.append(message)
+        more_body = message.get("more_body", False)
+
+        if len(replay) > len(messages):
```
Ruwann, line 165: Why is this check needed?

Ruwann (changes requested):

tests/test_validation.py
RobbeSneyders: If the first message is always returned as before, we get stuck in this while loop. Updated this to an assert statement with some more info.

connexion/validators/abstract.py
RobbeSneyders: Denied :stop_sign:

Ruwann (approved):
LGTM, one minor comment on the assertion message

tests/test_validation.py
```diff
@@ -140,3 +140,30 @@ def test_parameter_validator(monkeypatch):
...
+
+    replay = []
+    more_body = True
+    while more_body:
+        message = await wrapped_receive()
+        replay.append(message)
+        more_body = message.get("more_body", False)
+
+        assert len(replay) <= len(messages), (
+            "Replayed more messages than received, " "break out of while loop"
```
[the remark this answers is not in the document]
RobbeSneyders: The solution is ugly though :vomiting_face:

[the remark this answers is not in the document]
Ruwann: yikes

PR #1681, opened 2023-04-08
Add tests for strict validation in combination with api key in query

Ruwann:
Continues on the work of #1077 by moving the test fixtures into the `secure_endpoint` fixture and adding a test for it in `test_secure_api.py`.

The test will check whether an api key in the query will not lead to an error when `strict_validation` is enabled.

### DIFF
--- a/tests/api/conftest.py
+++ b/tests/api/conftest.py
@@ -115,6 +115,17 @@ def secure_endpoint_app(spec, app_class):
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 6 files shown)


RobbeSneyders (approved):
Nice, glad to see some things move out of the simple app :)

Was this fixture just not used before?

Ruwann:
Exactly, almost everything was there (fixture, swagger, view function) except an actual test and app with `strict_validation` enabled

PR #1686, opened 2023-04-13
Inspect middleware parameters before passing in lifespan

### LINKED ISSUE #1682: Connexion 3.0.0a4, GZIP Middleware
### Description

In 3.0.0a2 I was able to use the Starlette GZipMiddleware class but in 3.0.0a4 I get an exception at runtime.

### Expected behaviour

I would hope to be able to use Starlette middleware in my app's middleware stack.

### Actual behaviour

Using 
```
middlewares = [GZipMiddleware] + ConnexionMiddleware.default_middlewares
super().__init__(__name__,
                 lifespan=self.lifespan,
                 middlewares=middlewares,
                 specification_dir="spec/",
                 resolver=RelativeResolver(root_path="app.controllers"),
                 *args, **kwargs)

```
in my application constructor. But in 3.0.0a4 I get TypeError: GZipMiddleware.__init__() got an unexpected keyword argument 'lifespan'

It's clear that this is a result of the lifespan changes in the previous release. 

Am I doing something wrong here?

### Steps to reproduce

As above

### Additional info:

Output of the commands:

- `python --version`

RobbeSneyders:
Fixes #1682 

This PR inspects the parameters of a Middleware class before passing in the `lifespan` keyword argument. I was doubting to check the class instead (`isinstance(LifespanMiddleware)`), but that's less flexible if users want to write custom middleware to handle the lifespan.

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -5,7 +5,6 @@ import abc
 import asyncio
 import builtins
 import functools
-import inspect
 import keyword
 import logging
 import re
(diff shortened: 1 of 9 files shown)


Ruwann (approved):
LGTM. I agree that this approach looks more extensible than checking the class.

PR #1690, opened 2023-04-21
Remove context keyword from security handlers

Ruwann:
Follow-up on #1671 

Since the request context is available globally, we can remove this complexity from the security handlers ("There should be one-- and preferably only one --obvious way to do it.")

In addition it seems like there weren't any tests for the context in security handler functions, and I don't think there's a lot of value in a test for checking the context in security functions specifically.

### DIFF
--- a/connexion/security.py
+++ b/connexion/security.py
@@ -67,7 +67,6 @@ NO_VALUE = object()
 class AbstractSecurityHandler:
 
     required_scopes_kw = "required_scopes"
-    context_kw = "context_"
     client = None
     security_definition_key: str
     """The key which contains the value for the function name to resolve."""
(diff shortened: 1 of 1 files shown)


RobbeSneyders (approved):
Thx @Ruwann!
So should we do this for the view functions as well?

Ruwann:
I think it would indeed make sense to also do it for the view functions as well. I haven't found the reasoning for adding it but the same argument applies: one way of accessing the context instead of two.

For the security handlers, the context argument was added in #869 but not really clear what the reason was - I assume it was because it was the easiest (?) way of making the context available for `aiohttp` in security handler functions?

PR #1691, opened 2023-04-21
Add test for security_map argument

Ruwann:
Follow-up of #1671 

Let me know if there is an easier way doing the test instead of the current "custom" basic auth.


RobbeSneyders (approved):
Thanks @Ruwann. This looks good to me. Seems like there's a merge conflict though.

PR #1695, opened 2023-04-24
Unpack error handler when registering on middleware

RobbeSneyders:
Registering error handlers on the async app leads to an error on 3.0.0a5 because the error handlers are not unpacked correctly when being registered internally.


Ruwann (approved):
LGTM, thanks for adding a test!

PR #1704, opened 2023-05-02
Starlette had new releases.

marcin-lulek-cint:
Fixes # .
I want to use connexion with newer Starlette

 - Loosen requirements on starlette version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ Jinja2 = "^3.0.0"
 python-multipart = "~0.0.5"
 PyYAML = ">= 5.1, < 7"
 requests = "^2.27"
-starlette = "^0.25"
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Thanks @marcin-lulek-cint!

The Poetry packaging doesn't work as I expected. It's recommended that we [switch to inequality versioning instead of caret versioning](https://github.com/orgs/python-poetry/discussions/3757#discussioncomment-435337).

I added a commit to change this for all dependencies.

RobbeSneyders:
I had to revert the python version to caret notation since installing otherwise fails because `py-swagger-ui` does define an upper bound for the python version via caret notation. We'll have to update it there first.

Ruwann (approved):
Thanks, LGTM!

PR #1708, opened 2023-06-01
Add traceback info to common error handler

Ruwann:
Still up for discussion: right now, when an internal server error occurs, the logged error is not really useful for the user.
For example, a `KeyError` will only print the missing key instead of the full stacktrace and error.

Adding it to the common error handler as that one is meant to catch all remaining errors, `HTTPException`s are handled by another error handler.

Another option would be to do like `Flask` and `Starlette` by adding a `debug` option, but that seems a lot of work while this current solution doesn't really have big downsides (?)

### DIFF
--- a/connexion/middleware/exceptions.py
+++ b/connexion/middleware/exceptions.py
@@ -52,7 +52,7 @@ class ExceptionMiddleware(StarletteExceptionMiddleware):
 
     @staticmethod
     def common_error_handler(_request: StarletteRequest, exc: Exception) -> Response:
-        logger.error(exc)
+        logger.error(exc, exc_info=exc)
 
         response = InternalServerError().to_problem()
 


connexion/middleware/exceptions.py
```diff
@@ -52,7 +52,7 @@ def http_exception(_request: StarletteRequest, exc: HTTPException) -> Response:
 
     @staticmethod
     def common_error_handler(_request: StarletteRequest, exc: Exception) -> Response:
-        logger.error(exc)
+        logger.error(exc, exc_info=exc)
```
RobbeSneyders, line 55: I guess this is equivalent to `logger.exception(exc)`?

RobbeSneyders (approved):
Yes, this was bugging me already :)

connexion/middleware/exceptions.py
Ruwann: Not entirely, it would be the same as `logger.exception(exc, exc_info=exc)`

Source code:
```python
    def exception(self, msg, *args, exc_info=True, **kwargs):
        """
        Convenience method for logging an ERROR with exception information.
        """
        self.error(msg, *args, exc_info=exc_info, **kwargs)
```

But will not work properly as were not in an exception handler scope here

julienschuermans:
> `logger.error(exc, exc_info=exc)`
This was much needed, thanks!

@Ruwann @RobbeSneyders I think it would it make sense to apply this same change to the `http_exception` and `problem_handler` functions in this file because neither `connexion.middleware.exceptions.ProblemException` nor `starlette.exceptions.HTTPException` have a decent `__str__` implementation.

two other ideas - in case dumping the exc_info is overkill for each HTTPException or ProblemException:
- `logger.error(repr(exc))`  would be useful.
- OR just implement the `__str__` method for those Exception subclasses

PR #1712, opened 2023-06-12
Feature/get delete body

### LINKED ISSUE #1689: Body in GET/DELETE requests in 3.x
### Description
Hello,
Since version 2.x connexion does not support body in GET and DELETE requests. 
https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/connexion/operations/abstract.py#L283

However, many APIs are designed this way and connexion is being too restrictive to API implementations which causes problems with further library updates.

### Expected behaviour
IMHO it is best for the library to be as flexible as possible while at the same time promoting correct API design. But currently, it is too restrictive and blocks the projects that have been developed for years to upgrade to a higher version of connexion.

Even the authors of OpenAPI understand that this change is too restrictive

> In OpenAPI 3.1, we decided to go back on this change for the main reason that indeed many APIs are designed this way (this is not to say that we believe it's the right approach).

https://github.com/swagger-api/swagger-ui/issues/2136#issuecomment-1055839700

I think it is best if connexion 3.x also reverts to this behavior in order to be more flexible and allow easier transition for projects.

### Actual behaviour
Connexion from version 2.x does not automatically parse body in arguments and breaks all endpoints that currently utilize this functionality. I know about the fix of using `connexion.request.json` however this requires changes of many APIs especially of old APIs that have been developed for years.

### Additional info:

Output of the commands:

- 3.11
- 2.14.2

Ruwann:
Fixes #1689 

 - Pass through request body in case of `GET` and `DELETE` requests

### DIFF
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -376,6 +376,9 @@ def _get_body_argument(
     if len(arguments) <= 0 and not has_kwargs:
         return {}
 
+    if operation.request_body == {}:
+        return {}
+
     body_name = sanitize(operation.body_name(content_type))
 
     if content_type in FORM_CONTENT_TYPES:
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -376,7 +376,7 @@ def _get_body_argument(
     if len(arguments) <= 0 and not has_kwargs:
         return {}
 
-    if operation.request_body == {}:
+    if not operation.is_request_body_defined:
         return {}
 
     body_name = sanitize(operation.body_name(content_type))
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -211,6 +211,10 @@ def get_arguments(
         )
     )
 
+    if operation.method.upper() == "TRACE":
+        # TRACE requests MUST NOT include a body (RFC7231 section 4.3.8)
+        return ret
+
     ret.update(
         _get_body_argument(
             body,
--- a/connexion/decorators/parameter.py
+++ b/connexion/decorators/parameter.py
@@ -211,18 +211,17 @@ def get_arguments(
         )
     )
 
-    if operation.method.upper() in ["PATCH", "POST", "PUT"]:
-        ret.update(
-            _get_body_argument(
-                body,
-                operation=operation,
-                arguments=arguments,
-                has_kwargs=has_kwargs,
-                sanitize=sanitize,
-                content_type=content_type,
-            )
+    ret.update(
+        _get_body_argument(
+            body,
+            operation=operation,
+            arguments=arguments,
+            has_kwargs=has_kwargs,
+            sanitize=sanitize,
+            content_type=content_type,
         )
-        ret.update(_get_file_arguments(files, arguments, has_kwargs))
+    )
+    ret.update(_get_file_arguments(files, arguments, has_kwargs))
     return ret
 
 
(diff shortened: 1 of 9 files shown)


tests/api/test_parameters.py
```diff
@@ -456,11 +456,11 @@ def test_args_kwargs(simple_app):
     app_client = simple_app.test_client()
     resp = app_client.get("/v1.0/query-params-as-kwargs")
     assert resp.status_code == 200
-    assert resp.json() == {}
+    assert resp.json() == {"body": {}}
 
     resp = app_client.get("/v1.0/query-params-as-kwargs?foo=a&bar=b")
     assert resp.status_code == 200
-    assert resp.json() == {"foo": "a"}
+    assert resp.json() == {"foo": "a", "body": {}}
```
Ruwann, line 463: Not 100% sure about changing these tests this way.
I believe these view functions use `**kwargs`, so the `body` gets passed in here - so in that sense it seems logical. On the other hand, if the endpoint doesn't define a body, it's weird that it will be put in the `kwargs` (although I would think it currently also happens for `POST` methods and the like)

RobbeSneyders: Agree that it's not ideal. For `POST` this is probably more expected than for `GET`. I'm wondering if we can add a check so we only pass in the body if it's present, but I believe it's hard to discern between no body and an empty body. So then we'd probably have to check if the spec defines one.

Ruwann: I am now checking whether the operation has defined a request body, and only pass the body to the view function in case it has.

connexion/decorators/parameter.py
```diff
@@ -377,6 +376,9 @@ def _get_body_argument(
     if len(arguments) <= 0 and not has_kwargs:
         return {}
 
+    if operation.request_body == {}:
```
RobbeSneyders, line 379: Can you add a comment with some explanation here? It wasn't immediately clear to me that `operation.request_body == {}` was checking if a request body was defined in the specification.

Another option is that we add a boolean `reauest_body_defined` property to the `Operation` class which wraps this check, so it's more readable where it's used.

RobbeSneyders (commented):
Thanks @Ruwann!

Should we add a specific test for this behavior? Now it's only implicit in existing tests and we don't test for the GET and similar methods.

Also, I guess this can be a breaking change for users that handled the passed in empty body, so we might want to add this to the smaller breaking changes in the changelog.

RobbeSneyders:
Can you have a look at my comments above @Ruwann? :slightly_smiling_face:

Ruwann:
Thanks for the comments, I think I got them all :)

docs/v3.rst
```diff
@@ -126,6 +126,7 @@ Smaller breaking changes
 * The ``MethodViewResolver`` has been renamed to ``MethodResolver``, and a new ``MethodViewResolver``
   has been added to work with Flask's ``MethodView`` specifically.
 * Built-in support for uWSGI has been removed. You can re-add this functionality using a custom middleware.
+* The request body is now passed through for ``PATCH``, ``POST``, and ``PUT`` methods as well.
```
RobbeSneyders, line 129:
```suggestion
* The request body is now passed through for ``GET``, ``HEAD``, ``DELETE``, ``CONNECT`` and ``OPTIONS`` methods as well.
```

RobbeSneyders (commented):
Thanks Ruwan!

I had another look at the [HTTP spec](https://www.rfc-editor.org/rfc/rfc7231#section-4.3), and seems like there is one method which MUST NOT include a request body, which is the TRACE method. Maybe we should still exclude it.

Ruwann:
Thanks, I indeed didn't spot that one - nice catch. Had another look and I also think that's the only one, I'll update the PR accordingly!

docs/v3.rst
Ruwann: Small oopsie

RobbeSneyders (approved):

Issue #1709, opened 2023-06-06
API with base path not reachable when registered after API without base path.


RobbeSneyders:
### Description

When registering multiple APIS, where the base path of a later API contains the base path of the earlier API, the later API is not accessible. Eg:

```python
app = connexion.FlaskApp(__name__, specification_dir="spec/")
app.add_api("openapi.yaml", base_path="/")
app.add_api("swagger.yaml", base_path="/swagger")
```

We can fix this by sorting the APIs on their base path before registering them, which we do [here](https://github.com/spec-first/connexion/blob/a34da315db294299278f679b04760f049fbf6337/connexion/middleware/main.py#L257-L264).

PR #1716, opened 2023-06-16
Bugfix/basepath

Ruwann:
Working towards #1709 

I think we're almost there, some tests I did are now working properly.

Would love to get some feedback/ideas on the implementation and the tests :)

### DIFF
--- a/connexion/middleware/main.py
+++ b/connexion/middleware/main.py
@@ -308,6 +308,9 @@ class ConnexionMiddleware:
                 app = middleware(app)  # type: ignore
             apps.append(app)
 
+        # We sort the APIs by base path so that the most specific APIs are registered first.
+        # This is due to the way Starlette matches routes.
+        self.apis = utils.sort_apis_by_basepath(self.apis)
         for app in apps:
             if isinstance(app, SpecMiddleware):
                 for api in self.apis:
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -432,27 +432,46 @@ def inspect_function_arguments(function: t.Callable) -> t.Tuple[t.List[str], boo
 T = t.TypeVar("T")
 
 
-def sort_routes(routes: t.List[T], *, key: t.Optional[t.Callable] = None) -> t.List[T]:
+@t.overload
+def sort_routes(routes: t.List[str], *, key: None = ...) -> t.List[str]:
+    ...
+
+
+@t.overload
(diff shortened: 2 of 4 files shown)


connexion/middleware/main.py
```diff
@@ -308,6 +308,9 @@ def _build_middleware_stack(self) -> t.Tuple[ASGIApp, t.Iterable[ASGIApp]]:
                 app = middleware(app)  # type: ignore
             apps.append(app)
 
+        # We sort the APIs by base path so that the most specific APIs are registered first.
```
Ruwann, line 311: There might also be an issue with the ordering of routes inside an API (when adding the operations of a spec to an API)

RobbeSneyders: Indeed, we can probably quite easily reuse the same sorting function there?

connexion/utils.py
```diff
@@ -423,3 +427,44 @@ def inspect_function_arguments(function: t.Callable) -> t.Tuple[t.List[str], boo
...
+
+    class SortableRoute:
+        def __init__(self, path: str) -> None:
+            self.path = path
+            self.path_regex, _, _ = compile_path(path)
+
+        def __lt__(self, other: "SortableRoute") -> bool:
+            return bool(other.path_regex.match(self.path))
+
+    return sorted(routes, key=lambda r: SortableRoute(key(r) if key else r))
```
RobbeSneyders, line 460: Smart :exploding_head:

RobbeSneyders (commented):
Thanks @Ruwann!

This approach looks good to me. Unless I'm missing something, I don't think we're sorting paths without path parameters correctly though?

Eg. if you look at the example in the issue, we should also sort `"/swagger"` before `"/"`. I think this is only necessary for the APIs, not for individual routes, but I don't think it has any downside there.

connexion/middleware/main.py
Ruwann: Yes, I'll investigate further in a separate issue as I did a small test that seemed to work fine even though the routes were in the "wrong" order.

Ruwann:
Thanks, good catch. Indeed, this was not yet working properly because the tests added the `/{path:path}` part that's actually needed. Fixed this now :)
Agreed, don't think it has downsides for routes as `/` and `/basepath` are technically not comparable, neither is more specific than the other, but this changes for basepaths.

RobbeSneyders (approved):
Thanks @Ruwann!

PR #1718, opened 2023-07-04
Support for generating examples in MockResolver for...

NanuakShadow:
…string, number, integer, bool

Fixes # .

 - Implements the generation of an example dependent on information provided in the case that an example isn't provided already.
 - Types included: strings, incl. maximum and minimum string length and regular expressions (with the use of an external dependency , rstr.  Integers, incl. maximum and minimum values as well as their respective exclusive values, i.e. <, > over <=, >=. Boolean and numbers, both in a more simple state since they don't have much variations.
 -


NanuakShadow:
@Ruwann @RobbeSneyders could you review or make any comments?

connexion/utils.py
```diff
@@ -10,6 +10,7 @@
 import pkgutil
 import sys
 import typing as t
+from rstr import xeger
```
mjp4, line 13: This is an external dependency that will need to be added to https://github.com/spec-first/connexion/blob/main/pyproject.toml#L46.

It should possibly be optional (like `flask` and `swagger-ui`).

tests/test_mock2.py
```diff
@@ -0,0 +1,110 @@
+from connexion.utils import generate_example
+from re import fullmatch, Match
+
+
+def test_short_string():
```
mjp4, line 5: I think a better name would be `test_generate_example_string`. And similar below.

tests/test_mock2.py
```diff
@@ -0,0 +1,110 @@
...
+    pattern = "^\d{3}-\d{2}-\d{4}$"
+
+    schema = {
+                "type": "string",
+                "pattern": pattern,
+            }
+    
+    example = generate_example(schema)
+
+    assert type(fullmatch(pattern, example)) == Match and isinstance(example, str)
```
mjp4, line 109: `fullmatch` either returns a `Match` object or `None. Instead of checking the if the returned type is `Match`, it is cleaner to do the inverse and check it is not `None`.

NanuakShadow: When I'd tried previously, it didn't work for some reason. It's working now though so I'll correct that.

RobbeSneyders:
Thanks @NanuakShadow!

While I agree that generating example responses based on the schema would be useful, I don't want to take on the effort to maintain this in Connexion. This PR implements this logic only for the most simple schema's, which means that we'll get issues for every case where this is not working.

If there is a third-party library that can provide this behavior, I'm happy to integrate it.

RobbeSneyders:
Closing this based on my comment above.

Issue #1531, opened 2022-05-06
Improve and update documentation for version 3.0


RobbeSneyders:
Proposed structure:

- [x] Index (#1646)
- [x] Getting started (#1701)
- [x] Middleware (#1706)
- [x] Routing (#1738)
- [x] Swagger UI (#1739)
- [x] Request handling (#1741)
- [x] Response handling (#1742)
- [x] Validation (#1743)
- [x] Security (#1764)
- [x] Context (#1747)
- [x] Exceptions (#1758)
- [x] CLI (#1763)
- [x] Lifespan (#1759)
- [x] Testing (#1760)
- [x] Migrating from 2 to 3 (#1646)

We should also
- [ ] Update the README (#1769)
- [x] Write a blogpost about the migration

jagadeeshgarvis:
Is there any examples for uploading and accessing files. I am facing issues regarding this in connexion 3.0.

`await connexion.request.form()` makes the application go into a limbo..

PR #1747, opened 2023-10-17
Add context documentation

### LINKED ISSUE #937: Attach operation specification to connexion.request

RobbeSneyders:
Contributes towards #1531 

Fixes #937

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -1,0 +1,114 @@
+Context
+=======
+
+The ``ContextMiddleware`` included in Connexion provides some information about the current request
+context as thread-safe request-level global variables.
+
+You can access them by importing them from ``connexion.context``:
+
+.. code-block:: python
+
+    from connexion.context import context, operation, receive, request, scope
+    from connexion import request  # alias for connexion.context.request
+
+Note that when trying to access these context variables outside of the request handling flow, or
+without running the ``ContextMiddleware``, the following ``RuntimeError`` will be thrown:
+
+.. code-block:: text
+
+    RuntimeError: Working outside of operation context. Make sure your app is wrapped in a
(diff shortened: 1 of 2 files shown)


Ruwann (approved):
Thanks, LGTM!

Although there is still the limitation in the security functions as the operation (specification) is not available there (which is mentioned in one of the comments)

RobbeSneyders:
Yes true. I'll add it to the `security` documentation.

We could make it available btw. The `SecurityOperation` that calls the security function has access to the `Operation` specification, as well as the other information we set as context variables.

We could either pass it into the security function directly, or set (and unset?) the context variables in each middleware, so code executed by the middleware has access to the context at that stage of the stack. Eg. custom validators would be able to use it as well. Let's discuss further in #1750

PR #1749, opened 2023-10-22
Remove debug argument from example

RobbeSneyders:
Small fix


Ruwann (approved):
Thanks, @RobbeSneyders !

PR #1752, opened 2023-10-23
Drop Python 3.7 in favor of 3.11 and 3.12

RobbeSneyders:
Python 3.7 is end of life. We should support the newer versions instead.

I also changed tox to only run against the minimum dependency versions on the lowest Python version, since this should lead to the lowest versions over all Python versions, and hopefully helps speed up our pipelines :)


Ruwann (approved):
Thanks!

PR #1761, opened 2023-10-29
Improve JSON req error on disallowed empty body

### LINKED ISSUE #1152: Should `None is not of type 'object'` be friendlier?
### Description
When an endpoint that expects a body is called without one, `None is not of type 'object'` is returned. Would you agree that a friendlier message would be preferable?

### Expected behaviour
A friendlier message saying something along the lines "A request body is required".

### Actual behaviour
`None is not of type 'object'`

### Steps to reproduce
- Endpoint expecting a request with a JSON body
- Call it without providing a body

### Additional info:
N/A

uranusjr:
Fixes #1152.

Currently, when a request body is empty, the JSON request validator would parse it into None, which is later passed down to the JSON Schema validator. However, jsonschema's validation error message for this case (when nullable is false) "None is not of type 'object'" is not particularly friendly to either the API user, nor the website developer.

This change adds a specific check before the None value is passed to jsonschema to emit a better error message directly.

I also added some drive-by improvements on function argument typing since _validate in validators don't seem to expect receiving None, but _parse (the result of which is passed to _validate) is totally allowed to return None (or anything really). This does not seem to reflect the logic well.

I’m not exactly sure if this is the best way to do this in Connexion 3.x. [We do have a patch in Connexion 2.x to achieve a similar effect](https://github.com/apache/airflow/commit/e89a7eeea6a7a5a5a30a3f3cf86dfabf7c343412) so if anyone understands how the two implementations correspond please tell me whether the two do the same thing 🙂

### DIFF
--- a/connexion/validators/form_data.py
+++ b/connexion/validators/form_data.py
@@ -70,12 +70,13 @@ class FormDataValidator(AbstractRequestBodyValidator):
 
         return data
 
-    def _validate(self, data: dict) -> None:
+    def _validate(self, body: t.Any) -> t.Optional[dict]:
+        if not isinstance(body, dict):
+            raise BadRequestProblem("Parsed body must be a mapping")
         if self._strict_validation:
-            self._validate_params_strictly(data)
-
+            self._validate_params_strictly(body)
         try:
-            self._validator.validate(data)
+            self._validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
             logger.error(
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Thanks @uranusjr!

I see your patch contains a test as well. Could you add it to this PR? [test_json_validation.py](https://github.com/spec-first/connexion/blob/main/tests/test_json_validation.py) is probably the most logical place to put it.

The pipelines are failing due to mypy in pre-commit. Could you have a look? You can easily set up pre-commit locally by running
```
pip install poetry
poetry install pre-commit
pre-commit install
```

RobbeSneyders (approved):

PR #1764, opened 2023-10-29
Update security docs

RobbeSneyders:
Contributes to #1531

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -1,46 +1,136 @@
 Security
 ========
 
+Connexion implements a pluggable security validation mechanism and provides built-in support for
+some of the most popular security schemes.
+
+.. csv-table::
+    :widths: 30, 70
+    :header-rows: 1
+
+    **Swagger 2**, **Connexion support**
+    Basic Authentication, |:white_check_mark:|
+    API key, |:white_check_mark:|
+    Oauth2, |:white_check_mark:|
+    **OpenAPI**,
+    HTTP Basic, |:white_check_mark:|
+    HTTP Bearer, |:white_check_mark:|
+    Other HTTP schemes (RFC 7253), "No built-in support, use a `custom security handler <#custom-security-handlers>`_"
+    API key, |:white_check_mark:|
+    Oauth2, |:white_check_mark:|
+    OpenID, "No built-in support, use a `custom security handler <#custom-security-handlers>`_"
+
+General authentication flow
+---------------------------
+
+For each supported authentication type, Connexion lets you register a validation function to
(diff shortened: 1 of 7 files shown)


docs/security.rst
```diff
@@ -1,87 +1,146 @@
...
 
-With Connexion, the API security definition **must** include a
-``x-basicInfoFunc`` or set ``BASICINFO_FUNC`` env var. It uses the same
-semantics as for ``x-tokenInfoFunc``, but the function accepts three
-parameters: username, password and required_scopes.
+For Basic authentication, the API security definition must include an
+``x-basicInfoFunc`` definition or set the ``BASICINFO_FUNC`` environment variable.
+
+The function should accept the following arguments:
+- username
```
Ruwann, line 69: These don't get rendered properly:
image

I believe an extra empty line is needed before the first list item

docs/security.rst
```diff
@@ -96,45 +155,70 @@ Multiple OAuth2 security schemes in AND fashion are not supported.
...
+    .. tab-item:: AsyncApp
+        :sync: AsyncApp
+
+        .. code-block:: python
+            :caption: **app.py**
+
+            from connexion import AsyncApp
+
+            app = AsyncApp(__name__, security_map=security_map)
+            app.add_api("openapi.yaml", security_map=security_map)
```
Ruwann, line 194: Minor comment: clarify only one is needed?

docs/security.rst
```diff
@@ -1,87 +1,146 @@
...
+    Oauth2, |:white_check_mark:|
+    **OpenAPI**,
+    HTTP Basic, |:white_check_mark:|
+    HTTP Bearer, |:white_check_mark:|
+    Other HTTP schemes (RFC 7253), "No built-in support, use a `custom security handler <#custom-security-handlers>`_"
+    API key, |:white_check_mark:|
+    Oauth2, |:white_check_mark:|
+    OpenID, "No built-in support, use a `custom security handler <#custom-security-handlers>`_"
+
+General authentication flow
```
Ruwann, line 23: Good indeed to explain the general idea

docs/security.rst
RobbeSneyders: Was thinking about this as well, but only added it in the text for now.

I use this pattern all over the docs though, so if we change it, we should do it everywhere, and I would do it in a separate PR.

Ruwann: Makes sense, better to keep it consistent indeed :)

Let's keep it for now and we can always clarify in the future if needed

Ruwann (approved):

PR #1765, opened 2023-10-29
Remove cookbook docs and sort index better

RobbeSneyders:
Contributes to #1531


docs/cookbook.rst
```diff
@@ -1,151 +0,0 @@
...
-          parameters:
-            - name: "path"
-              in: path
-              description: "Remainder of path, including slashes."
-              schema:
-                type: string
-                format: path
-
-
-Custom type format
```
Ruwann, line 31: I don't think we have any documentation elsewhere on this, perhaps it could be interesting to keep a small reference in another part. E.g. in the `Validation` section?

Ruwann (approved):
LGTM, the cookbook was a bit unclear :)
Perhaps one part that we can move instead of just remove?

docs/cookbook.rst
RobbeSneyders: Good catch, I dismissed it too quickly, thinking this was about path parameter formats, which I already added to the routing documentation.

PR #1778, opened 2023-11-02
Fail on invalid token when checking multiple OR security schemes

### LINKED ISSUE #1767: Allow fast fail when mutliple security handlers are provided
### Description

When multiple security handlers are provided, for instance
```
security:
- Bearer: []
- {}
```
To allow either a token or a non authenticated request, assuming a token is provided, but for instance has a wrong signature, or is expired, it should be possible to reject the request.

### Expected behaviour
There should be a mechanism to reject the request without looking for the next provider.

For instance raising an OAuthProblem or returning some specific value.

### Actual behaviour
The request is considered non-authenticated and the incorrect token is ignored.

### Additional info:

Output of the commands:

- `python --version`

RobbeSneyders:
Fixes #1767

### DIFF
--- a/connexion/security.py
+++ b/connexion/security.py
@@ -558,24 +558,15 @@ class SecurityHandlerFactory:
     @classmethod
     def verify_security(cls, auth_funcs):
         async def verify_fn(request):
-            token_info = NO_VALUE
-            errors = []
             for func in auth_funcs:
-                try:
-                    token_info = func(request)
-                    while asyncio.iscoroutine(token_info):
-                        token_info = await token_info
-                    if token_info is not NO_VALUE:
-                        break
-                except Exception as err:
-                    errors.append(err)
-
+                token_info = func(request)
+                while asyncio.iscoroutine(token_info):
+                    token_info = await token_info
+                if token_info is not NO_VALUE:
+                    break
             else:
-                if errors != []:
-                    cls._raise_most_specific(errors)
-                else:
-                    logger.info("... No auth provided. Aborting with 401.")
-                    raise OAuthProblem(detail="No authorization token provided")
+                logger.info("... No auth provided. Aborting with 401.")
+                raise OAuthProblem(detail="No authorization token provided")
 
             request.context.update(
                 {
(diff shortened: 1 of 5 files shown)


Ruwann (commented):
I'm not sure whether this prohibits some valid options. 
If you have 2 different OAuth2 providers for example (which I believe is valid), both using the `Authorization: Bearer` header/format, this will no longer be possible.

So I'm thinking we need to make it optional to "stop early" in the security checking, e.g. by making a special exception that basically wraps the real exception you want to throw. Then we can catch that specific one and fail early in that case, otherwise we continue and evaluate at the end.

RobbeSneyders:
> I'm not sure whether this prohibits some valid options.
> If you have 2 different OAuth2 providers for example (which I believe is valid), both using the Authorization: Bearer header/format, this will no longer be possible.

Hmm interesting, I didn't think of that.

> So I'm thinking we need to make it optional to "stop early" in the security checking, e.g. by making a special exception that basically wraps the real exception you want to throw. Then we can catch that specific one and fail early in that case, otherwise we continue and evaluate at the end.

I think we can maybe still generalize it in the following way: "If a known security token is included in the request, it should be valid against at least one of the schemes".

Maybe an implementation like this could work (pseudo):
```python
valid_tokens = {}
invalid_tokens = defaultdict(list)

token_info = NO_VALUE
for func in auth_funcs:
    token_name = auth_func.token_name
    try:
        token_info = func(request)
        while asyncio.iscoroutine(token_info):
            token_info = await token_info
        if token_info is not NO_VALUE:
            valid_tokens[token_name] = token_info
    except Exception as err:
        invalid_tokens[token_name].append(error)

if not valid_tokens:
        logger.info("... No auth provided. Aborting with 401.")
        raise OAuthProblem(detail="No authorization token provided")
    
for token, errors in invalid_tokens.items():
    if token in valid_tokens:
        continue
    else:
        cls._raise_most_specific(errors)

# TODO: Combine valid token_infos like with multiple AND schemes
```

Ruwann:
That sounds like a valid approach. I don't know how multiple security schemes can be handled properly here though, as the `auth_funcs` can be a `verify_multiple_schemes` function, so can be mapped to multiple token names

PR #1796, opened 2023-11-08
Fix middleware doc example

ajoga:
I'm starting to use the project with the newly release 3 version.

https://github.com/spec-first/connexion/blob/70084bcc4c21713cbf861841ebda977263d3ab8c/docs/middleware.rst?plain=1#L115-L120

It seems this example is not correct. 

1. missing import for SecurityMiddleware
2. `ConnexionMiddleware.default_middlewares` is a list of classes, not class instances: the filtering never happen.


RobbeSneyders (approved):
Thanks @ajoga!

PR #1818, opened 2023-11-20
Fix missing colons in request docs; rename /path to /foo for consistency with other examples

danielbprice:


RobbeSneyders (approved):
Thanks @danielbprice!

PR #1819, opened 2023-11-24
Only instantiate specification once

### LINKED ISSUE #1801: Slowdown of cold start in connexion 3.0
### Description
When switching to connexion 3.0 we noticed a considerable slowdown at startup.
Startup with connexion 3.x: 700ms-1s
Startup with connexion 2.x: 1ms-10ms

### Expected behaviour
No slowdown at startup compared to connexion 2.x

### Actual behaviour
Slowdown at startup.

### Steps to reproduce
* Have a medium sized OpenApi specification in yaml format (700 lines)
* Have a simple app created with a factory method:
  ```python
  # openapi_server/factory.py

  from connexion import App


  def create_app() -> App:
      connexion_app = App(__name__, specification_dir="./openapi/")
      connexion_app.add_api("openapi.yaml")
      return connexion_app

  ```
* Use pytest and have a simple dummy unit test:
  For connexion 3.x:
  ```python
  # tests/test_dummy.py

  import pytest
  from openapi_server.factory import create_app

  @pytest.fixture()
  def app():
    return create_app().test_client()

  @pytest.mark.parametrize("arg", range(10))
  def test_startup(app, arg):
    app.get("/")
  ```
  For connexion 2.x:
  ```python
  # tests/test_dummy.py

  import pytest
  import webtest
  
  from openapi_server.factory import create_app
  
  
  @pytest.fixture()
  def app():
      return webtest.TestApp(create_app())
  

  @pytest.mark.parametrize("arg", range(10))
  def test_startup(app, arg):
    app.get("/")

  ```
* Timings:
  * connexion 3.x: `pytest tests/test_dummy.py  8,97s user 0,11s system 98% cpu 9,188 total`
  * connexion 2.x: `pytest tests/test_dummy.py  2,36s user 0,10s system 85% cpu 2,888 total`

### Additional info:

Output of the commands:

- `python --version`
  Python 3.9.18
  Version: 3.0.1

What I tried:
* Use AsyncApp (no speedup)
* Use different backends for starlette (no speedup)
* Use a different python version (some speedup)

Looking at the profiler it seems like the yaml parsing now takes more time?

RobbeSneyders:
Fixes #1801 

I had to make quite a few additional changes to satisfy mypy.

### DIFF
--- a/connexion/middleware/security.py
+++ b/connexion/middleware/security.py
@@ -37,6 +37,16 @@ class SecurityOperation:
         next_app: ASGIApp,
         security_handler_factory: SecurityHandlerFactory,
     ) -> "SecurityOperation":
+        """Create a SecurityOperation from an Operation of Specification instance
+
+        :param operation: The operation can be both an Operation or Specification instance here
+            since security is defined at both levels in the OpenAPI spec. Creating a
+            SecurityOperation based on a Specification can be used to create a SecurityOperation
+            for routes not explicitly defined in the specification.
+        :param next_app: The next ASGI app to call.
+        :param security_handler_factory: The factory to be used to generate security handlers for
+            the different security schemes.
+        """
         return cls(
             next_app=next_app,
             security_handler_factory=security_handler_factory,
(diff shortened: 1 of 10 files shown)


connexion/middleware/security.py
```diff
@@ -31,7 +32,7 @@ def __init__(
     @classmethod
     def from_operation(
         cls,
-        operation: AbstractOperation,
+        operation: t.Union[AbstractOperation, Specification],
```
Ruwann, line 35: I don't immediately understand where this is coming from: why can this now be a `Specification` object as well? The variable name is pretty confusing now :)

examples/helloworld/hello.py
```diff
@@ -9,7 +9,7 @@ def post_greeting(name: str) -> str:
 
 app = connexion.FlaskApp(__name__, specification_dir="spec/")
 app.add_api("openapi.yaml", arguments={"title": "Hello World Example"})
-app.add_api("swagger.yaml", arguments={"title": "Hello World Example"})
+# app.add_api("swagger.yaml", arguments={"title": "Hello World Example"})
```
Ruwann, line 12: Remant from local testing?

```suggestion
app.add_api("swagger.yaml", arguments={"title": "Hello World Example"})
```

PR #1830, opened 2023-12-07
Specification use raw spec on clone

### LINKED ISSUE #1829: Connexion3 Failed to load API definition in the Swagger UI
### Description
It works perfectly fine locally and it works fine during some time after deployment. But after some time returns 500 error code. API itself works fine and all end-points works fine. The issue with Swagger UI only. 
Before we added an exception handler we had just 500 error with the default message saying that something went wrong or app overloaded.
Now we use the following error_handler:
```
def connexion_base_exception_handler(
    request: ConnexionRequest, exc: Exception
) -> ConnexionResponse:
    return ConnexionResponse(
        status_code=500,
        body=json.dumps({"error": str(exc)}),
        content_type="application/json",
    )
```
As a result, str(exc) is actually openapi.json content

Screenshot 2023-12-06 at 09 36 49

### Expected behaviour
API definition loaded correctly

### Additional info:

Output of the commands:

- `python --version`  python:3.10
- `pip show connexion | grep "^Version\:"`. connexion[flask]==3.0.3
-  `starlette`. 0.33.0 or 0.32.0.post1

nielsbox:
Fixes #1829.

 - Use _raw_spec when cloning

### DIFF
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -196,7 +196,7 @@ class Specification(Mapping):
         return OpenAPISpecification(spec, base_uri=base_uri)
 
     def clone(self):
-        return type(self)(copy.deepcopy(self._spec))
+        return type(self)(copy.deepcopy(self._raw_spec))
 
     @classmethod
     def load(cls, spec, *, arguments=None):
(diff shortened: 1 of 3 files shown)


RobbeSneyders (approved):
Thanks @nielsbox!

PR #1833, opened 2023-12-10
Bugfix/starlette root path

Ruwann:
Fixes the change in behaviour in starlette v0.33

### DIFF
--- a/connexion/middleware/routing.py
+++ b/connexion/middleware/routing.py
@@ -35,8 +35,12 @@ class RoutingOperation:
             scope.get("path_params", {})
         )
 
-        api_base_path = scope.get("root_path", "")[
-            len(original_scope.get("root_path", "")) :
+        api_base_path = scope.get("route_root_path", scope.get("root_path", ""))[
+            len(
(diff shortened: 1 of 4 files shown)


connexion/middleware/routing.py
```diff
@@ -35,8 +35,12 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
         )
 
-        api_base_path = scope.get("root_path", "")[
-            len(original_scope.get("root_path", "")) :
+        api_base_path = scope.get("route_root_path", scope.get("root_path", ""))[
+            len(
+                original_scope.get(
+                    "route_root_path", original_scope.get("root_path", "")
+                )
+            ) :
```
Ruwann, line 43:
```suggestion
        def get_root_path(scope):
            return scope.get("route_root_path", scope.get("root_path", ""))

        api_base_path = get_root_path(scope)[
            len(get_root_path(original_scope) :
```

RobbeSneyders (approved):
Thanks @Ruwann!

PR #1837, opened 2023-12-14
Add API-key scope checking

etvahala:
Fixes # N/A

The current documentation mentions that API-key
security supports scopes: "
The function should accept the following arguments:
- apikey
- required_scopes (optional) "

However, the scopes were not passed to the checker.

 - Add missing parameter routing to ApiKeySecurityHandler
 - Add a unit test for API-key scopes

### DIFF
--- a/tests/decorators/test_security.py
+++ b/tests/decorators/test_security.py
@@ -209,7 +209,7 @@ async def test_verify_apikey_query():
 
     security_handler_factory = ApiKeySecurityHandler()
     wrapped_func = security_handler_factory._get_verify_func(
-        apikey_info, "query", "auth"
+        apikey_info, "query", "auth", None
     )
 
     request = ConnexionRequest(scope={"type": "http", "query_string": b"auth=foobar"})
@@ -225,7 +225,7 @@ async def test_verify_apikey_header():
 
     security_handler_factory = ApiKeySecurityHandler()
     wrapped_func = security_handler_factory._get_verify_func(
-        apikey_info, "header", "X-Auth"
+        apikey_info, "header", "X-Auth", None
     )
 
     request = ConnexionRequest(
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Thanks @etvahala!

2 points:
- The [openapi docs](https://swagger.io/docs/specification/authentication/#scopes) only describes scopes for `oauth2` and `OpenID` security schemes, and mentions the scopes are empty for all other schemes. https://github.com/OAI/OpenAPI-Specification/issues/1366 seems to mention that this was changed in the 3.1 spec, however I don't see this included in the [actual spec](https://spec.openapis.org/oas/v3.1.0).
  Adding scopes to an API key security block does seem to pass validation for 3.X specs, but not for 2.X specs. So I'm not sure if we should add support for this.
  @etvahala could you post the relevant parts of your spec to show how you define this?
  @Ruwann what do you think?

- The tests are failing. You can wait until we have decided about point 1, although I would be more inclined to support this if you can show that it can pass the test suite without introducing a lot of complexity to maintain.

RobbeSneyders:
Ping @etvahala @Ruwann

etvahala:
Fixed the unit tests.

The scope support for API keys was introduced in [this OpenAPI spec commit](https://github.com/OAI/OpenAPI-Specification/pull/1829/commits/5129d9667fa4f7848882d603b10f321dd71ae864), by introducing weasel words ["...and the list MAY be empty if authorization does not require a specified scope."](https://spec.openapis.org/oas/v3.1.0#patterned-fields-2).

tests/decorators/test_security.py
```diff
@@ -235,6 +235,36 @@ def apikey_info(apikey, required_scopes=None):
...
+        return_value=apikey_info,
+    ) as mock_resolve_func:
+        wrapped_func = security_handler_factory.get_fn(scheme_apikey, ["admin"])
+        mock_resolve_func.assert_called_once()
+
+        request = ConnexionRequest(
+            scope={"type": "http", "headers": [[b"x-auth", b"admin foobar"]]}
+        )
+
+        assert await wrapped_func(request) is not None
```
RobbeSneyders, line 265:
```suggestion
        assert await wrapped_func(request) == {"sub": "foo"}
```

RobbeSneyders (commented):
Thanks, one small suggestion.

Ruwann (approved):
Thanks! The current implementation looks good to me.
I believe we should also extend it to other security schemes, not only API keys?

For full reference, in the past, connexion re-used the `scopes` value from other security schemes as `required_scopes` as argument value security schemes that didn't allow `scopes`. This was addressed in PR #1474

evahala:
Can somebody merge this in?

PR #1841, opened 2023-12-22
Remove common error handler from ExceptionMiddleware

julienschuermans:
This PR removes the catch-all error handler for unhandled exceptions from the `ExceptionMiddleware`.

Since the introduction of @nielsbox' `ServerErrorMiddleware`, the duplication of this feature has become kind of confusing / unnecessary.

### DIFF
--- a/connexion/middleware/exceptions.py
+++ b/connexion/middleware/exceptions.py
@@ -58,7 +58,6 @@ class ExceptionMiddleware(StarletteExceptionMiddleware):
     def __init__(self, next_app: ASGIApp):
         super().__init__(next_app)
         self.add_exception_handler(ProblemException, self.problem_handler)  # type: ignore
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Thanks @julienschuermans. 

Agree with the goal of the PR, but seems like we need some more work to fix this since the tests are failing.

As far as I understand the issue, we are wrapping the `error_response` in the `ServerErrorMiddleware` with our `connexion_wrapper` from the `ExceptionMiddleware` which returns a coroutine, while the `ServerErrorMiddleware` expects a normal function here. We could solve this by just implementing the relevant part (translating between Connexion and Starlette) synchronously in the `error_response` method.

Another difference is that the `ServerErrorMiddleware` still raises the error after returning the error response. So we'll need to update our tests to accommodate this and catch the error. We also need to check what the effect of this is when running the application with different servers.

PR #1843, opened 2023-12-29
Fix file upload for base64 encoded files

### LINKED ISSUE #1825: Base64 upload not working (tested with swagger-ui)
### Description
I am working on understanding multipart/form-data and in my example I have a binary and a base64 file example.

### Expected behaviour

I should be able to pass through both of them, together or seperately in the swagger-ui interface

### Actual behaviour

Binary upload is working, but base64 is not

### Steps to reproduce
I have zipped the workbench, just run it. And try to upload base64 according to the openapi

### Additional info:

I had to downgrade starlett to "<=0.32.0.post1" just to start it
[Base64Issue.zip](https://github.com/spec-first/connexion/files/13546915/Base64Issue.zip)

Output of the commands:

- `python 3.11.0`
- `Version: 3.0.3`

RobbeSneyders:
Fixes #1825 

Files can be encoded as bytes or base64 ([openapi docs](https://swagger.io/docs/specification/describing-request-body/file-upload/)), we were only handling bytes before.

### DIFF
--- a/connexion/validators/form_data.py
+++ b/connexion/validators/form_data.py
@@ -61,10 +61,10 @@ class FormDataValidator(AbstractRequestBodyValidator):
                 value = data.getlist(key)
 
                 def is_file(schema):
-                    return (
-                        schema.get("type") == "string"
-                        and schema.get("format") == "binary"
-                    )
+                    return schema.get("type") == "string" and schema.get("format") in [
+                        "binary",
+                        "base64",
+                    ]
 
                 # Single file upload
                 if is_file(param_schema):


connexion/validators/form_data.py
```diff
@@ -61,10 +61,10 @@ async def _parse(self, stream: t.AsyncGenerator[bytes, None], scope: Scope) -> d
...
 
                 def is_file(schema):
-                    return (
-                        schema.get("type") == "string"
-                        and schema.get("format") == "binary"
-                    )
+                    return schema.get("type") == "string" and schema.get("format") in [
+                        "binary",
+                        "base64",
+                    ]
```
RobbeSneyders, line 67: y u do dis black :cry:

PR #1845, opened 2023-12-31
Fix quickstart WSGI example typo

rparini:
Small fix to the example code in the documentation for using the `a2wsgi` workaround. The `app` variable is not defined prior to `app = ConnexionMiddleware(app)`


RobbeSneyders (approved):
Thanks @rparini!

PR #1852, opened 2024-01-17
Allow serving swagger-ui as index

karisN:
Allow serving swagger-ui as index

### DIFF
--- a/connexion/middleware/swagger_ui.py
+++ b/connexion/middleware/swagger_ui.py
@@ -151,12 +151,12 @@ class SwaggerUIAPI(AbstractSpecAPI):
             url += "/"
             return RedirectResponse(url=url)
 
-        self.router.add_route(methods=["GET"], path=console_ui_path, endpoint=redirect)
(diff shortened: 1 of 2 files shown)


connexion/middleware/swagger_ui.py
```diff
@@ -151,12 +151,12 @@ async def redirect(request):
             url += "/"
             return RedirectResponse(url=url)
 
-        self.router.add_route(methods=["GET"], path=console_ui_path, endpoint=redirect)
+        self.router.add_route(methods=["GET"], path=console_ui_path or "/", endpoint=redirect)
```
RobbeSneyders, line 154: I would rather change the way the path is stripped above.

```python
console_ui_path = self.options.swagger_ui_path.strip().rstrip("/") + "/"
```

and removing the leading "/" from the suffixes we add everywhere.

RobbeSneyders (commented):
Thanks @karisN!

Pre-commit is failing. You can install it locally using:

```
pip install poetry
poetry install --all-extras
pre-commit install
```

PR #1857, opened 2024-01-30
ci: add dependabot config

afuetterer:

 - add dependabot config

Context:

There are some outdated GitHub Actions in your workflows. Instead of updating those manually, you could go for a monthly dependency update from the bot.

What do you think?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# dependabot
+# Ref: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


afuetterer:
I did not touch the python files at all, no idea why the tests are failing.

RobbeSneyders:
Thanks @afuetterer 

The tests are failing due to an updated dependency. I synced with main to resolve this.

I still need some convincing that this PR is an improvement though. Auto-updates might still require manual updates to the pipeline if the interface changes, while we have never had any issues with outdated Github actions in the past. So I'm seeing potential additional overhead so solve a problem we've never had.

afuetterer:
Alright, fair enough. So I am closing this?

PR #1870, opened 2024-02-09
Generate examples in MockResolver with JSF library

### LINKED ISSUE #1719: MockResolver should generate valid example data if no example is provided by the schema.
### Description
If an OpenAPI spec has examples, then sensible reponses are returned. Including building up nested objects and arrays (as of https://github.com/spec-first/connexion/pull/1148).

However, if no example field exists, then a default (and invalid) response is returned. This is causing a problem when trying to mock out a third-party schema (that I do not want to modify).

This issue is being resolved in https://github.com/spec-first/connexion/pull/1718. An earlier attempt to solve the same problem never got merged https://github.com/spec-first/connexion/pull/513. 

### Expected behaviour
Server returns some dummy data that matches the schema requirements.

### Actual behaviour
Server returns a body of `'No example response was defined.'`.

mjp4:
Fixes #1719 .

Builds on work of #1718, pulling in external library to provide the fake data from JSON schema.

 - Use JSF library to generate sample data for mocking APIs without examples.
 - Add this as a new extra dependency "mock"

### DIFF
--- a/connexion/apps/abstract.py
+++ b/connexion/apps/abstract.py
@@ -143,8 +143,8 @@ class AbstractApp:
         Register an API represented by a single OpenAPI specification on this application.
         Multiple APIs can be registered on a single application.
 
-        :param specification: OpenAPI specification. Can be provided either as dict, or as path
-            to file.
+        :param specification: OpenAPI specification. Can be provided either as dict, a path
+            to file, or a URL.
         :param base_path: Base path to host the API. This overrides the basePath / servers in the
             specification.
         :param name: Name to register the API with. If no name is passed, the base_path is used
--- a/connexion/cli.py
+++ b/connexion/cli.py
@@ -117,10 +117,13 @@ def create_app(args: t.Optional[argparse.Namespace] = None) -> AbstractApp:
 
     logging.basicConfig(level=logging_level)
 
-    spec_file_full_path = os.path.abspath(args.spec_file)
-    py_module_path = args.base_module_path or os.path.dirname(spec_file_full_path)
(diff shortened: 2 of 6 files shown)


mjp4:
@RobbeSneyders I hope that you will be happy to merge this in based on https://github.com/spec-first/connexion/pull/1718#issuecomment-1628662131

Let me know if you have any questions or comments.

connexion/operations/openapi.py
```diff
@@ -210,6 +210,10 @@ def _nested_example(self, schema):
         try:
             # Recurse if schema is an array
             return [self._nested_example(schema["items"])]
+        except KeyError:
+            pass
+        try:
+            return generate_example(schema)
         except KeyError:
             raise
```
RobbeSneyders, line 218: I don't think the `generate_example` will raise a `KeyError`. I think we only need to catch `ImportError`s here since the dependency is optional?

Same for `swagger2.py`.

RobbeSneyders (commented):
Thanks @mjp4! This indeed looks like a good way to support example generation.

connexion/operations/openapi.py
mjp4: Good spot. I've reworked that code to be a bit nicer.

RobbeSneyders:
You'll need to bump the `jsonschema` version to `>=4.17.3` to fix the pipeline.

Can you also add this to the [CLI docs](https://github.com/spec-first/connexion/blob/main/docs/cli.rst#running-a-mock-server)?

RobbeSneyders:
You'll need to bump `typing-extensions` to `>=4.6.1` as well.

PR #1871, opened 2024-02-09
Allow the specification to be specified as a URL.

mjp4:

 - Allow a specification to be specified as a URL that is downloaded when the App runs. In combination with the existing mock features, this makes it a single command to run a mock server for any published API.

### DIFF
--- a/connexion/middleware/main.py
+++ b/connexion/middleware/main.py
@@ -408,7 +408,11 @@ class ConnexionMiddleware:
         if self.middleware_stack is not None:
             raise RuntimeError("Cannot add api after an application has started")
 
-        if isinstance(specification, (pathlib.Path, str)):
+        if isinstance(specification, str) and (
+            specification.startswith("http://") or specification.startswith("https://")
+        ):
+            pass
+        elif isinstance(specification, (pathlib.Path, str)):
             specification = t.cast(pathlib.Path, self.specification_dir / specification)
 
             # Add specification as file to watch for reloading
--- a/connexion/middleware/main.py
+++ b/connexion/middleware/main.py
@@ -371,8 +371,8 @@ class ConnexionMiddleware:
         Register een API represented by a single OpenAPI specification on this middleware.
         Multiple APIs can be registered on a single middleware.
 
-        :param specification: OpenAPI specification. Can be provided either as dict, or as path
-            to file.
+        :param specification: OpenAPI specification. Can be provided either as dict, a path
+            to file, or a URL.
         :param base_path: Base path to host the API. This overrides the basePath / servers in the
             specification.
         :param name: Name to register the API with. If no name is passed, the base_path is used
--- a/connexion/spec.py
+++ b/connexion/spec.py
@@ -8,20 +8,18 @@ import json
 import os
 import pathlib
 import pkgutil
-import tempfile
 import typing as t
 from collections.abc import Mapping
 from urllib.parse import urlsplit
 
 import jinja2
 import jsonschema
-import requests
 import yaml
 from jsonschema import Draft4Validator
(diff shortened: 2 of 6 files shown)


perrystallings:
Do you have a few tests for this PR? 

In particular I'm interested in how this would work for specs without defined operation ids.

For context, I could see this being awesome for mocking external API service during development or even as part of a testing pipeline so that you automatically get the default/example response combined with request validation.

RobbeSneyders:
Thanks @mjp4,

I'm indeed wondering the same as @perrystallings. Another option would be to include this in the CLI instead, if that's the only place where it would be useful.

mjp4:
Thanks both, I've added a test.

I don't have an immediate proposal for handling APIs without Operation IDs, the ones I'm interested in all do.

As it stands, and combined with #1874 and #1870, it has the potential to provide useful dummy output in many situations.

I was also thinking of submitting a PR for a resolver that maps arbitrary Operation IDs to functions without making assumptions about module structure. Something like:

```
class DictResolver(connexion.Resolver):
    def __init__(self, handler_dict):
        def function_resolver(operation_id):
            function = handler_dict.get(operation_id)
            if function is not None:
                return function
            msg = f'Cannot resolve operationId "{operation_id}"! Supported operations: {list(handler_dict.keys())}'
            raise ValueError(msg)

        self.function_resolver = function_resolver

OPERATIONS = {
    "OperationOne": lambda: ("Body Text", 200)
}

resolver = DictResolver(OPERATIONS),
```

connexion/spec.py
```diff
@@ -158,6 +160,17 @@ def from_file(cls, spec, *, arguments=None, base_uri=""):
...
+    @classmethod
+    def from_url(cls, spec, *, arguments=None, base_uri=""):
+        """
+        Takes in a path to a YAML file, and returns a Specification
+        """
+        spec_content = requests.get(spec).content
+        with tempfile.NamedTemporaryFile() as tfile:
+            tfile.write(spec_content)
+            tfile.flush()
+            return cls.from_file(tfile.name, arguments=arguments, base_uri=base_uri)
```
RobbeSneyders, line 172: Could we reuse the code we already use to resolve remote references?

https://github.com/spec-first/connexion/blob/5de6dccfbb6a0ed39966dcff5038e7708c93d550/connexion/json_schema.py#L56-L62

I like that it raises a clear error for non-successful requests and doesn't use a tempfile.

connexion/middleware/main.py
```diff
@@ -408,7 +408,11 @@ def add_api(
         if self.middleware_stack is not None:
             raise RuntimeError("Cannot add api after an application has started")
 
-        if isinstance(specification, (pathlib.Path, str)):
+        if isinstance(specification, str) and (
```
RobbeSneyders, line 411: Can you add in the docstrings that this can be a URL? The docstring is repeated in a couple of places across the code base, so would be good to do a find and replace.

Probably also good to explicitly mention this in the CLI docs.

RobbeSneyders (commented):
> As it stands, and especially combined with https://github.com/spec-first/connexion/pull/1874 and https://github.com/spec-first/connexion/pull/1870, it has the potential to provide useful dummy output in many situations.

Really appreciate the effort across these PRs :+1: 

> I was also thinking of submitting a PR for a resolver that maps arbitrary Operation IDs to functions without making assumptions about module structure.

Happy to accept it if you can fit it cleanly in the current `Resolver` interface.

mjp4:
@RobbeSneyders I think this is ready to go in?

RobbeSneyders (approved):
Looks good indeed, thanks @mjp4!

PR #1872, opened 2024-02-09
Fix for JSON array parsing in multipart/form-data (#1869)

### LINKED ISSUE #1869: Arrays are parsed incorrectly with multipart/form-data
### Description

I use `multipart/form-data` with a property `bar` which is of type `array`. The encoding is set to `application/json`. The spec looks like this:

```yaml
openapi: 3.0.3
info:
  title: LIQUID Dataset Service
  version: 1.0.0
paths:
  /foo:
    post:
      x-openapi-router-controller: liquid_dataset_service.controllers.dataset_controller
      operationId: create_bar
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                bar:
                  type: array
                  items:
                    type: object
                    properties:
                      baz:
                        type: string
              required:
                - bar
            encoding:
              bar:
                contentType: application/json
      responses:
        204:
          description: No Content
```

### Expected behaviour

When running the request I'd expect to get the `bar` property passed into the body argument of my controller function.

### Actual behaviour

Server raises error: `Validation error: '["{\\n  \\"baz\\": \\"string\\"\\n}"]' is not of type 'object' - 'bar.0'`.

To me it looks like that the string is not properly parsed if it's an array. To verify I changed the type of `bar` from `array` of `object`'s to `array` of `string`'s:

```yaml
bar:
  type: array
  items:
    type: string
```

I passed two strings "one" and "two" and received this dict in my controller function: `{'bar': ['["one"', '"two"]']}`.

### Steps to reproduce

Create a controller function, e.g.:

```
def create_bar(body):
    return
```

Adjust property `x-openapi-router-controller` accordingly.

Fire up the swagger page or use curl to run the request.

### Additional info:

Output of the commands:

- `python --version`: Python 3.11.6
- `pip show connexion | grep "^Version\:"`: Version: 3.0.5

alfechner:
Fixes #1869 .

- Prioritise parsing JSON over splitting array

### DIFF
--- a/connexion/uri_parsing.py
+++ b/connexion/uri_parsing.py
@@ -160,10 +160,10 @@ class OpenAPIURIParser(AbstractURIParser):
             form_data[k] = self._resolve_param_duplicates(
                 form_data[k], encoding, "form"
             )
-            if defn and defn["type"] == "array":
-                form_data[k] = self._split(form_data[k], encoding, "form")
-            elif "contentType" in encoding and all_json([encoding.get("contentType")]):
+            if "contentType" in encoding and all_json([encoding.get("contentType")]):
                 form_data[k] = json.loads(form_data[k])
+            elif defn and defn["type"] == "array":
+                form_data[k] = self._split(form_data[k], encoding, "form")
             form_data[k] = coerce_type(defn, form_data[k], "requestBody", k)
         return form_data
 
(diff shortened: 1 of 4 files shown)


alfechner:
I just wanted to check if that change is according to your standards. I'm also not very familiar with GitHub. Please let me know if there is anything else I should do regarding this change.

Thanks a lot for your time and work!

RobbeSneyders (approved):
This looks good to me @alfechner, thanks!

Just had to rebase on main to validate that the tests were fine.

PR #1874, opened 2024-02-10
Add option to CLI, App and add_api to disable security verification.

mjp4:

 - Add option to CLI, App and add_api to disable security verification. This will make it easier to mock out external APIs that specify security, or to disable security for prototyping, for testing, or for deployment behind an API gateway handling security. (As per use case https://connexion.readthedocs.io/en/latest/middleware.html#customizing-the-middleware-stack)

### DIFF
--- a/connexion/apps/abstract.py
+++ b/connexion/apps/abstract.py
@@ -51,6 +51,7 @@ class AbstractApp:
         validate_responses: t.Optional[bool] = None,
         validator_map: t.Optional[dict] = None,
         security_map: t.Optional[dict] = None,
+        no_security: t.Optional[bool] = None,
     ) -> None:
         """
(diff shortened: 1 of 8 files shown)


RobbeSneyders:
Thanks @mjp4! Agree that this would be useful, we use a gateway for security as well. 

I'm still thinking about the approach though. Another option would be to remove the `SecurityMiddleware` from the stack. That would prevent us from having to propagate this argument everywhere, but would only work on an `App` level. Let me think about it and come back.

If we add this option, we should update the documentation example you linked though to prevent confusion.

Ruwann:
I would prefer removing the security middleware from the stack so we don't need to pass the information along everywhere. For example, imo the security middleware shouldn't have to parse this option but we should not have the security middleware in the middleware stack as a whole

PR #1889, opened 2024-02-21
[FIX] replicate swagger-ui problem and hacky solution

jdkent:
Fixes # .

 - add external reference to simple app
 - try to clone with raw spec, unless there are external references
   - I'm unsure what the specific problem was in #1829 so I don't know how to differentiate use cases.


RobbeSneyders:
Thanks @jdkent, but the tests are still failing with your fix.

Note that we do have [tests for relative refs](https://github.com/spec-first/connexion/tree/main/tests/fixtures/relative_refs), so not sure why this is failing. If there is an issue, I assume it is in the ref resolving, and we should fix it there.

jdkent:
thanks for pointing towards an existing test, for a more minimal example, try cloning the specification in the test:
```python
def test_relative_refs(relative_refs, spec):
    spec_path = relative_refs / spec
    specification = Specification.load(spec_path)
    assert "$ref" not in specification.raw
    # clone specification
    specification_clone = specification.clone()
    assert "$ref" not in specification_clone.raw
```

The clone step will fail since `base_uri` is no longer set and clone is being applied to the "raw spec" before dereferencing is applied instead of to the "loaded spec" where dereferencing is already applied. 

I could see a solution where base_uri is an optional argument passed with clone, or some reversion of this pull request (https://github.com/spec-first/connexion/pull/1830), but I do not know what problem that pull request was specifically fixing.

jdkent:
if I change the clone function back to:

```python
def clone(self):
        return type(self)(copy.deepcopy(self._spec))
```

and run this test from `test_swagger_ui.py`:
```
def test_simple(swagger_ui_app):
    app_client = swagger_ui_app.test_client()
    response = app_client.get("/v1.0/spec.json")
```

the tests still pass, so I'm unsure what the minimal example that makes deepcopying `self._spec` not viable in #1829, but deepcopying `self._raw_spec` fixes.

eharvey71:
This seems related to my issue. Unsure. #1909

chrisinmtown:
For what it's worth, I cannot make relative references work in a `FlaskApp`. That constructor does not accept a `base_uri` parameter, if that matters. I get this error:
```
    raise exceptions._RefResolutionError(exc) from exc
jsonschema.exceptions._RefResolutionError: [Errno 2] No such file or directory: '/components.yaml'
```

I put up a /draft/ PR with a small self-contained example to demonstrate, please see: https://github.com/spec-first/connexion/pull/2009

RobbeSneyders:
Closing, replaced by https://github.com/spec-first/connexion/pull/2089

PR #1893, opened 2024-02-23
Fix lifespan example

fcollonval:
Thanks a lot for this amazing tool.

When adding a `lifespan_handler` to my server, it only works if the handler is asynchronously. That is consistent with the [tests](https://github.com/spec-first/connexion/blob/994f53fb045f1668ab84c472f7dc06d093470cf1/tests/test_lifespan.py#L13) but not the documentation. So this PR aims to correct the documentation


RobbeSneyders (approved):
Thanks @fcollonval!

PR #1898, opened 2024-03-18
Sort routes in RoutingMiddleware

### LINKED ISSUE #1879: RoutingMiddleware should sort paths
Connexion middleware sorts all available APIs from most specific to least because of starlettte's routing behavior
https://github.com/spec-first/connexion/blob/5d7eb0bdab9ed404e1d294dac6ed951af3be7952/connexion/middleware/main.py#L334

Similar sorting algorithms should be done within an api's rout in the RoutingMiddleware

kosciCZ:
Fixes #1879

Please let me know if this is the correct place to perform the sort or if any more changes are needed.

I've ran into this issue while migrating a large API from connexion 2 to connexion 3 (Async app variant) and the changes implemented here fixed route mismatching.

### DIFF
--- a/connexion/middleware/routing.py
+++ b/connexion/middleware/routing.py
@@ -14,6 +14,7 @@ from connexion.middleware.abstract import (
 from connexion.operations import AbstractOperation
 from connexion.resolver import Resolver
 from connexion.spec import Specification
+from connexion.utils import sort_routes
 
 _scope: ContextVar[dict] = ContextVar("SCOPE")
 
@@ -129,6 +130,8 @@ class RoutingMiddleware(SpecMiddleware):
             next_app=self.app,
             **kwargs,
         )
+        # sort routes to make sure that more specific routes are matched first
+        api.router.routes = sort_routes(api.router.routes, key=lambda r: r.path)
 
         # If an API with the same base_path was already registered, chain the new API as its
         # default. This way, if no matching route is found on the first API, the request is


RobbeSneyders:
Thanks @kosciCZ, I think we need to update this in the [`AbstractRoutingMiddleware`](https://github.com/spec-first/connexion/blob/ee399fbfe627f86c461b353132e4d710f7127440/connexion/middleware/abstract.py#L87) though.

RobbeSneyders:
Hi @kosciCZ, do you have time to have another look at this?

kosciCZ:
@RobbeSneyders sorry about that, yes, definitely. Could you throw me some pointers on the general outline of the solution?

My idea with this was that this is mostly Starlette specific, since their route matching iterates over routes and performs regex matches until a match is found, so order is important. With Flask, this does not seem to be an issue (our API on sync connexion 2.x matched routes correctly without having to sort them). So, that's why I chose to sort them all after the routes have been mounted, as that seemed like the appropriate place. Would you like me to replicate that in the method you referenced?

kosciCZ:
@RobbeSneyders any thoughts on ^?

RobbeSneyders:
Hi @kosciCZ, I would prefer to update it in the [`AbstractRoutingMiddleware`](https://github.com/spec-first/connexion/blob/ee399fbfe627f86c461b353132e4d710f7127440/connexion/middleware/abstract.py#L87), so the behavior is consistent across all places where we are routing requests, even if the resulting behavior is the same.

PR #1918, opened 2024-04-30
Can return HTTPStatus member when default response is declared

### LINKED ISSUE #1917: Fails with 500 error when returning HTTPStatus and default response is configured
### Description

Connexion accepts return statements like `return {}, HTTPStatus.CREATED` but it fails with 500 error if response validation is enabled and if there is a default response with required fields configured in the schema.

### Expected behaviour

Just use `HTTPStatus` members like integers

### Actual behaviour

See above.

### Steps to reproduce

```py
from http import HTTPStatus

import pytest
from connexion import AsyncApp


async def fails():
    return {"data": []}, HTTPStatus.CREATED


async def ok():
    return {"data": []}, 201


@pytest.fixture
def app():
    app = AsyncApp(__name__)
    app.add_api(
        validate_responses=True,
        specification={
            "openapi": "3.0.0",
            "info": {"title": "test", "version": "v1"},
            "paths": {
                "/fails": {
                    "get": {
                        "x-openapi-router-controller": __name__,
                        "operationId": "fails",
                        "responses": {
                            "201": {
                                "description": "ok",
                            },
                            "default": {
                                "description": "default",
                                "content": {
                                    "application/json": {
                                        "schema": {
                                            "type": "object",
                                            "properties": {
                                                "error_code": {"type": "integer"},
                                            },
                                            "required": ["error_code"],
                                        },
                                    }
                                },
                            },
                        },
                    }
                },
                "/ok": {
                    "get": {
                        "x-openapi-router-controller": __name__,
                        "operationId": "ok",
                        "responses": {
                            "201": {
                                "description": "ok",
                            }
                        },
                    }
                },
            },
        },
    )
    return app


def test_ok(app):
    with app.test_client() as client:
        res = client.get("/ok")
        assert res.status_code == HTTPStatus.CREATED


def test_fails(app):
    # SHOULD NOT FAIL
    with app.test_client() as client:
        res = client.get("/fails")
        assert res.status_code == HTTPStatus.CREATED

```

### Additional info:

Output of the commands:

- `python --version`: 3.9.16
- `pip show connexion | grep "^Version\:"`: 3.0.6

chbndrhnns:
Fixes #1917  .

 - Look at enum value for return status code if given
 -
 -

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -138,7 +138,7 @@ class BaseResponseDecorator:
         elif len(handler_response) == 2:
             data, status_code_or_headers = handler_response
             if isinstance(status_code_or_headers, int):
-                status_code = status_code_or_headers
+                status_code = int(status_code_or_headers)
             elif isinstance(status_code_or_headers, Enum) and isinstance(
                 status_code_or_headers.value, int
             ):
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -138,6 +138,7 @@ class BaseResponseDecorator:
         elif len(handler_response) == 2:
             data, status_code_or_headers = handler_response
             if isinstance(status_code_or_headers, int):
+                # Extra int call because of int subclasses such as http.HTTPStatus (IntEnum)
                 status_code = int(status_code_or_headers)
             elif isinstance(status_code_or_headers, Enum) and isinstance(
                 status_code_or_headers.value, int
--- a/connexion/middleware/response_validation.py
+++ b/connexion/middleware/response_validation.py
@@ -97,7 +97,11 @@ class ResponseValidationOperation:
                 if message["status"] < 400:
                     self.validate_mime_type(mime_type)
 
-                status = str(message["status"])
+                try:
+                    status = str(message["status"].value)
+                except AttributeError:
+                    status = str(message["status"])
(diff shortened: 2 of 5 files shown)


RobbeSneyders:
Hi @chbndrhnns,

Thanks for the PR.
The response validation is not always hit though. Eg. if it is deactivated, or if there is no validator for the returned content type. We should probably implement this in the [response decorator](https://github.com/spec-first/connexion/blob/main/connexion/decorators/response.py).

chbndrhnns:
Please take a look at the linked issue. My fix only applies when `validate_responses` is `True`. Let me know if you still want to move the logic.

Ruwann:
The root cause is indeed in the response decorator, as `HTTPStatus` members are of type `IntEnum` so they pass the `isinstance(..., int)` check. We can check explicitly for this first, or we should also be able to resolve it by simply wrapping the status code in another `int()` call here, so:

https://github.com/spec-first/connexion/blob/823fa6c00a54e32b1edabc23fb288d34ab3f50c7/connexion/decorators/response.py#L141

becomes

```python
status_code = int(status_code_or_headers)
```

Edit: seems like the second if-branch should also catch it, so swapping the order could also work? Not sure in which cases that branch is entered :)

PR #1926, opened 2024-05-14
Log handled errors to warning

### LINKED ISSUE #1925: Error Logging
There are several places Connexion logs errors, for example:

* When exceptions are handled in [`problem_handler`](https://github.com/spec-first/connexion/blob/fcfe151e690c8d8864605ff94923c8b8ce834bfb/connexion/middleware/exceptions.py#L73), [`http_exception`](https://github.com/spec-first/connexion/blob/fcfe151e690c8d8864605ff94923c8b8ce834bfb/connexion/middleware/exceptions.py#L80) or [`common_error_handler`](https://github.com/spec-first/connexion/blob/fcfe151e690c8d8864605ff94923c8b8ce834bfb/connexion/middleware/exceptions.py#L93) in the [`ExceptionMiddleware`](https://connexion.readthedocs.io/en/latest/middleware.html)
* When validation of the input/output data fails in [`connexion.validators.json`](https://github.com/spec-first/connexion/blob/fcfe151e690c8d8864605ff94923c8b8ce834bfb/connexion/validators/json.py#L71)

Connexion does a great job of handling exceptions by retuning a problem details with the appropriate status code to the user.

Now it's pretty common for operation teams to monitor exceptions in the logs and trigger alarms.

And here comes the issue:

Operations want to be alerted about everything that results in `5xx` status codes. That indicates that a situation couldn't be handled from a user perspective. Operations don't need to be alerted if the status code is `4xx`. Because those cases are gracefully handled.

I'd propose to

1. change [`ExceptionMiddleware`](https://connexion.readthedocs.io/en/latest/middleware.html) to log to `error` if the resulting status code is `5xx` and to `warning` for `4xx` and
2. remove the error log from [`connexion.validators.json`](https://github.com/spec-first/connexion/blob/fcfe151e690c8d8864605ff94923c8b8ce834bfb/connexion/validators/json.py#L71) since the `BadRequestProblem` raised is logged via the Exception middleware anyway.

I'm happy to contribute those changes. What do you think?

alfechner:
Fixes #1925.

 - Log handled errors to `warning` instead of `error`.
 - Log validation errors to `info` because the intent of the log lines in informational. The error is handled by raising a new error.

### DIFF
--- a/connexion/validators/json.py
+++ b/connexion/validators/json.py
@@ -68,7 +68,7 @@ class JSONRequestBodyValidator(AbstractRequestBodyValidator):
             return self._validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
-            logger.error(
+            logger.info(
                 f"Validation error: {exception.message}{error_path_msg}",
                 extra={"validator": "body"},
             )
@@ -129,7 +129,7 @@ class JSONResponseBodyValidator(AbstractResponseBodyValidator):
             self.validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
-            logger.error(
+            logger.info(
                 f"Validation error: {exception.message}{error_path_msg}",
                 extra={"validator": "body"},
             )
--- a/connexion/validators/json.py
+++ b/connexion/validators/json.py
@@ -77,7 +77,8 @@ class JSONRequestBodyValidator(AbstractRequestBodyValidator):
 
 class DefaultsJSONRequestBodyValidator(JSONRequestBodyValidator):
     """Request body validator for json content types which fills in default values. This Validator
-    intercepts the body, makes changes to it, and replays it for the next ASGI application."""
+    intercepts the body, makes changes to it, and replays it for the next ASGI application.
+    """
 
     MUTABLE_VALIDATION = True
     """This validator might mutate to the body."""
@@ -129,7 +130,7 @@ class JSONResponseBodyValidator(AbstractResponseBodyValidator):
             self.validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
-            logger.info(
+            logger.warning(
                 f"Validation error: {exception.message}{error_path_msg}",
                 extra={"validator": "body"},
             )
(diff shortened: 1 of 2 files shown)


connexion/validators/json.py
```diff
@@ -129,7 +129,7 @@ def _validate(self, body: dict):
             self.validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
-            logger.error(
+            logger.info(
```
Ruwann, line 132: For this error, I propose to use the `warning` level as the response body should adhere to the schema. So in that sense it is more a server error than a client error (i.e. something for the application to fix).
What do you think?

Ruwann (commented):
Thanks for the PR!

I left one comment for a log that I would keep on something more severe than `info`.

connexion/validators/json.py
alfechner: Thanks for reviewing my proposal, really appreciate it! 👍 

The reason I put it to `info` is, that is doesn't contribute any value:

* `NonConformingResponseBody` inherits from `InternalServerError` and results in a `500` response . Therefore it's  logged to `error` level in `ExceptionMiddleware`.
* `BadRequestProblem` results in a `400` response. Therefore it's logged to `warning` level in `ExceptionMiddleware`.

I think that the error handling should be responsible for deciding about log levels, not the validation logic itself.

Now that I write about it: I'd rather propose to delete both log statements from the json logger completely.

alfechner: Changed the log level accordingly.

alfechner:
@Ruwann kind reminder to merge 🙏 🙇

connexion/validators/json.py
```diff
@@ -68,7 +68,7 @@ def _validate(self, body: t.Any) -> t.Optional[dict]:
             return self._validator.validate(body)
         except ValidationError as exception:
             error_path_msg = format_error_with_path(exception=exception)
-            logger.error(
+            logger.info(
```
Lewiscowles1986, line 71: It would be nice to configure this, as I'd quite like it as a warning; but not an error, and it's clear to me from this that many people may have different use-cases. I Know I can redirect a log-stream.

PR #1931, opened 2024-05-22
Reintroduce header param type conversion

CordulaGuder:
This PR fixes the type conversion of header params that seems to have gone missing during the refactoring.

In Connexion 2, the ParameterValidator used to do type conversion in its main function validate_parameter, so it did that for all kinds of params it came across. In Connexion 3, this has moved to the URIParser classes (connexion.uri_parsing), so header params are not being type converted anymore. This PR fixes this bug.

 - add type conversion to validate_header_parameter in ParameterValidator

### DIFF
--- a/connexion/validators/parameter.py
+++ b/connexion/validators/parameter.py
@@ -6,7 +6,7 @@ from jsonschema import Draft4Validator, ValidationError
 from starlette.requests import Request
 
 from connexion.exceptions import BadRequestProblem, ExtraParameterProblem
-from connexion.utils import boolean, is_null, is_nullable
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Thanks @CordulaGuder,

We explicitly deduplicated all `coerce_type` calls during the refactoring and moved them into the `uri_parser` so it is done in a single place. I indeed see that this misses the headers.

However by implementing it in the validation, the types in the headers are validated after coercion while they are passed to the application without coercion. If we want to reactivate this, we should centralize it as well. We can either also do it in the `URIParser`, although I'm not sure if that makes sense, or in the `ConnexionRequest` class.

See #1934 which was triggered by reviewing this PR.

PR #1934, opened 2024-05-27
Split parsing from validation completely

RobbeSneyders:
During the refactoring for Connexion 3, we deduplicated all `coerce_type` calls during the refactoring and moved them into the `uri_parser` so it is done in a single place. When looking into #1931 however, I noticed we are still using the `uri_parser` from more places than we should.

This PR centralizes the parsing further into the `ConnexionRequest` class. During validation, we now instantiate a `ConnexionRequest` instead of a Starlette `Request`, and leverage it instead of calling the `uri_parser` directly.

I split the parsing and validation in the tests so they can be tested in isolation.

### DIFF
--- a/connexion/validators/parameter.py
+++ b/connexion/validators/parameter.py
@@ -3,9 +3,9 @@ import copy
 import logging
 
 from jsonschema import Draft4Validator, ValidationError
-from starlette.requests import Request
 
 from connexion.exceptions import BadRequestProblem, ExtraParameterProblem
+from connexion.lifecycle import ConnexionRequest
 from connexion.utils import boolean, is_null, is_nullable
 
 logger = logging.getLogger("connexion.validators.parameter")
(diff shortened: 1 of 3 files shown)


Ruwann (approved):
Nice, thanks!

CordulaGuder:
Hi @RobbeSneyders, thanks a lot for looking into my PR and providing this change, which seems to make a lot of change! Trying it out, I found that it only applies to path and query params, and not to header params. Parameter validator gets [request.headers](https://github.com/spec-first/connexion/blob/main/connexion/validators/parameter.py#L93), which, on further investigation, is not a property of ConnexionRequest, but only exposes the headers property of the underlying Starlette request object. On this, no type conversion is being done, so I'm now seeing the exact same issue as before.

Trying to fix this by adding a header_params property to ConnexionRequest, I stumbled over more issues. For some reason, I don't know whether that is a bug in Connexion or a misconfiguration on my side, the URIParser used in the ConnexionRequest doesn't have the param_defn from the API spec, so it skips the type conversion. I decided to stop the migration effort here for a while as it's been taking too much time already. This is just to inform you about the things I've noticed.

PR #1940, opened 2024-06-18
Custom auth types in security_map

anna-intellegens:
Allow definition of new auth types (not just schemes) to be passed in to security_map
This is particularly motivated by wanting to support OpenIDConnect auth type, but was running into an error as it couldn't find the `scheme` key (as it's not permitted in the schema by the openAPI validator).

The [docs](https://connexion.readthedocs.io/en/stable/security.html#security) imply that you should be able to implement a custom scheme for this, instead I would get a [`KeyError`](https://gist.github.com/anna-intellegens/0161d967de923030a1cc9f822160a13a).

 - Update `parse_security_schemes` to allow the detection of auth types, not just schemes.

### DIFF
--- a/connexion/security.py
+++ b/connexion/security.py
@@ -492,9 +492,14 @@ class SecurityHandlerFactory:
                 security_handler = self.security_handlers["apiKey"]
                 return security_handler().get_fn(security_scheme, required_scopes)
 
-        # Custom security handler
-        elif (scheme := security_scheme["scheme"].lower()) in self.security_handlers:
+        # Custom security scheme handler
+        elif "scheme" in security_scheme and (scheme := security_scheme["scheme"].lower()) in self.security_handlers:
             security_handler = self.security_handlers[scheme]
+            return security_handler().get_fn(security_scheme, required_scopes)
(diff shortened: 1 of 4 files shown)


Ruwann (approved):
Thanks! I have updated the code a bit to already add `openIdConnect` as an option, without default implementation.

@RobbeSneyders feel free to review, I've kept the "other" custom scheme/type options even though the types allowed are limited, but didn't see much harm in keeping them, as new security types should then also be possible to be provided by the user

chrisinmtown:
Greetings to @RobbeSneyders and @Ruwann. The community opened two separate issues (#2061 and #2074) to report the problem that this PR fixes in Version 3.  Please advise if we can expect a new release of version 3 with this change anytime soon? Thanks in advance.

giuliohome:
Maybe also my issue https://github.com/spec-first/connexion/issues/2081 is related and it would be valuable to understand if this project is still under maintenance especially considered that this thread started more than a year ago and it looks like there is no public release even now that the PR was merged 4 months ago

Issue #1907, opened 2024-04-04
required readOnly fields are required in request body when they shouldn't be


sampoh0523:
### Description

When an API object has a field that is required but has readOnly: true, a request containing an object without that field fails to validate.

OpenAPI spec says:

> You can use the `readOnly` and `writeOnly` keywords to mark specific properties as read-only or write-only. This is useful, for example, when GET returns more properties than used in POST – you can use the same schema in both GET and POST and mark the extra properties as `readOnly`. `readOnly` properties are **included in responses but not in requests**, [...]

> If a `readOnly` or `writeOnly` property is included in the required list, required affects just the relevant scope – responses only or requests only. That is, read-only required properties **apply to responses only**, [...]

This is a regression from Connexion 2.x.

### Expected behaviour

Object validates successfully.

### Actual behaviour

Object fails to validate, because it is missing a read-only field:
```
ERROR:connexion.validators.json:Validation error: 'objectId' is a required property
ERROR:connexion.middleware.exceptions:BadRequestProblem(status_code=400, detail="'objectId' is a required property")
```

Presumably Connexion 3.x fails to use jsonschema validation properly here.

### Steps to reproduce

1. Use the openapi.yaml and testcontroller.py from below
2. Start with `connexion run openapi.yaml --stub`
3. Attempt to do the following request: PUT .../objects/45d2847a-ed55-440a-bfa7-5e759d0d671f, with body
```
{
    "name": "abc"
}
```

Minimal example:

#### openapi.yaml
```
openapi: 3.0.3
info:
  title: test
  version: '1.0'
paths:
  '/objects/{objectId}':
    parameters:
      - schema:
          type: string
          format: uuid
        name: objectId
        in: path
        required: true
    put:
      summary: Create or update object
      operationId: object_put
      x-openapi-router-controller: testcontroller
      responses:
        '200':
          description: OK - Object updated
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Object'
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Object'
components:
  schemas:
    Object:
      title: BaseObject
      type: object
      properties:
        objectId:
          type: string
          format: uuid
          readOnly: true
        name:
          type: string
      required:
      - objectId
      - name
```

#### testcontroller.py
```
from connexion import FlaskApp, request

app = FlaskApp(__name__)


def object_put(objectId, body):
    return {"objectId": objectId, **body}


app.add_api("openapi.yaml")
[...]

Issue #942, opened 2019-05-10
400 error for readOnly=True


Jyhess:
### Description
When a property marked readOnly is sent, connexion raises a 400 error.

This is not compliant with [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-20), which document a readOnly property *should not* be sent, and not *must not*:

> Declares the property as "read only". This means that it MAY be sent as part of a response but SHOULD NOT be sent as part of the request.

Moreover, this force to create a new version of an API if an existing field is marked readOnly, forcing all clients to change their implementation. It's easier if server can just ignore this field.

It also imply more complex code client side, as we need to clean sent object from readOnly fields for simple GET/modify/PUT workflows.

### Expected behaviour
When a property marked readOnly is sent, connexion accept it without issue.
If possible, it should be removed in the body received by endpoint.

### Actual behaviour
A 400 error is raised.

### Steps to reproduce
Create a schema with a readlony property:

    properties:
      name:
        type: string
        readOnly: true

and send it with a POST or PUT.
### Additional info:

This seems to have been implemented from issue #350 

Output of the commands:

- `python --version`
Python 3.6.8
- `pip show connexion | grep "^Version\:"`
Version: 2018.0.dev1

badcure:
I am normally against silently ignoring data with API...but agree here if it is part of the OpenAPI standard. That said, I think:
Default behavior:
* Remove the ReadOnly/WriteOnly fields.
* In the headers, such as `X-ReadOnly-Drop`, list out the fields that were removed from the app that was provided.
* Make sure logs, possibly `info`, document that dropping of the fields as well and what can be done to cause an error.

Allowed if a flag is set, such as strict_validation:
* Continue with the existing behavior, return a 400.

Essentially make it extremely verbose if the library is ignoring data.

dtkav:
I used to be in favor of rejecting the entire message (strict behavior) when this was implemented, but I've since changed my opinion. One of the folks from the JSON Schema org (https://github.com/handrews) had this to say about readOnly [0]

```
readOnly never causes JSON Schema's validation process to fail. It's not an assertion, just an annotation that applications can use to take action as they please.
```

Perhaps the decision to accept `readOnly` fields could be toggled by `additionalProperties` in OAS3.

IIRC we fixed the `additionalProperties` implementation (`true` by default) after adding readOnly/writeOnly validation.

[...]

PR #1943, opened 2024-06-26
Validate required fields based on readOnly & writeOnly

whoseoyster:
Fixes https://github.com/spec-first/connexion/issues/1907.

- Add `required` and `readOnly` validators to the `Draft4ResponseValidator` and `Draft4RequestValidator` objects

Let me know if anything else needs adjustments.

### DIFF
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -128,6 +128,29 @@ def allow_nullable(validation_fn: t.Callable) -> t.Callable:
     return nullable_validation_fn
 
 
+def validate_required(validator, required, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    for prop in required:
+        if prop not in instance:
+            properties = schema.get('properties')
+            if properties is not None:
+                subschema = properties.get(prop)
+                if subschema is not None:
+                    if 'readOnly' in validator.VALIDATORS and subschema.get('readOnly'):
+                        continue
+                    if 'writeOnly' in validator.VALIDATORS and subschema.get('writeOnly'):
+                        continue
+                    if 'x-writeOnly' in validator.VALIDATORS and subschema.get('x-writeOnly') is True:
+                        continue
+            yield ValidationError("%r is a required property" % prop)
+
+
+def validate_readOnly(validator, wo, instance, schema):
+    yield ValidationError("Property is read-only")
+
+
 def validate_writeOnly(validator, wo, instance, schema):
     yield ValidationError("Property is write-only")
 
@@ -140,6 +163,8 @@ Draft4RequestValidator = extend(
     {
         "type": NullableTypeValidator,
         "enum": NullableEnumValidator,
+        "readOnly": validate_readOnly,
+        "required": validate_required,
     },
 )
 
@@ -150,5 +175,6 @@ Draft4ResponseValidator = extend(
         "enum": NullableEnumValidator,
         "writeOnly": validate_writeOnly,
         "x-writeOnly": validate_writeOnly,
+        "required": validate_required,
     },
 )


chrisinmtown:
I cannot agree with @sampoh0523 here. I believe the intent of `readOnly` is that a request containing the field must be rejected, otherwise what is the point of marking it as such.

chrisinmtown:
@RobbeSneyders @Ruwann would you please approve a workflow run on this PR?

connexion/json_schema.py
```diff
@@ -128,6 +128,29 @@ def nullable_validation_fn(validator, to_validate, instance, schema):
     return nullable_validation_fn
 
 
+def validate_required(validator, required, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    for prop in required:
+        if prop not in instance:
+            properties = schema.get('properties')
```
chrisinmtown: I think the prevailing style in this code is to use double quotes (not single quotes) around strings, please consider matching that.

sampoh0523:
> I cannot agree with @sampoh0523 here. I believe the intent of `readOnly` is that a request containing the field must be rejected, otherwise what is the point of marking it as such. I'm going to add my approval vote to this PR as-is.

The OpenAPI specification argues otherwise, see e.g. <https://swagger.io/specification/>:

> Fields that are both required and read-only are an example of when it is beneficial to ignore a `readOnly: true` constraint in a PUT, particularly if the value has not been changed.

Note also the wording on the specification for version 3.0.3: https://spec.openapis.org/oas/v3.0.3.html

> [readOnly] means that it MAY be sent as part of a response but SHOULD NOT be sent as part of the request.

This is RFC 2119 wording, which explicitly allows readOnly fields to be sent as part of a request (only telling that it should not be done). If it was forbidden, it would be `MUST NOT`.

connexion/json_schema.py
```diff
@@ -128,6 +128,29 @@ def nullable_validation_fn(validator, to_validate, instance, schema):
     return nullable_validation_fn
 
 
+def validate_required(validator, required, instance, schema):
+    if not validator.is_type(instance, "object"):
```
chrisinmtown, line 132: Please add a hint for the return type, also possibly a Pydoc comment, maybe:

    """Check a request for required fields, ignoring fields marked readOnly"""

connexion/json_schema.py
```diff
@@ -128,6 +128,29 @@ def nullable_validation_fn(validator, to_validate, instance, schema):
...
+
+    for prop in required:
+        if prop not in instance:
+            properties = schema.get('properties')
+            if properties is not None:
+                subschema = properties.get(prop)
+                if subschema is not None:
+                    if 'readOnly' in validator.VALIDATORS and subschema.get('readOnly'):
+                        continue
+                    if 'writeOnly' in validator.VALIDATORS and subschema.get('writeOnly'):
```
chrisinmtown, line 143: Please help me understand why this validation condition is detected and ignored: a field is required by the schema, the field is marked `writeOnly` in the schema, and the field is not present in the instance. I think in this situation, Connexion should declare error?

chrisinmtown:
@sampoh0523  thanks for the quick reply. The problem stated in issue #1907 is that Connexion rejects a request that is MISSING a field if the schema marks that field as both `required` and `readOnly`, and this PR changes that behavior to accept. I think I finally understand you are stating something very closely related, namely that this change also absolutely blocks a request if it CONTAINS a field that is both `required` and `readOnly`; and you are stating Connexion should tolerate the request with the field. Did I finally get that right?

Parsing the difference between SHOULD NOT and MUST NOT gets kind of subtle here. :/

sampoh0523:
Yes, that is correct. #942 has further discussion on it (including quoted commentary by one of the JSON Schema designers), but (my point as well is that) `readOnly` should not cause a request to fail validation, even if the field in question is present in that request.

connexion/json_schema.py
```diff
@@ -128,6 +128,29 @@ def nullable_validation_fn(validator, to_validate, instance, schema):
...
+                        continue
+                    if 'x-writeOnly' in validator.VALIDATORS and subschema.get('x-writeOnly') is True:
+                        continue
+            yield ValidationError("%r is a required property" % prop)
+
+
+def validate_readOnly(validator, wo, instance, schema):
+    yield ValidationError("Property is read-only")
```
chrisinmtown: @sampoh0523 has convinced me that there is no need for this function, the argument is that Connexion should tolerate a readOnly field in requests without complaint.

sampoh0523: The current implementation of `validate_required` checks that some kind of function is specified for validating `readOnly` fields, since that is used to determine whether such fields can be skipped from the `required` validation. Deleting the function would therefore require some additional changes.

chrisinmtown: Oh good point. Why is it important for the `validate_required` function to check for presence of a `validate_readOnly` function? Asked differently, why is just checking for the `readOnly` subschema property not sufficient?

sampoh0523: It is because the same `validate_required` is used for both requests and responses, and it recognizes which is which by checking the validators used (if there is a `readOnly` validator, it is a request; if there is a `writeOnly` validator, it is a response).

chrisinmtown:
I think this PR should add the following new test cases:
* Schema defines field as required and readOnly, submit a request *without* the field, the request is accepted
* Schema defines field as required and readOnly, submit a request *with* the field, the request is accepted.

connexion/json_schema.py
chrisinmtown, line 143: Now I finally grasp that this line of code is validating a response, the field is marked as required, the field is also marked as writeOnly, and the function has detected that a response instance is MISSING the required writeOnly field. Adding the `continue` causes this condition to be ignored. Is this another subtle JSON schema validation thing?

connexion/json_schema.py
chrisinmtown: Please let me amend my first comment here.  @sampoh0523  has convinced me that this function should be a *no-op*.  It has to exist and be registered, so the validation function can detect it is validating a request, but this function should never reject an instance.

chrisinmtown:
hi @whoseoyster do you have time to update your PR? When I run tox on this PR I see these test failures:
```
FAILED tests/test_json_validation.py::test_readonly[swagger.yaml-FlaskApp] - assert 400 == 200
FAILED tests/test_json_validation.py::test_readonly[openapi.yaml-FlaskApp] - assert 400 == 200
FAILED tests/test_json_validation.py::test_readonly[openapi.yaml-AsyncApp] - assert 400 == 200
FAILED tests/test_json_validation.py::test_readonly[swagger.yaml-AsyncApp] - assert 400 == 200
FAILED tests/decorators/test_validation.py::test_invalid_type - AssertionError: assert '20 is not of...ance:
```
The `test_readonly` test POSTs a request to `/user` with a `user_id` field in the body; the schema has that marked `readOnly`, and the new code rejects it.

chrisinmtown:
Just for the record, here's a summary of the difference from V2 to V3.  In V2, one JSON schema test checks that a request with a  `readOnly` field is *rejected*. In V3, that same test checks that a request with a `readOnly` field is *accepted*. So it's not possible to copy/reuse the V2 code verbatim in V3.

chrisinmtown:
@whoseoyster please consider this revision to your PR. It changes the `validate_readOnly` function and adds tests for a required/readOnly field. On GET, check that Connexion *rejects* a response without the field. On POST, check that Connexion *accepts* a request without the field.
```
diff --git a/connexion/json_schema.py b/connexion/json_schema.py
index 47b6ab5..7bee3df 100644
--- a/connexion/json_schema.py
+++ b/connexion/json_schema.py
@@ -148,7 +148,15 @@ def validate_required(validator, required, instance, schema):
 
 
 def validate_readOnly(validator, wo, instance, schema):
-    yield ValidationError("Property is read-only")
+    """
+    The OpenAPI specification states:
+    > [readOnly] means that it MAY be sent as part of a response
+    > but SHOULD NOT be sent as part of the request.
+    Note the phrase is "SHOULD NOT" (not "MUST NOT").
+    Presence of this no-op function indicates to `validate_required`
+    that the instance being validated is a request.
+    """
+    pass
 
 
 def validate_writeOnly(validator, wo, instance, schema):
diff --git a/tests/fakeapi/hello/__init__.py b/tests/fakeapi/hello/__init__.py
index b6d6c00..5351f19 100644
--- a/tests/fakeapi/hello/__init__.py
+++ b/tests/fakeapi/hello/__init__.py
@@ -634,6 +634,10 @@ def get_user():
     return {"user_id": 7, "name": "max"}
 
 
+def get_user_without_userid():
+    return {"name": "max"}
+
+
 def get_user_with_password():
     return {"user_id": 7, "name": "max", "password": "5678"}
 
@@ -644,6 +648,10 @@ def post_user(body):
     return body
 
 
+def post_user_without_userid(body):
+    return post_user(body)
+
+
 def post_multipart_form(body):
     x = body["x"]
     x["name"] += "-reply"
diff --git a/tests/fixtures/json_validation/openapi.yaml b/tests/fixtures/json_validation/openapi.yaml
index c336eb0..59595e4 100644
--- a/tests/fixtures/json_validation/openapi.yaml
+++ b/tests/fixtures/json_validation/openapi.yaml
@@ -22,6 +22,20 @@ components:
         password:
           type: string
           writeOnly: true
+    UserWithRequiredReadOnly:
+      type: object
+      required:
+      - user_id
+      - name
+      properties:
+        user_id:
+          type: integer
+          readOnly: true
+        name:
+          type: string
+        password:
+          type: string
+          writeOnly: true
     X:
       type: object
       properties:
@@ -70,6 +84,32 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/User'
+
+  /user_without_userid:
+    get:
+      operationId: fakeapi.hello.get_user_without_userid
+      responses:
+        200:
+          description: User object
+          content:
+              schema:
+                $ref: '#/components/schemas/UserWithRequiredReadOnly'
+    post:
+      operationId: fakeapi.hello.post_user_without_userid
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UserWithRequiredReadOnly'
+      responses:
+        200:
+          description: User object
+          content:
+              schema:
+                $ref: '#/components/schemas/UserWithRequiredReadOnly'
+
   /user_with_password:
     get:
       operationId: fakeapi.hello.get_user_with_password
diff --git a/tests/fixtures/json_validation/swagger.yaml b/tests/fixtures/json_validation/swagger.yaml
index 0f9bcb5..355272a 100644
--- a/tests/fixtures/json_validation/swagger.yaml
+++ b/tests/fixtures/json_validation/swagger.yaml
@@ -20,6 +20,20 @@ definitions:
       password:
         type: string
         x-writeOnly: true
+  UserWithRequiredReadOnly:
+    type: object
+    required:
+    - user_id
+    - name
+    properties:
+      user_id:
+        type: integer
+        readOnly: true
+      name:
+        type: string
+      password:
+        type: string
+        x-writeOnly: true
 
 paths:
   /minlength:
@@ -59,6 +73,29 @@ paths:
           description: User object
           schema:
             $ref: '#/definitions/User'
+
+  /user_without_userid:
+    get:
+      operationId: fakeapi.hello.get_user_without_userid
+      responses:
+        200:
+          description: User object
+          schema:
+            $ref: '#/definitions/UserWithRequiredReadOnly'
+    post:
+      operationId: fakeapi.hello.post_user_without_userid
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/UserWithRequiredReadOnly'
+      responses:
+        200:
+          description: User object
+          schema:
+            $ref: '#/definitions/UserWithRequiredReadOnly'
+
   /user_with_password:
     get:
       operationId: fakeapi.hello.get_user_with_password
diff --git a/tests/test_json_validation.py b/tests/test_json_validation.py
index df613e1..024dc50 100644
--- a/tests/test_json_validation.py
+++ b/tests/test_json_validation.py
@@ -77,6 +77,20 @@ def test_readonly(json_validation_spec_dir, spec, app_class):
     )
     assert res.status_code == 200
 
+    # reject a response missing a required/readOnly field
+    res = app_client.get("/v1.0/user_without_userid")
+    assert res.status_code == 500
+    assert res.json()["detail"].startswith(
+        "Response body does not conform to specification"
+    )
+
+    # accept a request missing a required/readOnly field
+    res = app_client.post(
+        "/v1.0/user_without_userid",
+        json={"name": "max", "password": "1234"},
+    )
+    assert res.status_code == 200
+
 
 def test_writeonly(json_validation_spec_dir, spec, app_class):
     app = build_app_from_fixture(
```

connexion/json_schema.py
```diff
@@ -128,6 +128,29 @@ def nullable_validation_fn(validator, to_validate, instance, schema):
...
+
+
 def validate_writeOnly(validator, wo, instance, schema):
     yield ValidationError("Property is write-only")
```
chrisinmtown, line 168: Extending the scope here slightly, for `writeOnly` the OpenAPI spec section https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-19 says "MAY be sent as part of a request but SHOULD NOT be sent as part of the response." Well, this test case checks that Connexion rejects/fails a response that contains a `writeOnly` field. I don't much like it, but to conform to the spec, apparently Connexion should *allow* a response with such a field. The V2 and V3 branches agree on this point, for whatever that is worth.  @sampoh0523 would you please comment?

chrisinmtown:
@whoseoyster I sure was hoping to hear from you. If you cannot find time to update this PR, I can submit a new PR with the essential changes. Please comment.

whoseoyster:
@chrisinmtown added your changes here

connexion/json_schema.py
sampoh0523: Sorry, I did not see this before.

I do not have strong opinions about `writeOnly`, because the code using Connexion has control over its responses, unlike with `readOnly` where it has no control over the requests it receives.

chrisinmtown:
Greetings, I really was hoping the maintainers would comment on the direction they prefer here.  After thinking about this a while, I realize now that I just made the situation worse and I apologize to @whoseoyster. You proposed changes in this PR to make V3 work the same as V2. Nobody was complaining about the treatment of readOnly/writeOnly in V2. So I now think that your original changes should have been approved, merged and released. If that happens, if V2 and V3 behave the same, we all can discuss how to evolve the behavior to be more compliant with the JSON schema requirements, such as actually allowing (as strange as it seems to me) an input message that contains fields marked "readOnly".

Issue #1557, opened 2022-06-22
Take advantage of type hints


aholmes:
### Description

Type hints allow for safer code and, in some cases, self-documenting code. Connexion currently lacks type hints, which hurts its usability. In an effort to avoid a debate or argument over whether static typing is advantageous in a code base, I ask to consider type hints for the purpose of making the API more discoverable and simpler to use, and to enable type safety within code bases that do care about it.

As a simple example, consider the following. Here, we're calling `add_api` without a guarantee that 1. what we're using as parameters are correct and 2. that the documentation is up-to-date with the implementation.
![image](https://user-images.githubusercontent.com/277061/175100926-bdf9ce90-6144-4eaf-aa05-8af005043c6b.png)

Of note, the static types are already documented in the docblock comment. Using these types as Python type hints instead tends to require less effort to maintain (tooling helps with this), and goes a lot further toward the capabilities of tooling to use those type hints. In fact, although this is inconsequential here, the order of parameters in the docblock is inconsistent with the order of arguments for the method parameters.

Surprisingly, despite the docblocks, my tooling (pyright/pylance, vs code) wasn't able to offer up the parameters I was typing.
![image](https://user-images.githubusercontent.com/277061/175101393-36ecd977-66ae-4edd-a0db-2a9fb96c1790.png)

### Expected behaviour

Ideally, type hints would grant our tooling the ability to produce something like this.

![image](https://user-images.githubusercontent.com/277061/175108900-0d5ff0e3-6eb5-4f58-b2a3-f19667fe2abe.png)

![image](https://user-images.githubusercontent.com/277061/175108979-4fac092c-b487-4ae1-bd7f-d48a7f4e9656.png)

The change that enabled this was just to use type hints with the method parameters.

```python
    def add_api(
        self,
        specification: pathlib.Path | str | Dict[Any, Any],
        base_path: str | None = ...,
        arguments: Dict[Any, Any] | None = ...,
        auth_all_paths: bool = ...,
        validate_responses: bool = ...,
        strict_validation: bool = ...,
        resolver: Resolver | types.FunctionType = ...,
        resolver_error: int | None = ...,
        pythonic_params: bool = ...,
        pass_context_arg_name: str | None = ...,
        options: Dict[Any, Any] = ...,
        validator_map: Dict[Any, Any] = ...,
    ) -> AbstractAPI: ...
```

### Actual behaviour

Static type guarantees in Connexion are spurious and sometimes problematic for applications that attempt to remain type safe.

[...]

PR #1947, opened 2024-07-10
add py.typed marker for typing

fridex:

 - see https://peps.python.org/pep-0561/
 - related #1111
 - related #1557

CC @RobbeSneyders

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,9 @@ classifiers = [
 [tool.poetry.scripts]
 connexion = 'connexion.cli:main'
 
+[tool.setuptools.package-data]
+connexion = ["py.typed"]
+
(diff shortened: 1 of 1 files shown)


mharding-hpe (approved):

RobbeSneyders:
Thanks @fridex!

There were some doubts previously if this is the best way to go if our types are not perfect, see [this comment](https://github.com/spec-first/connexion/issues/1674#issuecomment-1506696647). Do you have a view on that?

RobbeSneyders:
Can you expand a bit?

Note that [we are running mypy on every commit](https://github.com/spec-first/connexion/blob/c89e2d2fff0e2287ddb67d92aed95e0536785d1b/.pre-commit-config.yaml#L45) already.

PR #1955, opened 2024-07-26
fix: update scope when content-length changes

### LINKED ISSUE #1954: using validator DefaultsJSONRequestBodyValidator with more content-length breaks json data
### Description

I ran into this issue when applying the DefaultsJSONRequestBodyValidator on an openapi spec, and where a default was to a body request, and it increaed the length. I noticed that my json body was getting dropped and not making it into the final operation (via the body name).

It seems to me the issue is related to this line of code:

[DefaultsJSONRequestBodyValidator](https://github.com/spec-first/connexion/blob/main/connexion/validators/abstract.py#L79)

I suspect the actual scope needs to be updated there in place. WIth the copy() being used, its not clear how that would ever actually update the scope. I will have a pull request for this shortly that corrects this behavior.

### Expected behaviour

I expect that json body defaults that increase the content-length will still be applied and function correctly.

### Additional info:

Output of the commands:

- `python --version`

mfmarche:
The actual arg (scope), needs to be updated when the content-length changes.

Fixes #1954 

 -
 -
 -

### DIFF
--- a/connexion/middleware/request_validation.py
+++ b/connexion/middleware/request_validation.py
@@ -137,7 +137,7 @@ class RequestValidationOperation:
                         self._operation.parameters, self._operation.body_definition()
                     ),
                 )
-                receive = await validator.wrap_receive(receive, scope=scope)
+                receive, scope = await validator.wrap_receive(receive, scope=scope)
 
         await self.next_app(scope, receive, send)
 
--- a/connexion/validators/abstract.py
+++ b/connexion/validators/abstract.py
@@ -64,14 +64,16 @@ class AbstractRequestBodyValidator:
         :raises: :class:`connexion.exceptions.BadRequestProblem`
         """
 
-    def _insert_body(self, receive: Receive, *, body: t.Any, scope: Scope) -> Receive:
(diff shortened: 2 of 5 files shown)


RobbeSneyders:
Thanks @mfmarche!

You are on to something. It indeed seems like the updated header does not have an effect. However, the scope does need to be copied whenever it is modified in middleware, so the changes in the scope are only propagated downstream.

I think the best solution is to update this method to return the modified scope as well.

Also, it would be great if you could add a test for this behavior.

mfmarche:
Thanks @RobbeSneyders . I took your comments and added appropriate test, and now return the content length. The test prior to the scope change gave a 500, which was similiar to what I saw.

RobbeSneyders:
Thanks @mfmarche, the change looks good to me. The pipeline is failing on pre-commit though. You can fix this by running `pre-commit run --all-files` and committing the results.

To have it run automatically in the future, you can run `pre-commit install`.

RobbeSneyders (approved):
Thanks @mfmarche!

PR #1957, opened 2024-08-01
Now will distinguish between status codes the expected inferred type

### LINKED ISSUE #1956: Failing to infer content when application/problem+json is present in another response code
### Description

When having 2 reponse types one is problem+json is unable to infer response

### Expected behaviour

Should be able to infer that is a dictionary with no header and asume is json response or just check 200 response is always application/json

### Actual behaviour

if another response like 401 is present with a different content type like application/problem+json should still be able to infer is a json

### Steps to reproduce
as API
``` yaml
  /potato/advanced-configuration:
    get:
      summary: xxx
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/xxx'
        '401':
          description: Returned if the calling user does not have permission to perform the operation.
          content:
            application/problem+json:
              schema:
                $ref: ../common/models/error.v1.yaml
      operationId: configuration.get_configuration
```

``` python
def get_configuration():
    try:
        response = cb_conf.read_conf() or {}
        status_code = 200

    except Exception as ex:
        logging.error(f"Error getting configuration: {ex}")
        response, status_code = unexpected_exception(ex), 400
    print(response, status_code)
    return response, status_code
```

### Additional info:

`{"time": "2024-07-31T10:51:30Z", "format_version": 1, "level": "ERROR", "message": "NonConformingResponseHeaders(status_code=500, detail='Multiple response content types are defined in the operation spec, but the handler response did not specify which one to return.')", "microservice": "cbclient", "funcName": "problem_handler", "threadName": "AnyIO worker thread"}`

Output of the commands:

- `python --version` 3.11
- `pip show connexion | grep "^Version\:"` 3.1

carloscbl:
Fixes #1956 .

 - Now instead merging 2 different responses as available options, will check for the status code the available options, after that if you still have multiple content types it will raise the error, and users should specify in headers the correct one
 -
 -

### DIFF
--- a/connexion/decorators/response.py
+++ b/connexion/decorators/response.py
@@ -27,7 +27,7 @@ class BaseResponseDecorator:
 
     def build_framework_response(self, handler_response):
         data, status_code, headers = self._unpack_handler_response(handler_response)
-        content_type = self._infer_content_type(data, headers)
+        content_type = self._infer_content_type(data, status_code, headers)
         if not self.framework.is_framework_response(data):
             data = self._serialize_data(data, content_type=content_type)
             status_code = status_code or self._infer_status_code(data)
@@ -37,7 +37,7 @@ class BaseResponseDecorator:
         )
 
     @staticmethod
-    def _infer_content_type(data: t.Any, headers: dict) -> t.Optional[str]:
+    def _infer_content_type(data: t.Any, status_code: int, headers: dict) -> t.Optional[str]:
         """Infer the response content type from the returned data, headers and operation spec.
 
         :param data: Response data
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Thanks @carloscbl!

Tests are failing though. Not sure if this is due to an issue with your code, or because the failing tests depend on the old behavior.

kaktusztea:
Hi, I just found this PR.
I had the same problem and would be great if you could merge this fix somehow.
Any chance?

chrisinmtown:
The test outputs that showed a failure have expired. Please rebase etc. and push, then hopefully the Github workflow will run again here. Thanks.

chrisinmtown:
@carloscbl maybe you'd like to close this PR, since @Boxap opened #2100 with the same content?

PR #1990, opened 2024-11-08
Change response codes to strings (#1112)

### LINKED ISSUE #1112: Response keys should be strings
There are a handful of yaml schemas with something like:
https://github.com/zalando/connexion/blob/f31ac615bfc8a3c696c5ebe53f3c57e5234ff6f8/examples/openapi3/helloworld/openapi/helloworld-api.yaml#L15-L16
pyyaml interprets these keys as integers, while the OpenAPI specification requires that they be strings:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#format

The above should be replaced with
```yaml
      responses:
        '200':
```
[openapi-spec-validator](https://github.com/p1c2u/openapi-spec-validator) does not validate this, but instead coerces `int` keys to `str`.

peso:
Fixes #1112  .

 - Convert response keys from ints to strings

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ The operation described in your specification is automatically linked to your Py
        post:
          operationId: run.post_greeting
          responses:
-           200:
+           '200':
              content:
                text/plain:
                  schema:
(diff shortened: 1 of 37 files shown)


RobbeSneyders (approved):
Thanks @peso!

PR #1992, opened 2024-11-09
Upgrade dependencies on v2 banch

### LINKED ISSUE #1969: Upgrade v2 dependencies to address recent security vulnerabilities like CVE-2024-34069 in werkzeug
Please patch version 2 security issues for us long-time loyal users! I really do want to upgrade to Connexion v3 but that path is long & windy.  Current security vulnerabilities are critical.

### Description

Connexion 2.x with Flask is flagged for vulnerabilities in the `werkzeug` library version < 3.0.6:

* Werkzeug debugger vulnerable to remote execution when interacting with attacker controlled domain, CVE-2024-34069, https://github.com/advisories/GHSA-2g68-c3qc-8985
* Werkzeug DoS: High resource usage when parsing multipart/form-data containing a large part with CR/LF character at the beginning, CVE-2023-46136, https://github.com/advisories/GHSA-hrfv-mqp8-q5rw
* Applications using Werkzeug to parse multipart/form-data requests are vulnerable to resource exhaustion. A specially crafted form body can bypass the Request.max_form_memory_size setting, CVE-2024-49767, https://github.com/advisories/GHSA-q34m-jh98-gwm2

The version that addresses al problems is 3.0.6, but that is a disruptive change. A version that addresses the 2023 CVS is 2.3.8.

### Expected behaviour

Security scans of Connexion/Flask dependencies yield no CVE numbers

### Actual behaviour

Security scans of Connexion/Flask dependencies yield the issues listed above

### Steps to reproduce

Let the Github security check run on a requirements.txt file

### Additional info:

- Python 3.11
- Connexion Version: 2.14.2

Just for the record, here are the dependencies from the Connexion V2 branch, which is what we're running from PyPI as 2.14.2:
```
install_requires = [
    'clickclick>=1.2,<21',
    'jsonschema>=2.5.1,<5',
    'PyYAML>=5.1,<7',
    'requests>=2.9.1,<3',
    'inflection>=0.3.1,<0.6',
    'werkzeug>=1.0,<2.3',
    'importlib-metadata>=1 ; python_version<"3.8"',
    'packaging>=20',
]
```

mfmarche:
Fixes #1969

    The main goal was upgrading werkzeug for CVE-2024-34069.
    After switching to python 3.12, it proved more difficult with changes to
    setuptools, etc. I decided to pull the pyproject from the main, and
    utilize that, alone with updated dependencies. Small changes were needed
    in various api changes, notably:
    
    - flask change of request_ctx
    - swagger_ui_bundle version change, default_template_dir change
    - aiohttp middleware api slightly changed
    - flask json change, using flask.json.provider
    
    I believe these changes will have minimal impact to users, but the
    changes are likely breaking for some, specifically, the move to latest
    flask.


mfmarche:
I tried to fixup the readthedocs build, i'm not sure how to resolve the issue. It could be a conflicting dependency? Is this build job required on this branch?

README.md
RobbeSneyders, line 1: We should keep the old README on this branch, so the information and examples match the code.

Dockerfile3.9
RobbeSneyders, line 1: Where is this file used?

pyproject.toml
RobbeSneyders, line 1: I assume we can remove `setup.cfg` and `setup.py` now?

tox.ini
```diff
@@ -1,68 +1,41 @@
...
 
 [gh-actions]
 python =
-    3.7: py37-min,py37-pypi
-    3.8: py38-min,py38-pypi
-    3.9: py39-min,py39-pypi,flake8,isort-check,isort-check-examples,isort-check-tests,mypy
+    3.9: py39-pypi
+    3.10: py310-pypi
+    3.11: py311-pypi,pre-commit
+    3.12: py312-pypi
```
RobbeSneyders, line 20: The [github actions workflow](https://github.com/spec-first/connexion/blob/3d67f16075349d7137c2848aef9fa3bd35e3a268/.github/workflows/pipeline.yml) should be updated to match this.

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+asgiref = ">= 3.4"
+httpx = ">= 0.23"
+inflection = ">= 0.3.1"
+jsonschema = ">=4.17.3"
+Jinja2 = ">= 3.0.0"
+python-multipart = ">= 0.0.5"
+PyYAML = ">= 5.1"
+requests = ">= 2.27"
+typing-extensions = ">= 4.6.1"
+werkzeug = ">= 2.2.1"
```
RobbeSneyders, line 57: Let's bump this to a version with the CVE fixes, since this was the main reason for this PR.
3.1.3 is the version used in the 3.9 workflow with passing tests.

RobbeSneyders (commented):
Thanks @mfmarche!

I haven't gone through the code changes in detail yet, but the passing tests with limited changes look promising! 

The docs pipeline is configured in this repo in `.readthedocs.yaml`, the dependencies are defined in `pyproject.toml`. The pipeline runs on ReadTheDocs. I suspect the issue is due to a mismatch in `sphinx` and `sphinx-autoapi` versions ([logs](https://readthedocs.org/projects/connexion/builds/26227804/)).

I don't want to break people's projects relying on Connexion 2 by releasing breaking changes. However I am open to releasing this fix under the [`connexion2`](https://pypi.org/project/connexion2/#description) pypi name instead.

mfmarche:
thanks @RobbeSneyders for the review. I addressed all issues I believe.

ykharko:
Hey guys, sorry for chasing. What's left here? Is there a chance to get it merged and released? I believe many people are waiting for werkzeug updates because of CVEs.

Are thare any blockers here and is there a chance to release it in the nearest future?

mfmarche:
Hi @ykharko, I believe all issues were resolved, just waiting for final review and hopefully it can be released. thanks.

noirbee:
Hi @mfmarche , thanks a lot for this, started working on this for the same reasons but your work is much further along, especially in terms of CI / packaging etc.

I've started testing our code base against your branch, and the first breakage I've come across is the fact that the `FlaskJSONEncoder` now inherits from Python's own default `JSONEncoder`; i.e. it does not end up calling `flask.json.provider._default(o)` after falling through the few types connexion explicitly handles.

Arguably the encoder could be removed as I'm not sure it was intended to be part of connexion's actual module API, unfortunately it's used by other projects, notably [openapi-code-generator's flask template](https://github.com/OpenAPITools/openapi-generator/blob/8f4cfa395507be3bb558610c9697b956f8955c62/modules/openapi-generator/src/main/resources/python-flask/encoder.mustache#L1), which we also use.

Would you consider the following patch against your branch to keep backwards compatibility ? 

```diff
diff --git a/connexion/apps/flask_app.py b/connexion/apps/flask_app.py
index e9efaef..9814a89 100644
--- a/connexion/apps/flask_app.py
+++ b/connexion/apps/flask_app.py
@@ -153,10 +153,8 @@ class FlaskApp(AbstractApp):


 class FlaskJSONProvider(DefaultJSONProvider):
-    def __init__(self, app):
-        super().__init__(app)
-
-    def default(self, o):
+    @classmethod
+    def default(cls, o):
         if isinstance(o, datetime.datetime):
             if o.tzinfo:
                 # eg: '2015-09-25T23:14:42.588601+00:00'
@@ -177,22 +175,7 @@ class FlaskJSONProvider(DefaultJSONProvider):

 class FlaskJSONEncoder(json.JSONEncoder):
     def default(self, o):
-        if isinstance(o, datetime.datetime):
-            if o.tzinfo:
-                # eg: '2015-09-25T23:14:42.588601+00:00'
-                return o.isoformat('T')
-            else:
-                # No timezone present - assume UTC.
-                # eg: '2015-09-25T23:14:42.588601Z'
-                return o.isoformat('T') + 'Z'
-
-        if isinstance(o, datetime.date):
-            return o.isoformat()
-
-        if isinstance(o, Decimal):
-            return float(o)
-
-        return json.JSONEncoder.default(self, o)
+        return FlaskJSONProvider.default(o)


 class NumberConverter(werkzeug.routing.BaseConverter):

```

(I can also open a separate PR against your branch if you'd rather review it separately)

mfmarche:
thanks @noirbee for providing the feedback. I have updated the MR with your suggested changes.

mfmarche:
If I'm missing something else @RobbeSneyders please let me know. I wasn't sure how you would like to publish to connexion2, if you want to change the pyproject.toml? Thank you.

hehe7318:
Hi team, hi @mfmarche, hi @RobbeSneyders, I am from Amazon AWS ParallelCluster team. We are having trouble when trying to upgrade werkzeug for CVE-2024-34069. If we use werkzeug~=3.0 with connexion[flask]~=3.0, many issues will arise. Currently, we imported connexion v2 codes into our codebase and maintain it ourselves. So that we can upgrade werkzeug to 3.0. But it's only a temporary workaround. If you can merge this change, I think it will help a lot! Do you have a estimate time to merge this change?
Or, if you can provide some guidance to us to migrate to connexion 3.0, that will be also great. Can you confirm if we can still use werkzeug 3.0 with connexion 3.0?

gmarciani:
Hi team,

what's the timeline to merge this PR and release a new 2.x version?
Thanks

sebastien-boulle:
Hi Team,

It would be very nice for security reason to merge it and release it :)
Especially because it's also blocking Werkzeug security update. 

Any timeline would be appreciate.
Thanks

advance512:
Hello @RobbeSneyders @chrisinmtown , could we proceed with merging this in and releasing a new v2 version?

It is important and apparently a mere approving review away to patch the security issues for us long-time loyal users. 
We really do want to upgrade to Connexion v3 but that path is long & windy. Current security vulnerabilities are critical.

Is there anything we could do to help with getting this over the finish line?

Thank you so much @mfmarche for your hard work here.

chrisinmtown:
I support merging this too! Please note I'm a fellow user and community member, not a maintainer.

chrisinmtown:
@mfmarche  would you please suggest the most obvious way to build/test an existing project that's based on Connexion v2 with the version in this PR? I'd like to try that for my two critical projects. FWIW I am blocked on the upgrade to V3 due to bugs that remain unresolved, like the problem with processing `allOf` in OpenAPI specs. Thanks in advance.

mfmarche:
Hi @chrisinmtown , you can use my fork on branch "upgrade_deps" to test it. For example, in pyproject.toml:

connexion = { git = "https://github.com/mfmarche/connexion", branch="upgrade_deps"}

I am also blocked on the upgrade. The move to uvicorn has implications for performance, since we heavily used gthreads (greenlets), so to achieve similiar performance, the WSGI app really needs to be fully async now.

tox.ini
chrisinmtown: This PR now includes an update to the GitHub workflow for pythons 3.9 .. 3.12

.github/workflows/pipeline.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.9', '3.10', '3.11', '3.12']
```
chrisinmtown, line 12: FWIW, since you are updating the list, I would include 3.13

connexion/apps/flask_app.py
```diff
@@ -150,24 +152,30 @@ def run(self,
...
 
         if isinstance(o, Decimal):
             return float(o)
 
-        return json.JSONEncoder.default(self, o)
+        return super().default(o)
+
+
+class FlaskJSONEncoder(json.JSONEncoder):
+    def default(self, o):
```
chrisinmtown, line 177: Would you please add a comment explaining why this tiny class must exist and override the `default` method?

tox.ini
```diff
@@ -1,68 +1,41 @@
...
-ignore_outcome=true
-commands=mypy --exclude='examples/' connexion tests
+    poetry
+allowlist_externals=
+    cp
+    sed
+    mv
+commands=
+    min: cp pyproject.toml .pyproject.toml
+    min: sed -i -E 's/"(\^|~|>=)([ 0-9])/"==\2/' pyproject.toml
```
chrisinmtown, line 32: Is the `testenv` still used? I ran tox and did not see these commands appear.  I was checking because this `sed`  line does not work on MacOS, please see PR #2047 for a fix. But, maybe these lines are obsolete? Sorry I'm not familiar enough with how poetry works.

chrisinmtown:
How should I run the tests? I ran `tox` and all tests passed, but I don't see any invocations of `mypy`, `ruff`, `isort`, `flake8` or other code format and style checkers.  That seems like a miss, but maybe I'm doing something wrong?

chrisinmtown:
> I don't see any invocations of `mypy`, `ruff`, `isort`, `flake8` or other code format and style checkers.

I see the 3.x branch has file `.pre-commit-config.yaml` which is missing here. Also the required lines at the end of file `tox.ini` are commented out here. I copied over the yaml file and changed the tox lines. Then all the checks run at the end, and the result is a bunch of format changes such as this:
```
diff --git a/tests/test_utils.py b/tests/test_utils.py
index 72c9e27..42e9921 100644
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -7,14 +7,14 @@ from connexion import utils
 
 
 def test_get_function_from_name():
-    function = utils.get_function_from_name('math.ceil')
+    function = utils.get_function_from_name("math.ceil")
     assert function == math.ceil
     assert function(2.7) == 3
```
Please consider how much of this you want to incorporate.  I tend to like the style standardization, especially in an open-source repo.

donbowman:
IMHO the best thing for the community would be to accept this PR onto the V2 branch, and publish as a new version on 2.x on connexion existing pypi, allowing those folks who are working on the major rework required for 3.x to work around the severe security issues that are currently pinned to v2.

Failing that, the next best alternative is to create a hard fork and keep it on the v2 compatibility, publish that to pypi as an alternate name. In which case we would need at least 2 maintainers who are independent of each other.

It would be nice to get some guidance from the maintainers as to which direction you prefer. If you are against the idea of updating the current v2 branch dependencies, then we can proceed with the hard fork. If its just a resourcing issue, is there something one of us can do to assist?

I believe @mfmarche and @chrisinmtown have done the work on the PR, so it really comes down to this question.

chrisinmtown:
> Hi @chrisinmtown , you can use my fork on branch "upgrade_deps" to test it. For example, in pyproject.toml:

I really want to test this so I can give comments here. FWIW my project still uses `setup.py` and `requirements.txt` files. When I enter the local version string `2.0.dev0` in either of those, tox attempts to fetch that from PyPI and of course fails. Stuck on this noob error. Maybe I'm not alone, for example https://github.com/tox-dev/tox/issues/2943   All suggestions welcome!

tox.ini
Ruwann: Indeed, there should ideally also be a `py39-min`.

The `min` is to run the tests against the minimum required versions of the dependencies, instead of the default of taking the latest allowed version. Then, we can adjust minimum versions accordingly.

mfmarche:
@Ruwann any comment on a plan that will allow this to be merged?

thanks.

.github/workflows/pipeline.yml
Ruwann, line 1: I think we also need to update the release Github Action as it still relies on `setup.py`. We should be able to reuse the one from v3 as well

tox.ini
chrisinmtown: I'm trying to find the min line you mention, am I blind ? I don't see any `pyxx-min` lines in the `tox.ini` file, especially not in `envlist`, that will trigger a test against the minimum dependencies.

.github/workflows/pipeline.yml
Ruwann: I don't see the `.pre-commit-comfig.yaml` file on your branch, which is why it is complaining. It indeed contains some formatting such as `black` to maintain consistency. It should work when you add that file (you can then also run the pre-commits yourself - most things should be auto-fixed, otherwise the error tells you what you should do).

chrisinmtown: > that will create a pile of file chages in this review, which is not ideal.

I also prefer small pull requests. Still, I don't think a batch of cosmetic, non-functional changes is that big a deal.

If you are not convinced, I volunteer to create a PR to this branch that makes all those cosmetic, non-functional changes.  If the maintainers accept that, then your PR to update dependencies could remain small-ish.

mfmarche: I can do it. Its more about maintainer expectations, as some projects are very strict about these sorts of things.

chrisinmtown:
Very nice @mfmarche the pipeline checks *all* pass!

Ruwann:
Thanks a lot for the efforts @mfmarche and your help @chrisinmtown . 

I have create a new release with these fixes as version `2.15.0rc1`. This is a pre-release, so you will need to explicitly target that version or include the `--pre` flag for pip. That way, it can already be used but it won't break existing v2 users. We can still always decide to release it as a full release later on.

chrisinmtown:
Thank you @Ruwann for moving this forward. I'm not deeply familiar with the PyPI pre-release feature. I checked https://pypi.org/project/connexion/#history and I see a new and shiny `2.15.0.rc1' there I will try to install that next.

Ruwann:
> Thank you @Ruwann for moving this forward. I'm not deeply familiar with the PyPI pre-release feature. I checked https://pypi.org/project/connexion/#history and I see a new and shiny `2.15.0.rc1' there I will try to install that next.

The pre-release version allows creating a new version on PyPI, but you have to be explicit if you want to install it. So just doing `pip install "connexion<3"` won't consider that version. You can install it with `pip install connexion==2.15.0rc1` or `pip install --pre "connexion<3"`

chrisinmtown:
I'm really happy to report a success with this RC. I added this line to `setup.py` in my Connexion v2 projects:
```
        "connexion[swagger-ui]==2.15.0rc1",
```
Then I ran `pip-compile` followed by `tox` and both suites of tox tests passed nicely!  Please say, what is the path for promoting the release candidate to a first-class release?

potiuk:
Thanks for all the efforts on that one, that will help us to address a  long-time standing issue with werkzeug being blocked for Airflow 2.

I will do some tests with Airflow 2.11 with this RC for Airflow 2 and will report here if I find any issues. 

Are there already plans to release the final version?

potiuk:
Question: I tried to attempt to upgrade Airflow 2 to use the 2.0.15rc1 - and one of the problems I see is that `flask` extra has `flask>=3` . How "strict" that limitation is ? We are also using Flask App Builder that requires Flask <3, so if that is a hard requirement, that would make it more difficult for us to upgrade.

chrisinmtown:
Greetings to @ykharko @noirbee @hehe7318 @gmarciani @sebastien-boulle @advance512 @donbowman @mihirverma7781 @rickeydbrown @potiuk you all asked about this PR. Now that @Ruwann has merged it and created a pre-release candidate, please test your code against the new pre-release and come back here to comment/vote for a final release. Hope we the community can provide evidence that the new 2.0.15 release candidate is ready for general use. Thanks!

chrisinmtown:
@potiuk wrote:
> I see is that `flask` extra has `flask>=3` . How "strict" that limitation is ?

For what it's worth, I checked PyPI where I see that the last Flask v2 release happened in August 2023, about 22 months ago. That's quite a while. So, I believe the upgrade to Flask v3 is essential for using current dependencies on this branch.

potiuk:
> @potiuk wrote:
> 
> > I see is that `flask` extra has `flask>=3` . How "strict" that limitation is ?
> 
> For what it's worth, I checked PyPI where I see that the last Flask v2 release happened in August 2023, about 22 months ago. That's quite a while. So, I believe the upgrade to Flask v3 is essential for using current dependencies on this branch.

The problem is that other dependencies might be incompatible. Apache Airflow 2 has Flask Application Builder as required dependency, and even in the latest (in-progress) version of FAB 5 - Flask is still <3. I raised question about it, but maybe it is easier for connexion to **still** support Flask 2 than to convert FAB to use Flask 3 ?

https://github.com/dpgaspar/Flask-AppBuilder/pull/2241#issuecomment-2976674169

So If there is no good reason to change it and concrete features that caused flask to be >= 3, it would be great if we can lower the limit. That would be my feedback from "testing".

> please test your code against the new pre-release and come back here to comment/vote for a final release.

The problem is that without it - I can't even test it properly in Airflow, because in order to test Connexion, I have to remove FAB but then Airlfow does not work without both FAB and Connexion. I attempted to do it -> but the only way I could make it  install https://github.com/apache/airflow/pull/51681/files#diff-d56d1b58c0ca28bdddd10c2cee2ab3cb4312766ea5de51f210dbc15a0fea42faL425 - I had to remove `[flask]` extra from connexion, but this makes Airflow 2 fail because it also never worked with Flask 3 (In Airflow 3 we removed flask as required dependency alongside connexion, so we have no problem with it, it's only for old versions that are now flagged with critical werkzeug issues - this the main reason why we might use that old "connexion 2" release - just to allow us to bump Werkzeug. But if it also means Flask 3 upgrade, this is far more involved and will take more time (providing of course FAB will also support Flask 3).

potiuk:
It would be **WAY** easier for us if we knew connexion 2.0.15 can work with Flask 2.

mik-laj:
Hello @chrisinmtown 

First of all, thank you for your hard work on `connexion` — lifting the werkzeug constraint is a huge and long-awaited step forward for many of us in the community, especially those maintaining production systems that rely on older Flask-compatible stacks like Apache Airflow 2.x.

However, I’d like to respectfully ask you to reconsider the hard `flask>=3.0.0` requirement in the `[flask]` extra. I’d like to propose the following :

* First, release a version (e.g. `2.15.x`) that focuses solely on lifting the Werkzeug restriction and retains Flask 2.x compatibility (e.g. `flask>=2.2.0,<4`). This would allow projects like Airflow to resolve security issues like CVE-2024-34069 quickly and safely, without requiring a major framework migration.

* Then, in a subsequent release (e.g. `2.16.x`), you can formally drop Flask 2 support and require Flask 3 going forward.

This phased approach would bring immediate security benefits to users on legacy stacks, while keeping your roadmap to modernize Connexion intact. It also aligns with Airflow 2.x’s timeline, which enters limited maintenance in October 2025 and reaches EOL in April 2026.

chrisinmtown:
@mik-laj all credit for the V2 branch goes to @mfmarche I'm just trying keep up with him!

You suggested downgrading the version pin to `flask>=2.2.0`. Is there something special about the 2.2.x family?  I see the latest Flask 2.x release is 2.3.3, and that's from 2023. So I am trying to test the V2 branch with a revised version pin of `>=2.3.3,<4`. This may also help to revive the `min` version tests on the V2 branch, but that's a whole other topic.

mik-laj:
`connexion==2.14.2` (latest 2.x release) has the following constraints: [`flask>=1.0.4,<2.3`](https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/setup.py#L36), so the use of version `>=2.2.0` would allow the user to gradual update of dependencies. Otherwise, the `connextion` update to the newer version will also require Flask update to the newer version.

potiuk:
There is nothing "special" about "2.2" other than dependencies of connexion 2.14.2 that are themselves limited to < 2.3.

When you look at dependencies of connexion 2.14.2 you will see this:

```
Requires-Dist: flask (<2.3,>=1.0.4)
```

And that means that if connexion 2.15.0 has `flask>=2.3` - that there is no flask version that would be working fine for both 2.14.2 and 2.15.0. This is not a problem on its own, and for example for Airlfow that should not be a problem. And  in`apache-airflow` itself we have other dependencies with those limitations (in Airflow 2.11), so we should be able to upgrade to 2.3

```
│   ├── Flask [required: >=1.0.4,<2.3, installed: 2.2.5] -> coming from connexion 2.14.2
│   └── Flask [required: Any, installed: 2.2.5]
│   ├── Flask [required: >=2.2, installed: 2.2.5]
│   ├── Flask [required: Any, installed: 2.2.5]
│   ├── Flask [required: >=2,<3.0.0, installed: 2.2.5]
│   │   ├── Flask [required: Any, installed: 2.2.5]
│   │   ├── Flask [required: >=2, installed: 2.2.5]
│   │   ├── Flask [required: >=1.0.4, installed: 2.2.5]
│   │   ├── Flask [required: >=0.10, installed: 2.2.5]
│   │   ├── Flask [required: Any, installed: 2.2.5]
│   │   ├── Flask [required: >=2.0,<4.0, installed: 2.2.5]
│   ├── Flask [required: >=1.0.4, installed: 2.2.5]
├── Flask [required: Any, installed: 2.2.5]
```

So yeah - if flask is 2.3+ I would be super-happy!

potiuk:
> connexion==2.14.2 (latest 2.x release) has the following constraints: [flask>=1.0.4,<2.3](https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/setup.py#L36), so the use of version >=2.2.0 would allow the user to gradual update of dependencies. Otherwise, the connextion update to the newer version will also require Flask update to the newer version.

Yeah  I agree with @mik-laj - those kind of "jumps" are always "suspicious" - and if possible I would also avoid it, but in this case, I see no **huge** problem and since 2.15.0 is so long time after 2.14.2, if this makes things easier, 2.3+ is also "OK".

chrisinmtown:
If you don't mind I suggest we move this discussion about rolling back the Flask version pin to this new PR that I'm trying to cobble together: https://github.com/spec-first/connexion/pull/2058

Ruwann:
Thanks for the follow-up and detailed explanations, @potiuk @mik-laj . We've released a new pre-release version 2.15.0rc2 thanks to @chrisinmtown, which has `flask>= 2.3.3` so you should be able to test Airflow v2 against it. Does that work for you guys or do you need more changes from connexion v2 to make it work? As I haven't checked all the details for the differences in flask versions.


> > connexion==2.14.2 (latest 2.x release) has the following constraints: [flask>=1.0.4,<2.3](https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/setup.py#L36), so the use of version >=2.2.0 would allow the user to gradual update of dependencies. Otherwise, the connextion update to the newer version will also require Flask update to the newer version.
> 
> Yeah I agree with @mik-laj - those kind of "jumps" are always "suspicious" - and if possible I would also avoid it, but in this case, I see no **huge** problem and since 2.15.0 is so long time after 2.14.2, if this makes things easier, 2.3+ is also "OK".

Do you mean avoid going from `<2.3` to `>3`? What would be the ideal path in that case for users?

potiuk:
> Do you mean avoid going from <2.3 to >3? What would be the ideal path in that case for users?

Let me explain what I mean. It's likely not a problem in this case because I checked all other dependencies of ours and it seems that we should be able to upgrade to 2.3, but general problem is (when you have a lot of dependencies like airflow) is that it's better that subsequent versions of our dependencies have overlapping ranges between the versions rather than previous version having < x and next version having >= x.

In this case before connexion 2.14. ad < 2.3 ad 2.14 has  >= 2.3. Usually (at least this is what we try to do) when we bump such dependencies, we try to make smooth transition - i.e. if we were upper-bounded, we try to support at least some of the last dependency versions that was working before - because it might happen that other dependencies will need to be upgrade as well and it might be we won't be able to upgrade. 

For example:

* < 2.3 - means that latest version was 2.2.2 -> and this is the version that Airlfow 2 is tested with
* now, when you bump to >= 2.3 -> we **hope** connexion is the last one that was holding us from upgrading - the problem in this case would be that we would not even be able to upgrade to test the RC - because maybe some other dependency also had flask < 2.3 - in which case result would be conflicting dependencies. If we had the ranges in 2.14 and 2.15 overlapping and we were already on 2.2.2 of Flask, it would work. 

With non-overlapping ranges it would mean that we would have to also work with other dependencies to upgrade - **just** to be able to test 2.15rc2

But the good news is that this is NOT a problem. I tested it and that's the case indeed. I just built airflow image and it looks like all is fine and we god Werkzeug == 3.1.3 and we have non-conflicting requirements so **ALL LOOKS GOOD** - we will just have to work out any incompatibilities - potentialy and we should be good !

Thanks a lot for doing it - I am running the tests now in https://github.com/apache/airflow/pull/51681 and 🤞  that we can confirm everything working soon.

potiuk:
Ok. We have some failures but they are about `flask-session` upgrade which we knew we will have to fix -> and we are unblocked. I will keep you posted here.

PR #1993, opened 2024-11-10
Document how to install prerequisites in examples READMEs

### LINKED ISSUE #1987: Extend helloworld RST with pip install-dependency command
### Description

Provide instructions so a novice can start the Connexion v3 "helloworld" example on the first try.

### Expected behaviour

The example runs on the first try.

### Actual behaviour

It takes a few tries to install all the prerequisites.

### Steps to reproduce

`python hello.py`

### Proposed addition to `README.rst`

* Create a new virtual environment
* Run `pip install 'connexion[flask,swagger-ui,uvicorn]>=3.1.0'`

### Discussion :)

I realize this is fairly trivial. Just trying to save the next person a little time. 

And  if you don't mind another suggestion, you might consider renaming the directory to "helloworld_flask", to keep it parallel with "helloworld_async" and similar names.

Please comment. I don't want to waste your time with a PR if this looks silly to you.

chrisinmtown:
Add server url /openapi to JWT example
Add TODO to SQLAlchemy example, does not work with current versions
Correct name of module in v3 instructions
    
Fixes #1987

### DIFF
--- a/examples/jwt/requirements.txt
+++ b/examples/jwt/requirements.txt
@@ -1,3 +1,1 @@
-..[swagger-ui]
 python-jose[cryptography]
-Flask>=0.10.1
--- a/examples/sqlalchemy/README.rst
+++ b/examples/sqlalchemy/README.rst
@@ -2,14 +2,36 @@
 SQLAlchemy Example
 ==================
 
+.. note::
+
+    TODO: Update this example to work with recent (2024) versions of Python and SQLAlchemy.
+
 A simple example of how one might use SQLAlchemy as a backing store for a 
 Connexion based application.
 
-Running:
+Preparing
+---------
+
+Create a new virtual environment and install the required libraries
+with these commands:
 
 .. code-block:: bash
 
+    $ python -m venv my-venv
+    $ source my-venv/bin/activate
+    $ pip install 'connexion[flask,swagger-ui,uvicorn]>=3.1.0'
     $ pip install -r requirements.txt
+
+Running
+-------
+
+Launch the connexion server with this command:
+
+.. code-block:: bash
+
     $ python app.py
 
-Now open your browser and go to http://localhost:8080/ui/ to see the Swagger UI.
+Now open your browser and view the Swagger UI for these specification files:
+
+* http://localhost:8080/openapi/ui/ for the OpenAPI 3 spec
+* http://localhost:8080/swagger/ui/ for the Swagger 2 spec
--- a/examples/apikey/README.rst
+++ b/examples/apikey/README.rst
@@ -1,16 +1,32 @@
-=======================
+===============
 API Key Example
-=======================
+===============
 
-Running:
+Preparing
+---------
+
+Create a new virtual environment and install the required libraries
+with these commands:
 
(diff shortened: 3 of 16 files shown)


chrisinmtown:
Please review, if you like this approach I will extend to other README.rst files.

RobbeSneyders:
Thanks @chrisinmtown, this looks good to me. Would be great to extend it to the other examples.

chrisinmtown:
I am extending the changes to all the examples. Please check example `sqlalchemy` which has become a bit stale. I dropped the `convert_unicode` parameter in the `create_engine` function and then it started, but the first GET attempt yields this:
```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 123145410666496 and this is thread id 140704480263680.
```
Do you have time for this?

RobbeSneyders:
Let's leave that example as is for now then. Would be great if you could add a notice that it is outdated to the README.

For the other examples, the `flask` extra is not needed for most of them.

chrisinmtown:
> Let's leave that example as is for now then. Would be great if you could add a notice that it is outdated to the README.

I will ask the cloud a few questions and see if I can figure it out :)

> For the other examples, the `flask` extra is not needed for most of them.

Right, I copy-pasted a little too quickly, will adjust.

chrisinmtown:
This PR has primarily documentation changes. I tweaked one API spec and one requirements file to make the examples work. Please consider.

chrisinmtown:
Would you please give me an opinion here? It's a real challenge to contribute when people go silent for a full week at a time.

examples/jwt/requirements.txt
```diff
@@ -1,3 +1,2 @@
-..[swagger-ui]
 python-jose[cryptography]
 Flask>=0.10.1
```
RobbeSneyders, line 2:
```suggestion
```

examples/sqlalchemy/README.rst
```diff
@@ -2,14 +2,36 @@
...
+
+Create a new virtual environment and install the required libraries
+with these commands:
 
 .. code-block:: bash
 
+    $ python -m venv my-venv
+    $ source my-venv/bin/activate
+    $ pip install 'connexion[flask,swagger-ui,uvicorn]>=3.1.0'
     $ pip install -r requirements.txt
```
RobbeSneyders, line 23: We should remove the flask and connexio dependencies from `requirements.txt` if we install them directly on the command line.

RobbeSneyders (commented):
Sorry for the delay @chrisinmtown 

Two small comments, then it's good to go for me.

examples/jwt/requirements.txt
chrisinmtown: Ok, dropped the Flask line

examples/sqlalchemy/README.rst
chrisinmtown: Ok, dropped the connexion and Flask lines from requirements.txt

RobbeSneyders (approved):
Thanks @chrisinmtown!

PR #1995, opened 2024-11-11
import python_multipart instead of multipart

dotlambda:

 - use `import python_multipart`, see https://github.com/Kludex/python-multipart/blob/master/CHANGELOG.md#0013-2024-10-20


pyproject.toml
```diff
@@ -50,7 +50,7 @@ httpx = ">= 0.23"
 inflection = ">= 0.3.1"
 jsonschema = ">=4.17.3"
 Jinja2 = ">= 3.0.0"
-python-multipart = ">= 0.0.5"
+python-multipart = ">= 0.0.13"
```
RobbeSneyders, line 53:
```suggestion
python-multipart = ">= 0.0.15"
```

RobbeSneyders (commented):
Thanks @dotlambda!

Looks like [poetry is failing](https://github.com/python-poetry/poetry/issues/9505) due to the [yanked releases](https://pypi.org/project/python-multipart/#history). I believe my suggestion below should fix this.

RobbeSneyders (approved):

PR #1999, opened 2024-11-20
Don't regenerate operation_id in MockResolver

RobbeSneyders:
When the same operation was being resolved multiple times (eg. in different middleware), a new mock operation_id was generated every time, breaking routing for these mocked operations.

```
connexion.middleware.abstract.MissingOperation: Encountered unknown operation_id
```

This PR keeps track of the generated operation_ids and returns a previously generated id if the same operation is resolved.

CC: @julienschuermans

### DIFF
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -14,6 +14,7 @@ class MockResolver(Resolver):
     def __init__(self, mock_all):
         super().__init__()
         self.mock_all = mock_all
+        self._resolved_operations = {}
         self._operation_id_counter = 1
 
     def resolve(self, operation):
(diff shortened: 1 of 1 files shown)


connexion/mock.py
```diff
@@ -23,10 +24,15 @@ def resolve(self, operation):
...
         """
         operation_id = self.resolve_operation_id(operation)
-        if not operation_id:
-            # just generate an unique operation ID
-            operation_id = f"mock-{self._operation_id_counter}"
-            self._operation_id_counter += 1
+        if operation_id is None:
+            operation_ref = operation.path + operation.method
+            if operation_ref in self._resolved_operations:
+                operation_id = self._resolved_operations.get(operation_ref)
```
Ruwann, line 30: In this branch, we know it exists, so we should be able to directly use access by key?

Ruwann (dismissed):
Thanks, left a minor (optional) comment.

Can we add a test for this? Doubting whether it is overkill to do this for the benefit we'll get from a test for the mock resolver but if only for the resolver, a test should be that hard to add?

RobbeSneyders:
Looking into adding a test makes me doubt if this is the right way to resolve this issue.

We are creating a new `SwaggerOperation` / `OpenAPIOperation` in every middleware ([code](**url**)), which might not be ideal. I'll check if we could handle this differently when I find the time and update this PR.

PR #2000, opened 2024-11-21
Return NoContent when mocking HTTP 204 operations

julienschuermans:
There's never a need to return a response for HTTP 204 - and these code are unlikely to have any "example" response documented  in the spec.

Avoids a uvicorn RuntimeError `"Response content longer than Content-Length"` when mocking a (3.0) spec containing:

```
...
      responses:
        '204':
          description: No Content
        '401':
          description: Unauthorized
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ProblemDetails'
...
```

* We could also move this `if/else` check into the operation's `example_response` method. Thoughts?

### DIFF
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -49,7 +49,9 @@ class MockResolver(Resolver):
 
     def mock_operation(self, operation, *args, **kwargs):
         resp, code = operation.example_response()
-        if resp is not None:
+        if code == 204:
+            return None, code
+        elif resp is not None:
             return resp, code
         return (
             "No example response defined in the API, and response "
--- a/connexion/mock.py
+++ b/connexion/mock.py
@@ -4,9 +4,7 @@ This module contains a mock resolver that returns mock functions for operations 
 
 import functools
 import logging
-from http import HTTPStatus
 
-from connexion.datastructures import NoContent
 from connexion.resolver import Resolution, Resolver, ResolverError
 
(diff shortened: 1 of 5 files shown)


chrisinmtown:
I guess you'll have to add/extend a test case to cover the newly added lines?

julienschuermans:
> I guess you'll have to add/extend a test case to cover the newly added lines?

I can do that - not sure where to place this though. In `test_mock.py` or `test_mock3.py` ? I guess we want to test behaviour for both Swagger 2.0 and OpenAPI 3? @RobbeSneyders do you have any input?

RobbeSneyders:
Thanks @julienschuermans!

I would indeed prefer to move it to the `example_response` method though.

For the test, I think you can add:
- A swagger test in `test_mock.py`
- An OpenAPI test in `test_mock3.py`

It should be quite straight forward if you start from a copy of an existing test.

Don't ask me why the numbering of the `test_mock` files is as it is :sweat:

julienschuermans:
I've moved the code into the `example_response` implementations and added some tests.
@RobbeSneyders should be good to go?

chrisinmtown:
I recommend you squash the 13 commits down to a single one.

Ruwann (approved):
Thanks, @julienschuermans. This one can be merged for me.

We can squash the commits during merging, as we're typically doing a squash and rebase.

PR #2010, opened 2024-12-03
use connexion logger instead of root logger

FelixSchwarz:
In `request_validation.py` and `response_validation.py` there is a call to `logging.info()` which means the log message goes through the root logger instead of the `connexion.middleware.validation` logger.

### DIFF
--- a/connexion/middleware/request_validation.py
+++ b/connexion/middleware/request_validation.py
@@ -120,7 +120,7 @@ class RequestValidationOperation:
             try:
                 body_validator = self._validator_map["body"][mime_type]  # type: ignore
             except KeyError:
-                logging.info(
+                logger.info(
                     f"Skipping validation. No validator registered for content type: "
                     f"{mime_type}."
                 )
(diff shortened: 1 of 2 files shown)


FelixSchwarz:
Btw: Maybe it would be good to downgrade the log level for the message. I think this is more a debug message. Not having a validator is a static condition (either the dev configured the validator or not) so the log message is pretty useless in production. My personal rule of thumb is that something which is not relevant at all in the normal mode of operation should have a log level below INFO.

chrisinmtown:
Nice catch! That's an easy one to miss, logging doesn't complain about direct use of the root logger. I didn't know you could just call `logging.info` like that.

chrisinmtown (approved):
Nice catch! That's an easy one to miss, logging doesn't complain about direct use of the root logger. I didn't know you could just call logging.info like that.

RobbeSneyders (approved):
Thanks @FelixSchwarz!

Not sure about the logging level. It depends on the content type of the incoming request, so it is not completely static. I would argue that validation is one of the main features of Connexion, so silently bypassing it is also no ideal.

PR #2011, opened 2024-12-05
Update example for latest version of SQLAlchemy

chrisinmtown:
Pass argument that disables SQLite thread check when creating the engine
Revise request-processing functions to work within a database session context
Drop TODO from the example README

### DIFF
--- a/examples/sqlalchemy/app.py
+++ b/examples/sqlalchemy/app.py
@@ -5,58 +5,60 @@ import connexion
 import orm
 from connexion import NoContent
 
-db_session = None
-
 
 def get_pets(limit, animal_type=None):
-    q = db_session.query(orm.Pet)
-    if animal_type:
-        q = q.filter(orm.Pet.animal_type == animal_type)
-    return [p.dump() for p in q][:limit]
+    with db_session_factory() as db_session:
+        q = db_session.query(orm.Pet)
+        if animal_type:
+            q = q.filter(orm.Pet.animal_type == animal_type)
+        return [p.dump() for p in q][:limit]
 
 
 def get_pet(pet_id):
-    pet = db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).one_or_none()
-    return pet.dump() if pet is not None else ("Not found", 404)
+    with db_session_factory() as db_session:
+        pet = db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).one_or_none()
+        return pet.dump() if pet is not None else ("Not found", 404)
 
 
 def put_pet(pet_id, pet):
-    p = db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).one_or_none()
-    pet["id"] = pet_id
-    if p is not None:
-        logging.info("Updating pet %s..", pet_id)
-        p.update(**pet)
-    else:
-        logging.info("Creating pet %s..", pet_id)
-        pet["created"] = datetime.datetime.utcnow()
-        db_session.add(orm.Pet(**pet))
-    db_session.commit()
-    return NoContent, (200 if p is not None else 201)
+    with db_session_factory() as db_session:
+        p = db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).one_or_none()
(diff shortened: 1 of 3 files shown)


chrisinmtown:
@FelixSchwarz  please review and comment!

Ruwann (approved):
DId a quick check and seems to work fine for me. Tbh, no too familiar with SQLAlchemy but judging from the link and documentation this seems to be the correct way to achieve this. Anything specific you'd like to get a 2nd pair of eyes on to double check?

examples/sqlalchemy/app.py
```diff
@@ -5,59 +5,54 @@
...
-        db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).delete()
-        db_session.commit()
-        return NoContent, 204
-    else:
-        return NoContent, 404
+    with db_sess_mkr() as db_session:
+        pet = db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).one_or_none()
+        if pet is not None:
+            logging.info("Deleting pet %s..", pet_id)
+            db_session.query(orm.Pet).filter(orm.Pet.id == pet_id).delete()
```
FelixSchwarz, line 43: This should be just `db_session.delete(pet)`

examples/sqlalchemy/orm.py
```diff
@@ -24,11 +25,12 @@ def dump(self):
         return {k: v for k, v in vars(self).items() if not k.startswith("_")}
 
 
-def init_db(uri):
-    engine = create_engine(uri, convert_unicode=True)
-    db_session = scoped_session(
-        sessionmaker(autocommit=False, autoflush=False, bind=engine)
+def init_db():
+    # https://stackoverflow.com/questions/6519546/scoped-sessionsessionmaker-or-plain-sessionmaker-in-sqlalchemy
+    engine = create_engine(
```
FelixSchwarz, line 30: Thank you for adding these options to the code. Maybe it is also a good idea to add a comment mentioning that `check_same_thread` and `StaticPool` can be helpful for unit testing of in-memory sqlite databases and should not be used in production.

examples/sqlalchemy/app.py
```diff
@@ -5,59 +5,54 @@
...
+            db_session.commit()
+            return NoContent, 204
+        else:
+            return NoContent, 404
 
 
 logging.basicConfig(level=logging.INFO)
-db_session = orm.init_db("sqlite:///:memory:")
+db_sess_mkr = orm.init_db()
```
FelixSchwarz, line 51: How about `db_session_maker`? I mean it is not that much longer but does not require additional "mental processing". I thought about suggesting to just keep `db_session` but I can see that this might be confusing as it is really not the session but a session factory/maker.

FelixSchwarz (approved):
Looks good to me as well. The delete should be fixed, the rest is mostly personal preference.

examples/sqlalchemy/app.py
chrisinmtown: Right, I should have spotted that, will fix.

examples/sqlalchemy/app.py
chrisinmtown: How do you feel about `db_session_factory` ?

examples/sqlalchemy/orm.py
chrisinmtown: Will add the comment.

chrisinmtown:
Thanks @Ruwann  and @FelixSchwarz for the comments. New version pushed.

examples/sqlalchemy/app.py
FelixSchwarz: Sure, my only beef was with the (unnecessary as I felt) abbreviation.

examples/sqlalchemy/app.py
FelixSchwarz: Oh, and I noticed only after writing the review comment that the code was always like this. :-)

Ruwann (approved):
Thanks, @chrisinmtown and @FelixSchwarz !

Issue #1988, opened 2024-11-06
Extend migration guide with details about Connexion test_client for testing and tox


chrisinmtown:
### Description

Please extend the migration guide with some details about the required change from getting a client by calling the Flask app's `test_client()` function, to getting a client by calling the Connexion app's `test_client()` function. 

### Expected behaviour

Migration guide also explains how to migrate tests. Ideally, one of the example projects includes a `tox.ini` file and a test case to demonstrate how to test a Connexion v3 app.

### Actual behaviour

The testing page is helpful for pointing people to the `test_client()` function, but you have to go looking for it, since the migration guide is silent.

https://connexion.readthedocs.io/en/latest/testing.html

chrisinmtown:
You already documented this, sorry that I was momentarily blind, finally found this: https://connexion.readthedocs.io/en/latest/testing.html

I will also add a comment on Kevin Martin's example project about this.

PR #2026, opened 2025-01-04
Add example of the Starlette test client on a simple Connexion REST app

chrisinmtown:
The example Connexion app processes JSON requests and responses as specified with OpenAPI v2 (aka Swagger) or OpenAPI v3 file format.
JSON responses are validated against the spec.
An error handler catches exceptions raised while processing a request.
Tests are run by `tox` which also reports code coverage.


chrisinmtown:
Please help me understand the test failures here, since I didn't change any code that is tested automatically. I see a message `Error: The operation was canceled.` which I cannot explain.

chrisinmtown:
Thanks for approving a run of the test pipeline on this PR. I botched the import order in my initial attempts, I'm learning to run `isort` on everything first.

examples/testclient/hello/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    Connexion asynchronous middleware framework without a wrapper.
+    If a plain function is provided, the framework calls the function
+    from a threadpool and the exception stack trace is not available.
+
+    :param request: Request that failed
+    :parm ex: Exception that was raised
+    :return: ConnexionResponse with RFC7087 problem details
+    """
+    # log the request URL, exception and stack trace
+    logger.exception("Request to %s caused exception", request.url)
```
Ruwann, line 25: Does this work properly, don't we need to supply the exception as an argument via `exc_info=ex`?

examples/testclient/hello/__init__.py
Ruwann, line 1: What do you think of renaming this file instead of using `__init__.py`? Or add it to `app.py`? To keep it more consistent with other examples

examples/testclient/hello/app.py
```diff
@@ -0,0 +1,30 @@
+import logging
+
+from . import conn_app
+
+# reuse the configured logger
+logger = logging.getLogger("uvicorn.error")
```
Ruwann, line 6: What's the reasong for using the `uvicorn.error` logger instead of a specific one for this file/application?

Ruwann (commented):
Thanks! Already looking very nice, left some small comments/questions

Ruwann:
> Thanks for approving a run of the test pipeline on this PR. I botched the import order in my initial attempts, I'm learning to run `isort` on everything first. I'm struggling to figure out the exact version of `isort` to use, or the right configuration, because the import ordering set by my local `isort` was rejected by the tox `pre-commit` step :/

No worries, I have been using ruff on my other projects as it combines a lot of these into a single tool, that's more easy to use imo. Is it still unclear?

chrisinmtown:
Thank you for the comments. I will revise. Would appreciate guidance on the right way, or the connexion way, or what have you :) to structure an app for use from tox. Creating the app object in `__init__.py` makes that easy but I'm sure it's not the only way.

examples/testclient/hello/app.py
chrisinmtown: This was basically a cheat, the level is set appropriately on this logger so the output from my code appears together with output from uvicorn, and no additional configuration is necessary.

examples/testclient/hello/__init__.py
chrisinmtown: The python logger does magic when you call its `exception` method, and that magic breaks when the context changes, like when running the function from a new thread pool. I will test what you suggest - pass the exception - and see if the full stack trace is shown.

examples/testclient/hello/__init__.py
chrisinmtown: Would appreciate guidance on the right way, or the connexion way, or what have you :) to structure an app for use from tox. Creating the app object in __init__.py makes that easy but I'm sure it's not the only way.

examples/testclient/hello/__init__.py
chrisinmtown: Thank you for @Ruwann for pointing me to the `exc_info` parameter. Yes, the stack trace appears as expected when that parameter is passed. That seems like a better solution than using `async` to define the exception handler function.

examples/testclient/hello/__init__.py
chrisinmtown: I restructured this example to use a single file, and eliminate the package `hello`. Please review again. Thanks in advance.

chrisinmtown:
I see these failures in a site-package that I cannot explain, please suggest if there's anything I can do:
```
/pc/clone/u0gC1H9STNixN0l108-6fw/py_env-python3/lib/python3.12/site-packages/referencing/_core.py:565: error: Invalid "type: ignore" comment
/pc/clone/u0gC1H9STNixN0l108-6fw/py_env-python3/lib/python3.12/site-packages/referencing/_core.py:566: error: Invalid "type: ignore" comment
```

chrisinmtown:
Would someone please advise if there's anything I can do about the test pipeline failure in `mypy`?

chrisinmtown:
Greetings @Ruwann and @RobbeSneyders I think I addressed all your comments on this PR, please advise if you would like this example in the V3 codebase, thanks.

examples/testclient/hello/app.py
chrisinmtown: Please say if using the `unicorn.error` logger is a deal breaker and I will change the logging.

chrisinmtown:
This PR addresses issue #1988 with an example instead of a revised document. Please comment.

chrisinmtown:
Looks like two my PRs are fighting with each other :(  This PR adds a "tests" directory, it's the only example with that. Unfortunately the very recently merged change to `tox.ini` (that allows choice of scope) now seems to cause `pytest` to pull in the newly added "tests/conftest.py" file from this new example instead of the top-level "test/conftest.py" file, leading to immediate failures. I speculate it's because of the entry "testspaths=tests" in `tox.ini`.  Trying to figure out a workaround.

chrisinmtown:
I submitted a PR to restore the `tests` argument to the poetry invocation of pytest, that limits the directory scope, and then the files introduced by this PR are not picked up and used during a normal test run of Connexion, see #2092

PR #2030, opened 2025-01-23
Update docstring of `connexion.ConnexionMiddleware.add_api`

bastantoine:
- Update docstring of `connexion.ConnexionMiddleware.add_api` to match return value to actual return.

### DIFF
--- a/connexion/middleware/main.py
+++ b/connexion/middleware/main.py
@@ -403,7 +403,6 @@ class ConnexionMiddleware:
             middlewares. This can be used to pass arguments to middlewares added beyond the default
             ones.
 
-        :return: None
         """
         if self.middleware_stack is not None:
             raise RuntimeError("Cannot add api after an application has started")
--- a/connexion/middleware/main.py
+++ b/connexion/middleware/main.py
@@ -403,7 +403,7 @@ class ConnexionMiddleware:
             middlewares. This can be used to pass arguments to middlewares added beyond the default
             ones.
 
-        :return: The Api registered on the wrapped application.
+        :return: None
         """
         if self.middleware_stack is not None:
             raise RuntimeError("Cannot add api after an application has started")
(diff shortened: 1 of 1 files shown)


connexion/middleware/main.py
```diff
@@ -403,7 +403,7 @@ def add_api(
             middlewares. This can be used to pass arguments to middlewares added beyond the default
             ones.
 
-        :return: The Api registered on the wrapped application.
+        :return: None
```
chrisinmtown, line 406: Please update the type hint also, which right now says `None`

bastantoine:
> This PR modifies a function that returns no value, to add `:return:` line to the function's Pydoc block. This just looks wrong to me. Please explain.

Actually this PR only updates the docstring of the function, and not the function itself.

The function still returns nothing, but the docstring said that it actually returned something, which can be confusing (and I can confirm because I got confused by it 😅)

chrisinmtown:
You're right, I misread your PR, thanks for clarifying.  +1 from me!

chrisinmtown (approved):

connexion/middleware/main.py
Ruwann, line 406:
```suggestion
```

We can just leave out the return documention instead

Ruwann (commented):
Thank for the PR! Although we can just delete the return documentation entirely

connexion/middleware/main.py
```diff
@@ -403,7 +403,6 @@ def add_api(
             middlewares. This can be used to pass arguments to middlewares added beyond the default
             ones.
```
Ruwann, line 405:
```suggestion
```

Ruwann (approved):

Ruwann:
Pipeline fails due to some unrelated issue (for some reason, mypy is checking some files in a transitive dependency that got recently added)

PR #2047, opened 2025-04-26
Add empty argument to sed -i option in tox.ini for use on MacOS

chrisinmtown:
sed on Linux accepts a `-i` option with or without an argument, a file extension to use.
sed on MacOS requires the `-i` option to have an argument, but it can be empty.

This change fixes the following error seen when running tox on macOS:
```
sed: 1: "s/"(\^|~|>=)([ 0-9])/"= ...": \2 not defined in the RE
```


chrisinmtown:
I didn't open an issue. I would be glad to create one if you prefer. Please comment.

chrisinmtown:
This didn't work in the GitHub build environment. Back to the drawing board, Wile E. Coyote.

chrisinmtown:
This solution gains test portability across Linux and MacOS, at the cost of leaving a file behind from every tox invocation. Please consider if you feel the benefits outweigh the costs.

PR #2049, opened 2025-05-04
Specify skip for separate referencing package

Ruwann:
Fixes the mypy error that is causing all pipelines to fail. Underlying reason seems to be that the `jsonschema` library was updated and factored out some code into a separate package. So updated it to also reference that file in our mypy overrides.

Might be fixed for newer versions by this recent commit https://github.com/python-jsonschema/referencing/commit/f830e4c8d6a3b9b792d091ee8c6ac2eafbcf0de0


chrisinmtown:
Thanks for this fix.

PR #2052, opened 2025-05-07
Worker threads WSGI Middleware Configurable

EyalMK:
Fixes https://github.com/spec-first/connexion/issues/1979.
"Make FlaskApp number of threads configurable"

- FlaskApp has an additional optional worker_threads constructor argument
- worker_threads default to 10; identical to the default no. threads for WSGIMiddleware from a2wsgi


EyalMK:
@RobbeSneyders  Would much appreciate a review for this. Current state makes Connexion 3 and non-thread safe libraries totally incompatible...

PR #2069, opened 2025-07-21
request body in openapi no need to append the components

frankzye:
Fixes # .

reproduce:

* start up server, and call one of the api with Post and body
* open swagger and refresh
error:

Failed validating 'oneOf' in schema['properties']['paths']['patternProperties']['^\\/']['patternProperties']['^(get|put|post|delete|options|head|patch|trace)$']['properties']['requestBody']:
    {'oneOf': [{'$ref': '#/definitions/RequestBody'},

 request body in openapi no need to append the components

### DIFF
--- a/connexion/operations/openapi.py
+++ b/connexion/operations/openapi.py
@@ -134,8 +134,8 @@ class OpenAPIOperation(AbstractOperation):
     def produces(self):
         return self._produces
 
-    def with_definitions(self, schema: dict):
-        if self.components:
+    def with_definitions(self, schema: dict, append_components=True):
+        if self.components and append_components:
             schema.setdefault("schema", {})
             schema["schema"]["components"] = self.components
         return schema
@@ -241,5 +241,5 @@ class OpenAPIOperation(AbstractOperation):
                 )
             content_type_dict = MediaTypeDict(self.request_body.get("content", {}))
             res = content_type_dict.get(content_type, {})
-            return self.with_definitions(res)
+            return self.with_definitions(res, append_components=False)
         return {}
--- a/tests/api/test_parameters.py
+++ b/tests/api/test_parameters.py
@@ -654,8 +654,12 @@ def test_openapi_schema_validate_with_request_body_change(simple_app):
     app_client = simple_app.test_client()
 
     if simple_app._spec_file == "openapi.yaml":
-        response = app_client.post("/v1.0/test-default-object-body", headers={"content-type": "application/json"}, json={"image_version": "2015-08-26"})
+        response = app_client.post(
+            "/v1.0/test-default-object-body",
+            headers={"content-type": "application/json"},
+            json={"image_version": "2015-08-26"},
+        )
         assert response.status_code == 200
-        
+
         response = app_client.get("/v1.0/openapi.json")
         assert response.status_code == 200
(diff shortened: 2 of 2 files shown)


chrisinmtown:
Not quite sure so I want to ask, does this PR fix this issue: https://github.com/spec-first/connexion/issues/2029 ?

frankzye:
> Not quite sure so I want to ask, does this PR fix this issue: #2029 ?
yes, fix the issule

chrisinmtown:
> Fixes # .

Thanks, very glad to hear it. Please extend your commit message to reference that issue with a "Fixes" comment like shown above.

frankzye:
> > Not quite sure so I want to ask, does this PR fix this issue: #2029 ?
> > yes, fix the issule

updated

chrisinmtown:
This project has an enormous suite of automated tests. Would you please consider adding a new test that checks the revised behavior? I.e., the test should fail on released version 3.2, succeed with this change.

frankzye:
> This project has an enormous suite of automated tests. Would you please consider adding a new test that checks the revised behavior? I.e., the test should fail on released version 3.2, succeed with this change.

added with tests

chrisinmtown:
@Ruwann @RobbeSneyders  would one of you maintainers please approve a run of the GitGHub workflow to validate this change?

tests/api/test_parameters.py
```diff
@@ -648,3 +648,18 @@ def test_cookie_param(simple_app):
     response = app_client.get("/v1.0/test-cookie-param")
     assert response.status_code == 200
     assert response.json() == {"cookie_value": "hello"}
+
+
+def test_openapi_schema_validate_with_request_body_change(simple_app):
+    app_client = simple_app.test_client()
+
+    if simple_app._spec_file == "openapi.yaml":
```
chrisinmtown, line 656: Would you please help me understand the need for this `if` clause? Are there multiple instances of `simple_app` and some have a different spec file?  I suppose you are trying to increase the chance that the app is listening at endpoint `/v1.0/test-default-object-body`. Seems like all the other tests in this file have this kind of assumption, so I'm not sure it's necessary to check in this one. Or maybe I'm missing something huge :)

connexion/operations/openapi.py
```diff
@@ -134,8 +134,8 @@ def consumes(self):
     def produces(self):
         return self._produces
 
-    def with_definitions(self, schema: dict):
-        if self.components:
+    def with_definitions(self, schema: dict, append_components=True):
+        if self.components and append_components:
```
chrisinmtown, line 138: I realize that only the init method has pydoc for parameters, the rest of the methods have basically none. With that said, since you are introducing a new parameter with a default value, I think it would be really helpful to add a little pydoc section with a sentence about the new parameter.  Please consider it. I'd like to suggest that text, but I am still trying to understand exactly what is being appended here (or not).

tests/api/test_parameters.py
chrisinmtown: One more comment, if you can please have patience with me. I see there is a test file `test_responses.py`. Because this new test case doesn't actually exercise any *parameters*, maybe you might consider moving this new case to that file?

PR #2070, opened 2025-07-25
Fix/multiple content type headers lead to 415 consistently

### LINKED ISSUE #2040: Passing in invalid headers to an API causes 500 error
I have a simple test case with incorrect headers that causes a 500 error.  Although the cause is in the test, it is something that could happen in production with a public facing API built on Connexion.  I have these concerns/opinions:

- 500 errors are pretty tough to manage, for both provider and consumer, since one should not return too much information about  the cause of the error. 
- By convention, they should be pretty rare to see, and ideally never caused by invalid input.  i.e. as a general matter, validations should prevent most 500's caused by invalid input.
- This case should be treated as a validation error
- Question:  Where do the details related to the 500 go, when connexion experiences them ?  I could only see the details in debug mode.  In a production environment, details related to 500 errors must be logged.
- I could not find information about configuring connexion logging in to docs.

Thanks for your work here, BTW, I've been using connexion for about 2 years, and am just switching to V3..

Code below.. The 'get' test function does not see an error, but the 'post' does.

Test app.py:

`import connexion
from flask import jsonify
from datetime import datetime

from connexion.middleware import MiddlewarePosition
from starlette.middleware.cors import CORSMiddleware

# In-memory storage for todos
todos = {}
next_id = 1

# API endpoints implementation
def get_todos():
    """Retrieve all todos"""
    return jsonify(list(todos.values()))

def create_todo(body):
    """Create a new todo"""
    global next_id
    
    new_todo = {
        'id': next_id,
        'task': body['task'],
        'completed': body.get('completed', False)
    }
    
    todos[next_id] = new_todo
    next_id += 1
    return jsonify(new_todo), 201

def get_todo(todo_id):
    """Retrieve a specific todo"""
    if todo_id not in todos:
        return jsonify({'error': 'Todo not found'}), 404
    return jsonify(todos[todo_id])

# Create the Connexion app
app = connexion.FlaskApp(__name__, specification_dir='./')
app.add_middleware(
    CORSMiddleware,
    position=MiddlewarePosition.BEFORE_EXCEPTION,
    allow_origins=["*"],
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)
with app.app.app_context():
    app.add_api('swagger.yaml')

# Add some sample data
todos[1] = {'id': 1, 'task': 'Learn Connexion', 'completed': False}
todos[2] = {'id': 2, 'task': 'Build REST API', 'completed': True}

if __name__ == '__main__':
    app.run(port=5000, debug=True)`

Sample pytest.  Note that the header has two definitions for content-type, which is the input error in the test:

`import pytest
from starlette.testclient import TestClient
from app import app  # Import the Connexion app from your main file
import json

# Test fixtures
@pytest.fixture
def client():
    """Create a test client for the app"""
    # Use Starlette's TestClient with the Connexion app
    with TestClient(app) as client:
        # Reset the todos dictionary before each test
        global todos, next_id
        from app import todos, next_id
        todos.clear()
        todos[1] = {'id': 1, 'task': 'Test task 1', 'completed': False}
        todos[2] = {'id': 2, 'task': 'Test task 2', 'completed': True}
        next_id = 3
        yield client

@pytest.fixture
def json_headers():
    """Common headers for JSON requests"""
    return {
      'content-type': 'application/json',
      'X-ApiKey': 'ApiKey',
      'Content-Type': 'application/json',
      'Accept': 'application/json'
      }
    
    
def test_get_all_todos(client):
    """Test retrieving all todos"""
    response = client.get('/todos')
    
    assert len(response.json()) == 2  # Should match initial test data
    assert response.json()[0]['id'] == 1
    assert response.json()[0]['task'] == 'Test task 1'
    assert response.json()[1]['completed'] == True

def test_create_todo_success(client, json_headers):
    """Test creating a new todo with valid data"""
    new_todo = {
        'task': 'New test task',
        'completed': False
    }
    
    response = client.post(
        '/todos',
        data=json.dumps(new_todo),
        headers=json_headers
    )
    
    assert response.status_code == 201
    assert response.json()['id'] in [1, 2, 3]  # Should be next available ID
    assert response.json()['task'] == 'New test task'
    assert response.json()['completed'] == False

`

yaml:
`openapi: 3.0.0
info:
  title: Todo List API
  version: 1.0.0
  description: A simple Todo List management API

paths:
  /todos:
    get:
      summary: List all todos
      operationId: app.get_todos
      responses:
        '200':
          description: Successful response
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Todo'
    post:
      summary: Create a new todo
      operationId: app.create_todo
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/TodoInput'
      responses:
        '201':
          description: Todo created successfully
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Todo'

components:
  schemas:
    Todo:
      type: object
      properties:
        id:
          type: integer
        task:
          type: string
        completed:
          type: boolean
      required:
        - id
        - task
    TodoInput:
      type: object
      properties:
        task:
          type: string
        completed:
          type: boolean
      required:
        - task`

Output:  <connexion==3.2.0, Flask==3.1.0>
Response.text for post test:
{"type": "about:blank", "title": "Internal Server Error", "detail": "The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.", "status": 500}

stdout in debug mode:
File "....venv/lib/python3.13/site-packages/connexion/lifecycle.py", line 120, in get_body
    if is_json_mimetype(self.content_type):
       ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "....venv/lib/python3.13/site-packages/connexion/utils.py", line 161, in is_json_mimetype
    maintype, subtype = mimetype.split("/")  # type: str, str
    ^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)

Lewiscowles1986:
I'm pleased to submit a pull request that I believe fixes #2040 on GitHub. The proposed changes aim to improve consistency of the behavior of content-type header inspection in Connexion.

Key modifications include:

* Ensuring that searching for content types does not short-circuit, aligning with other header inspections.
* Adding a test to safeguard against potential regressions.

The reason this fix is effective (tested in an interactive debugger with multiple clients and original tests) is that
the modified method was previously returning 'application/json', while subsequent implementations downstream returned
'application/json,application/json'. This resulted in three parts when splitting by '/' instead of two, which caused
issues.

I did investigate alternative approaches to fixing the '/'-split issue mentioned in the original report. However, this
led me to modify more code, which didn't resolve the 500 error and 415 response from Safari. I opted for a more precise
approach that focused on the specific method modification, as it yielded better results.

Please let me know if you have any questions or concerns about this pull request.

### DIFF
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -158,7 +158,7 @@ def is_json_mimetype(mimetype):
     if mimetype is None:
         return False
 
-    maintype, subtype = mimetype.split("/")  # type: str, str
+    maintype, subtype = mimetype.split("/", maxsplit=1)  # type: str, str
     if ";" in subtype:
         subtype, parameter = subtype.split(";", maxsplit=1)
     return maintype == "application" and (
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -310,10 +310,8 @@ def extract_content_type(
 
         if decoded_key.lower() == "content-type":
             if isinstance(value, bytes):
-                content_type = value.decode("latin-1")
-            else:
-                content_type = value
-            break
+                value = value.decode("latin-1")
+            content_type = ",".join([content_type, value] if content_type else [value])
 
     return content_type
 
--- a/tests/test_json_validation.py
+++ b/tests/test_json_validation.py
@@ -175,3 +175,26 @@ def test_defaults_body(json_validation_spec_dir, spec):
     )
     assert res.status_code == 200
     assert res.json().get("human")
+
+
+def test_multiple_json_content_type(json_validation_spec_dir, spec):
+    """ensure that defaults applied that modify the body"""
+
+    class MyDefaultsJSONBodyValidator(DefaultsJSONRequestBodyValidator):
+        pass
+
+    validator_map = {"body": {"application/json": MyDefaultsJSONBodyValidator}}
+
+    app = App(__name__, specification_dir=json_validation_spec_dir)
+    app.add_api(spec, validate_responses=True, validator_map=validator_map)
+    app_client = app.test_client()
+
+    res = app_client.post(
+        "/v1.0/user",
+        data=json.dumps({"name": "foo"}),
+        headers={
+            "content-type": "application/json",
+            "Content-Type": "application/json",
+        },
+    )
+    assert res.status_code == 415
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -62,16 +62,29 @@ def test_deep_get_list():
     assert utils.deep_get(obj, ["0", "properties", "id"]) == {"type": "string"}
 
 
-def test_is_json_mimetype():
-    assert utils.is_json_mimetype("application/json")
-    assert utils.is_json_mimetype("application/vnd.com.myEntreprise.v6+json")
-    assert utils.is_json_mimetype(
-        "application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0"
-    )
-    assert utils.is_json_mimetype(
-        "application/vnd.com.myEntreprise.v6+json; charset=UTF-8"
-    )
-    assert not utils.is_json_mimetype("text/html")
+@pytest.mark.parametrize(
+    "mime_type",
+    [
+        "application/json",
+        "application/vnd.com.myEntreprise.v6+json",
+        "application/vnd.scanner.adapter.vuln.report.harbor+json; version=1.0",
+        "application/vnd.com.myEntreprise.v6+json; charset=UTF-8",
+    ],
+)
+def test_is_json_mimetype_true(mime_type: str):
+    assert utils.is_json_mimetype(mime_type)
+
+
+@pytest.mark.parametrize(
+    "mime_type",
+    [
+        "application/json,application/json",
+        "text/html",
+        "text/json",
+    ],
+)
+def test_is_json_mimetype_false(mime_type: str):
+    assert not utils.is_json_mimetype(mime_type)
 
 
 def test_sort_routes():


connexion/utils.py
```diff
@@ -310,10 +310,8 @@ def extract_content_type(
```
chrisinmtown, line 310: The existing function seems to get the *first* Content-Type header and return it. If I am reading this correctly, how does a CSV value reach the `is_json` function? What actually arrives from `curl`? Does Connexion receive two headers, or one header with a CSV list?

I think at line 294 it would be very helpful to mention this edge case - multiple Content-Type headers - that's handled by this revised function. Interestingly the existing comment states plural `headers`.

Lewiscowles1986: Hi @chrisinmtown 

One thing I am confused about is that this line is unchanged in the PR.

Is communicating internals, not venturing into the territory of locking to a specific implementation? 

Normally I'd have a unit test to explicitly mention cover this functionality of this function.

What I've done (instead) is remove nuanced edge behaviour from this one place, so that it behaves the same as the other mime-type functions, returning the same output even. 

Before someone asks, this function (and I'm guessing others) have no unit tests. That and the fact there are many ways to get the mime-type could be common factors in the bug this is seeking to fix.

If you would like a comment edit, would you mind providing a suggestion, in a code block please?

chrisinmtown: Oh you're absolutely right, you didn't change line 294, it just seems to me that if the pydoc states clearly what this function handles (and does not), it will help people like us the next time some weird input triggers an undesired behavior.

Lewiscowles1986: I honestly feel like

```
"""Extract the mime type and encoding from the content type headers.
    :param headers: Headers from ASGI scope
    :return: The content type if available in headers, otherwise None
"""
```

Is the perfect multi-line header for this, but invite suggestions for improvements

Lewiscowles1986: this now behaves, exactly as other functions to get the content type from Flask and Starlette seem to be working

chrisinmtown:
Took me too long to see the whole picture here. Tell me if I describe the current request processing correctly:
* something 1 calls `extract_content_type` which stops at the first Content-Type and returns it; in this test, the value is OK and processing continues.
* something 2 calls a different function (?) that extracts the two Content-Type values, builds a CSV list, then calls `is_json_mimetype`; in this test, the value is NOT OK, the function blows up

If I have that right, I think that means there are at least two functions that extract the content type?

Finally returning to this PR, it changes `extract_content_type` to gather all Content-Type values and return a list; in this test case, the "something 1" caller validates the CSV in the result, doesn't like it, rejects with 415.  And "something 2" never is reached.

Lewiscowles1986:
Hi @chrisinmtown you 100% got it 💯 

Apologies I had not made this clearer. My storytelling chip is clearly on the fritz.
So this is about aligning the outputs to be the same

chrisinmtown:
This may be a case of belt-and-suspenders, but please hear me out. Would you consider including this minor change to the `is_json_mimetype` function to ensure the split never gives back more than 2 values:
```
diff --git a/connexion/utils.py b/connexion/utils.py
index 5458ca8..fa5f0d6 100644
--- a/connexion/utils.py
+++ b/connexion/utils.py
@@ -158,7 +158,7 @@ def is_json_mimetype(mimetype):
     if mimetype is None:
         return False
 
-    maintype, subtype = mimetype.split("/")  # type: str, str
+    maintype, subtype = mimetype.split("/", maxsplit=1)  # type: str, str
     if ";" in subtype:
         subtype, parameter = subtype.split(";", maxsplit=1)
     return maintype == "application" and (
```

chrisinmtown:
I agree that this PR makes Connexion behave similarly to how Flask/Werkzeug handles multiple `Content-Type` headers.  For me the interesting question is, why does the Connexion code re-implement a function to fetch `Content-Type` from a request? Why can't it just reuse a Flask/Werkzeug-provided value, and drop the problematic code patched by this PR?  I have not yet figured that out.

chrisinmtown (approved):

Lewiscowles1986:
@chrisinmtown I think that because Connexion uses multiple potential backends, one Starlette and one Flask, this is what leads to the situation.

Lewiscowles1986:
@chrisinmtown can I ask if we think that

```
% python3
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> "a;b;c".split(";", maxsplit=1)
['a', 'b;c']
>>>
```

might cause other issues?

chrisinmtown:
> might cause other issues?

Of course it might :) Still, that line of code is inherently fragile, it assumes the split will always yield a list of exactly length 2. Adding the `maxsplit` parameter makes the function more robust, for when the next clever person thinks of sending a malformed content-type value "foo/bar/baz" or what have you.

Lewiscowles1986:
So it looks like I was confusing the charset detection, from the mime and subtype. 😊 

Change made, and I've changed tests to use `pytest.mark.parametrize`. 

If one mime-type check fails, full visibility about impact of the other cases is there. I believe this is a developer experience win, but I'm open to feedback if this is less clear.

chrisinmtown (approved):

chrisinmtown:
LGTM. Now we wait for the maintainers to notice this PR.

tests/test_json_validation.py
```diff
@@ -175,3 +175,26 @@ class MyDefaultsJSONBodyValidator(DefaultsJSONRequestBodyValidator):
     )
     assert res.status_code == 200
     assert res.json().get("human")
+
+
+def test_multiple_json_content_type(json_validation_spec_dir, spec):
+    """ensure that defaults applied that modify the body"""
```
Ruwann, line 181: The docstring is the one from the test above, it should be about multiple content type headers?

tests/test_json_validation.py
```diff
@@ -175,3 +175,26 @@ class MyDefaultsJSONBodyValidator(DefaultsJSONRequestBodyValidator):
...
+
+    class MyDefaultsJSONBodyValidator(DefaultsJSONRequestBodyValidator):
+        pass
+
+    validator_map = {"body": {"application/json": MyDefaultsJSONBodyValidator}}
```
Ruwann, line 186: I don't think we need to update the custom body validator in this test

connexion/utils.py
```diff
@@ -310,10 +310,8 @@ def extract_content_type(
 
         if decoded_key.lower() == "content-type":
             if isinstance(value, bytes):
-                content_type = value.decode("latin-1")
-            else:
-                content_type = value
-            break
+                value = value.decode("latin-1")
+            content_type = ",".join([content_type, value] if content_type else [value])
```
Ruwann, line 314: There should be only one content-type header. If there are multiple, I believe it is more appropriate to raise a BadRequest error to return a 400 status code

tests/test_utils.py
```diff
@@ -62,16 +62,29 @@ def test_deep_get_list():
...
+    ],
+)
+def test_is_json_mimetype_true(mime_type: str):
+    assert utils.is_json_mimetype(mime_type)
+
+
+@pytest.mark.parametrize(
+    "mime_type",
+    [
+        "application/json,application/json",
```
Ruwann, line 81: This isn't a valid mimetype

Ruwann (commented):
I don't quite see what you mean with "while subsequent implementations downstream returned
'application/json,application/json'. "
Can we instead fix those?

Lewiscowles1986:
@Ruwann is 400 the desired response code? There seems to be conflicting feedback in this PR about approach and I would very much appreciate knowing:

- what we are looking to do, to fix #2040  ?
- if this is going to result in a merge if I change implementation ?

chrisinmtown:
I tested V2 and observed that it responds to a curl invocation with two content-type headers with status code 415, unsupported media type, which seems exactly right for this situation. This PR makes V3 behave similarly.  Of course there are several different ways to achieve that behavior. @Ruwann please say what you'd like here.

PR #2071, opened 2025-07-30
chore: move pre-commit to dev-dependency group

Lewiscowles1986:

 - just to move pre-commit out of the test group, and into the dev group.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,8 +69,10 @@ swagger-ui = ["swagger-ui-bundle"]
 uvicorn = ["uvicorn"]
 mock = ["jsf"]
 
-[tool.poetry.group.tests.dependencies]
+[tool.poetry.group.dev.dependencies]
 pre-commit = "~2.21.0"
+
+[tool.poetry.group.tests.dependencies]
 pytest = "7.2.1"
 pytest-asyncio = "~0.18.3"
 pytest-cov = "~2.12.1"


chrisinmtown:
Does the V2 branch need the same change?

Lewiscowles1986:
Hey @chrisinmtown hope you are well. Maybe, but isn't 2.x deprecated?

chrisinmtown:
I believe no, not deprecated. The community is trying to upgrade the V2 branch dependencies for all the Connexion users who are not ready to upgrade to ASGI & V3. You can see the open PRs.

Lewiscowles1986:
There isn't a way for this PR to do both, but I can take a follow-up action to PR the 2 branch as well if of use; unless you would like to?

I Do hope the 2 to 3 upgrade concludes soon, as otherwise I fear it will take up a lot of headspace.

The 415 PR #2070  hasnt merged yet either; so I'm guessing everyone is busy or enjoying summer as they should be. This is a nit-pick, because I think all local-dev, regardless of tests should run pre-commit

chrisinmtown:
> I Do hope the 2 to 3 upgrade concludes soon

I can only offer a personal perspective here.  For my projects, ASGI does not help. Further, there are critical bugs in v3.2 that break existing functionality like using swagger and certain Open API schema elements. There are open PRs for those problems that have not been acted upon. So the v3 upgrade is a huge challenge right now.

RobbeSneyders:
I don't see why this needs to be moved into a separate group. Poetry assigns no special meaning to "dev" or "tests". Moving it would also required changes in other places in the repo, where we install these dependencies.

PR #2088, opened 2025-10-13
Drop Python 3.9 and add 3.13, 3.14

RobbeSneyders:


connexion/utils.py
```diff
@@ -499,7 +499,7 @@ def __init__(self, path: str) -> None:
             self.path_regex, _, _ = compile_path(self.path)
 
         def __lt__(self, other: "SortableRoute") -> bool:
-            return bool(other.path_regex.match(self.path))
+            return bool(other.path_regex.match(self.path)) or self.path < other.path
```
chrisinmtown, line 502: This is so subtle. I feel it really is worth adding a comment why it was necessary.

chrisinmtown:
FWIW this PR dropped *3.8* (not 3.9) and added *3.14* (not 2.14).

Issue #1890, opened 2024-02-21
swagger-ui give a 500 error when attempting to view


jdkent:
### Description

when cloning the raw spec, the `base_uri` information is lost, meaning external references cannot be resolved.

### Expected behaviour

the spec to be displayed

### Actual behaviour

500 error

### Steps to reproduce

run the test in `test_swagger_ui.py` in #1889 with the modified simple app

### Additional info:

Output of the commands:

- `python --version`
- `pip show connexion | grep "^Version\:"`

Issue #2028, opened 2025-01-16
Swagger UI Infinite Recursion Issue in 3.2.0


Parthib:
We recently upgraded to the latest version of Connexion due to the `python_multipart` package issue.

```
connexion[swagger-ui]==3.1.0
connexion[flask]==3.1.0
```

to

```
connexion[swagger-ui]==3.2.0
connexion[flask]==3.2.0
```

Everything seemed to be working fine, but we can't access our Swagger UI page anymore. If we attempt to, we get a 500, and the logs show this:

```
INFO:     127.0.0.1:62383 - "GET /v1/ui/swagger-ui-config.json HTTP/1.1" 200 OK
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/starlette/routing.py", line 73, in app
    response = await f(request)
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/middleware/swagger_ui.py", line 110, in _get_openapi_json
    content=jsonifier.dumps(self._spec_for_prefix(request)),
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/middleware/swagger_ui.py", line 73, in _spec_for_prefix
    return self.specification.with_base_path(base_path).raw
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/spec.py", line 221, in with_base_path
    new_spec = self.clone()
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/spec.py", line 207, in clone
    return type(self)(copy.deepcopy(self._spec))
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/spec.py", line 82, in __init__
    self._validate_spec(raw_spec)
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/connexion/spec.py", line 96, in _validate_spec
    validator.validate(spec)
  File "/Users/parthib/miniconda3/envs/backend/lib/python3.9/site-packages/jsonschema/validators.py", line 450, in validate
    for error in self.iter_errors(*args, **kwargs):
[...]

Issue #1909, opened 2024-04-08
Splitting spec files still throws schema and file not found errors (json_schema exceptions) but still works with Connexion <= 3.0.2


eharvey71:
I have my yml files divided into subdirectories for paths, parameters, schemas, etc. This worked fine (and still works fine) on Connexion ~~3.0.4~~ 3.0.2.
Connexion ~~3.0.5 and 3.0.6~~ 3.0.3 - 3.0.6 are still throwing errors. The prior fixes for $refs don't seem to have fixed issues. I thought this was just a Windows filesystem or jsonschema issue but it's happening all over - in my MacOS, Windows and docker container deployments running Debian. 

***************
**If I downgrade to Connexion ~~3.0.4~~ 3.0.2 or prior, the error goes away, in all cases.**
**************

I am working with a separate swagger UI deployment, so I can do customizations.

In my configuration, I'm initializing my app like this:

```
basedir = pathlib.Path(__file__).parent.resolve()
swagoptions = SwaggerUIOptions(swagger_ui = True, swagger_ui_template_dir = basedir / 'swagger-ui')
connex_app = FlaskApp(__name__, specification_dir=basedir / "apispecs")
```

Paths are like this to specs, where my swagger.yml contains $refs to the proper paths of each of the divided spec files:
```
project/
      │── apispecs/
               |── swagger.yml
               |── parameters/
                         |── _index.html
               │── securitySchemas/
                         |── _index.html
               │── schemas/
                         |── _index.html
```

swagger.yml sample:

```
components:
  schemas:
    $ref: "./schemas/_index.yml"
  parameters:
    $ref: "./parameters/_index.yml"
  securitySchemes:
    $ref: "./security/_index.yml"
```

Full contents of error:
```
_RefResolutionError(_cause=FileNotFoundError(2, 'No such file or directory'))
Traceback (most recent call last):
  File "/Users/eharvey71/dev/integration-bridge/venv/lib/python3.12/site-packages/connexion/json_schema.py", line 88, in _do_resolve
    retrieved = deep_get(spec, path)
                ^^^^^^^^^^^^^^^^^^^^
  File "/Users/eharvey71/dev/integration-bridge/venv/lib/python3.12/site-packages/connexion/utils.py", line 112, in deep_get
    return deep_get(obj[keys[0]], keys[1:])
                    ~~~^^^^^^^^^
KeyError: 'schemas'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2089, opened 2025-10-13
Fix spec clone method and usage of raw vs resolved spec

RobbeSneyders:
This PR fixes an issue introduced in #2002, and the original issue #2002 was trying to address.

The original issue was that a cloned spec did not have properly resolved references. #2002 fixed this incorrectly by cloning the resolved spec, while the `Spec` initializer expects a raw spec.

This PR fixes this by cloning the raw spec, and passing the `base_uri` required to resolve it along to the initializer of the new `Spec` instance.

The swagger ui was also updated to use the resolved spec instead of the raw spec.

Supersedes:
#1889
#2080

Fixes:
#1890
#1909
#2028 
#2029

### DIFF
--- a/connexion/middleware/swagger_ui.py
+++ b/connexion/middleware/swagger_ui.py
@@ -70,7 +70,7 @@ class SwaggerUIAPI(AbstractSpecAPI):
         This is needed when behind a path-altering reverse proxy.
         """
         base_path = self._base_path_for_prefix(request)
-        return self.specification.with_base_path(base_path).raw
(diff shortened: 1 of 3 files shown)


chrisinmtown:
I would like to suggest adding a test for the `clone` method to test this directly, confirm the base URI is cached. I don't understand the change well enough yet to offer a better test that checks the failure directly.

RobbeSneyders:
Thanks @chrisinmtown, I expanded the test to include a clone step and reevaluate the cloned specification.

PR #2092, opened 2025-10-13
Restore `tests` argument to poetry invocation of pytest

chrisinmtown:
Drop ineffective options section "[pytest]"

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -33,9 +33,6 @@ python =
     3.13: py313-pypi
     3.14: py314-pypi
 
-[pytest]
-testpaths = tests
-
 [testenv]
 setenv=PYTHONPATH = {toxinidir}:{toxinidir}
 deps=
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
The tests still seem to run fine though?
https://github.com/spec-first/connexion/actions/runs/18464071838/job/52601839745?pr=2067

chrisinmtown:
> The tests still seem to run fine though?

Yes the tests run. If I understand the pytest behavior correctly, it simply scans recursively for a directory named `tests`.  Exactly one of those exists in this repo as of this commit, so the test workflows continue to function as expected. I first noticed the problem in my PR that offers a new directory in `examples` with a subdirectory `tests` and a file in there `conftest.py`.  When pytest is allowed to scan the entire repo, it picks those up, and that little `conftest.py` file breaks the run.  So I circled back to fix my botch here.  Please consider.

RobbeSneyders (approved):
Ok thanks for the explanation

PR #2097, opened 2025-11-04
Bump minimum version of httpx to 0.24

chrisinmtown:
Make tests work in Python 3.13 by avoiding `httpx` version 0.23  import of package `cgi`

 - Change pyproject.toml to require `httpx` at minimum version 0.24, which does not use `cgi`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ connexion = 'connexion.cli:main'
 [tool.poetry.dependencies]
 python = '^3.9'
 asgiref = ">= 3.4"
-httpx = ">= 0.23"
+httpx = ">= 0.24"
 inflection = ">= 0.3.1"
 jsonschema = ">=4.17.3"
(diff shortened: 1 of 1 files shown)


chrisinmtown:
When I run the `tox` suite from the beginning, in the py39-min run, poetry pins the `httpx` version at the minimum version 0.23.0.  The `poetry.lock` fie persists across the following tox runs. This fails when it reaches py3.13 because package `httpx==0.23.0` imports package `cgi` which is not available in py3.13.

If I change the `tox.ini` command to be `poetry lock --regenerate` then all tests pass including python 3.13. In fact, with that new option in place, I don't have to bump the minimum version of `httpx`.

Please comment. I'll be glad to change this PR to leave the versions alone and just change `tox.ini`

chrisinmtown:
I still recommend this version bump, but this PR is largely superseded by https://github.com/spec-first/connexion/pull/2098

PR #2118, opened 2026-04-23
#2117 :

### LINKED ISSUE #2117: Consistency between fallback of RequestValidation Middleware and Lifecycle
### Problem ###
We notice that the fallback used in RequestValidation and Lifecycle are not aligned.

**_In RequestValidation:_**
The fallback is the first mime type defined in the openapi file if one is defined and if None, then it is "application/octet-stream".


**_In Lifecycle:_**
The fallback in case of missing Content-Type is "application/octet-stream".

This is resulting in some bugs when mime-type is not defined when request body is a JSON.
Indeed, in the following code, `self.content_type` equals "`application/octet-stream`" resulting in object "`bytes`" being returned : 
```
  async def get_body(self):                                                                                                                                                                                                                                                                                         
      if is_json_mimetype(self.content_type):                                                                                                                                                                                                                                                                       
          return await self.json()                                                                                                                                                                                                                                                                                  
      ...                                                                                                                                                                                                                                                                                                           
      else:
          return await self.body() or None        
```

```
WSGI_0s - 0 0 - ERROR - api.provide.impl.magic_cart.controllers.manage_express_order_controller - 'bytes' object has no attribute 'buyer_id'
  File "/home/brultea/IdeaProjects/lcdp-magic-cart/api/provide/impl/magic_cart/controllers/manage_express_order_controller.py", line 26, in create_express_order
    buyer_id = express_order_creation_parameters.buyer_id
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'bytes' object has no attribute 'buyer_id'
AnyIO worker threads - 0 0 - ERROR - connexion.middleware.exceptions - HTTPException(status_code=500, detail="'bytes' object has no attribute 'buyer_id'")
AnyIO worker threads - 0 0 - ERROR - connexion.middleware.exceptions - HTTPException(status_code=500, detail="'bytes' object has no attribute 'buyer_id'")
MainThreads - 0 0 - INFO - uvicorn.access - 127.0.0.1:50613 - "POST /api/v1/express-orders HTTP/1.1" 500
```

This issue may be linked to https://github.com/spec-first/connexion/issues/1269.

### Proposal ###
In RequestValidation, if we fallback on a mime type located in the OpenApi file, then we should change the scope header "Content-Type" to match the fallback.

AntoineDuComptoirDesPharmacies:
Fixes #2117.

 - If mime type is inferred from openapi spec, then add it to the request scope

### DIFF
--- a/connexion/middleware/request_validation.py
+++ b/connexion/middleware/request_validation.py
@@ -33,14 +33,17 @@ class RequestValidationOperation:
 
     def extract_content_type(
         self, headers: t.List[t.Tuple[bytes, bytes]]
-    ) -> t.Tuple[str, str]:
+    ) -> t.Tuple[str, str, bool]:
         """Extract the mime type and encoding from the content type headers.
 
         :param headers: Headers from ASGI scope
 
-        :return: A tuple of mime type, encoding
+        :return: A tuple of (mime_type, encoding, inferred) where ``inferred``
+            is ``True`` when no ``Content-Type`` header was present in the
+            request and the mime type was guessed from the operation spec.
         """
         content_type = utils.extract_content_type(headers)
+        inferred = content_type is None
         mime_type, encoding = utils.split_content_type(content_type)
         if mime_type is None:
             # Content-type header is not required. Take a best guess.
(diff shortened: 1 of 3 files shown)


AntoineDuComptoirDesPharmacies:
No news ? Do i have something to change ?

AntoineDuComptoirDesPharmacies:
Hi @Ruwann,
Sorry to ping you but i would like to know if you are interested in reviewing this PR.

Thanky ou in advance !

PR #2124, opened 2026-05-25
Pin black to version 26.3.1 in .pre-commit-config.yaml

chrisinmtown:
Fixes these failures at pre-commit.ci when using Python 3.14:

    RuntimeError: There is no current event loop in thread 'MainThread'.


chrisinmtown:
The very latest version of black is 26.5.1 released a week ago. Here I chose a version that is two months old. I hope the maintainers will comment on their preference.

PR #2125, opened 2026-05-25
(v2) Pin black to version 26.3.1 in .pre-commit-config.yaml

chrisinmtown:
Fixes these failures at pre-commit.ci when using black 22.3.0 in Python 3.14:

    RuntimeError: There is no current event loop in thread 'MainThread'.


chrisinmtown:
Many of us still rely on Connexion v2 so I'm still trying to keep this branch healthy.

